PR #13, opened 2025-11-26
Rename go module name

Karthik-K-N:
Rename go module name after community repo migration.


PROJECT
```diff
@@ -7,14 +7,14 @@ domain: nodereadiness.io
 layout:
 - go.kubebuilder.io/v4
 projectName: nrgcontroller
-repo: github.com/ajaysundark/node-readiness-gate-controller
+repo: sigs.k8s.io/node-readiness-gate-controller
```
ajaysundark: Should we also change this to `node-readiness-controller` to align with the GitHub repo name?

PROJECT
```diff
@@ -7,14 +7,14 @@ domain: nodereadiness.io
...
 resources:
 - api:
     crdVersion: v1
   controller: true
   domain: nodereadiness.io
```
ajaysundark, line 15: We also need the domain name change from `nodereadiness.io` to `readiness.node.x-k8s.io` for the sigs project. Do you want to handle that in the same PR?

Karthik-K-N: While working on that I came across this, I will submit a separate PR for that.

PROJECT
Karthik-K-N: sure, will update that

ajaysundark (approved):
/lgtm

PR #17, opened 2025-12-03
Update Makefile to use the hack/tools/bin rather than bin

Karthik-K-N:
Just a refactor to Makefile, This would enable us to add more validation/test scripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,18 +12,53 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+.DEFAULT_GOAL:=help
+
+#
+# Directories.
+#
+# Full directory of where the Makefile resides
+ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+BIN_DIR := bin
+TEST_DIR := test
+TOOLS_DIR := hack/tools
+TOOLS_BIN_DIR := $(abspath $(TOOLS_DIR)/$(BIN_DIR))
+GO_INSTALL := ./scripts/go_install.sh
+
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -1,16 +1,65 @@
+#Copyright 2025.
```
ajaysundark: Would this likely need a license update as well?

Karthik-K-N: Done.

Makefile
```diff
@@ -12,19 +12,54 @@
...
+.DEFAULT_GOAL:=help
+
+#
+# Directories.
+#
+# Full directory of where the Makefile resides
+ROOT_DIR:=$(shell dirname $(realpath $(firstword $(MAKEFILE_LIST))))
+BIN_DIR := bin
+TEST_DIR := test
+TOOLS_DIR := hack/tools
```
ajaysundark, line 24: could you also check whether this path change need any update in any contributor guidelines/docs?

Makefile
```diff
@@ -56,16 +94,6 @@ all: build
 help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-##@ Development
```
ajaysundark, line 59: If possible, run the all targets to confirm this change does not break existing scripts or developer workflows

Karthik-K-N:
I have run all the targets and they work as expected. This should  be ready to be merged.

PR #22, opened 2025-12-04
Enable KAL linter

Karthik-K-N:
This PR enables KAL linter, More about it can be read here https://github.com/kubernetes-sigs/kube-api-linter

Only one linter is enabled, Follow up task is to explore and add other KAL linters https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md and also enable it in github action / CI jobs

### DIFF
--- a/.golangci-kal.yml
+++ b/.golangci-kal.yml
@@ -1,0 +1,32 @@
+version: "2"
+run:
+  go: "1.24"
+  allow-parallel-runners: true
+linters:
+  default: none
+  enable:
+    - kubeapilinter # linter for Kube API conventions
+  settings:
+    custom:
+      kubeapilinter:
+        type: module
+        description: KAL is the Kube-API-Linter and lints Kube like APIs based on API conventions and best practices.
+        settings:
+          linters:
+            enable:
+              - "nophase" # Phase fields are discouraged by the Kube API conventions, use conditions instead.
+            disable:
+            - "*" # We will manually enable new linters after understanding the impact. Disable all by default.
+  exclusions:
+    generated: strict
+    paths:
+      - ".*_test.go"  # Exclude test files.
+    rules:
+    ## KAL should only run on API folders.
+    - path-except: "api//*"
+      linters:
+        - kubeapilinter
+
+issues:
+  max-same-issues: 0
+  max-issues-per-linter: 0
--- a/hack/tools/.custom-gcl.yaml
+++ b/hack/tools/.custom-gcl.yaml
@@ -1,0 +1,6 @@
+version: v2.5.0
+name: golangci-lint-kube-api-linter
+destination: ./bin
+plugins:
+- module: 'sigs.k8s.io/kube-api-linter'
+  version: v0.0.0-20251107220451-f87b38edda8c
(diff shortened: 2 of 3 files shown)


hack/tools/.custom-gcl.yaml
```diff
@@ -0,0 +1,6 @@
+version: v2.5.0
+name: golangci-lint-kube-api-linter
+destination: ./bin
+plugins:
+- module: 'sigs.k8s.io/kube-api-linter'
+  version: v0.0.0-20251107220451-f87b38edda8c
```
ajaysundark, line 6: How do we select the version? is the convention just to pin to a commit?

.golangci-kal.yml
```diff
@@ -0,0 +1,32 @@
...
+  settings:
+    custom:
+      kubeapilinter:
+        type: module
+        description: KAL is the Kube-API-Linter and lints Kube like APIs based on API conventions and best practices.
+        settings:
+          linters:
+            enable:
+              - "nophase" # Phase fields are discouraged by the Kube API conventions, use conditions instead.
+            disable:
```
ajaysundark, line 18: How do we onboard to new rules or select what to onboard? I'd expect to be many existing violations in the api in its current form.

hack/tools/.custom-gcl.yaml
```diff
@@ -0,0 +1,6 @@
+version: v2.5.0
+name: golangci-lint-kube-api-linter
+destination: ./bin
+plugins:
+- module: 'sigs.k8s.io/kube-api-linter'
```
ajaysundark, line 5: Is there a github action for this? Should we add it to [.github/workflows/lint.yml](https://github.com/kubernetes-sigs/node-readiness-controller/blob/23f988015375d54c8d02f5aabfd313cc55a54977/.github/workflows/lint.yml) as well?

ajaysundark (approved):
Thanks for taking care of this. I agree that incorporating this linter is beneficial for overall code health.

I hadn’t focused much on the kubebuilder markers during the initial API design, so it’s helpful to have tools like this to enforce best practices. That said, maintaining additional linters might increase developer toil, so we should consider how to strike a balance and avoid slowing down new contributors.

/lgtm

.golangci-kal.yml
Karthik-K-N: Yes, Ultimately we want to enable most of the rules which will help us in making perfect API.
Here is a reference of K8s efforts: https://github.com/kubernetes/kubernetes/issues/134671
And this is how done in CAPI: https://github.com/kubernetes-sigs/cluster-api/blob/main/.golangci-kal.yml

Karthik-K-N: When enabling each linter we can debate and make required API changes, Given that we are in alpha I believe we are open to change.

hack/tools/.custom-gcl.yaml
Karthik-K-N: Github action is not yet enabled, should be added and its next step.

hack/tools/.custom-gcl.yaml
Karthik-K-N: This is until we have a release, Here is the reference https://github.com/kubernetes-sigs/kube-api-linter?tab=readme-ov-file#golangci-lint-module

PR #27, opened 2025-12-08
Rename kind name from nodereadinessgaterules to nodereadinessrules

Karthik-K-N:


config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml
```diff
@@ -4,23 +4,22 @@ kind: CustomResourceDefinition
...
   group: readiness.node.x-k8s.io
   names:
-    kind: NodeReadinessGateRule
-    listKind: NodeReadinessGateRuleList
-    plural: nodereadinessgaterules
+    kind: NodeReadinessRule
+    listKind: NodeReadinessRuleList
+    plural: nodereadinessrules
     shortNames:
     - nrgr
```
ajaysundark: we could as well change the shortName to `nrr` in sync with this change.

xref: https://github.com/kubernetes-sigs/node-readiness-controller/pull/10 where this was introduced

Karthik-K-N: Done. Thanks.

PR #37, opened 2025-12-17
Rename from nrg(node readiness gate) to nrr (node readiness rule)

Karthik-K-N:
Recently we renamed the KIND from NodeReadinessGate to NodeReadinessRule via PR https://github.com/kubernetes-sigs/node-readiness-controller/pull/27

As a follow up we need to change from nrg to nrr to be consistent.

### DIFF
--- a/docs/TEST_README.md
+++ b/docs/TEST_README.md
@@ -1,15 +1,15 @@
 # Node Readiness Gates E2E Test Guide (Kind)
 
-This guide details how to run an end-to-end test for the Node Readiness Gates (NRG) controller using a local Kind cluster.
+This guide details how to run an end-to-end test for the Node Readiness Rules (NRR) controller using a local Kind cluster.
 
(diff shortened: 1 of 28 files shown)


docs/TEST_README.md
```diff
@@ -25,7 +25,7 @@ The test uses a 3-node Kind cluster:
 The provided Kind configuration creates the 3-node topology with the necessary labels and taints.
 
 ```bash
-kind create cluster --config config/testing/kind/kind-config.yaml
+kind create cluster --config config/testing/kind/kind-3node-config.yaml
```
Karthik-K-N, line 28: Updated here.
@Priyankasaggu11929

Priyankasaggu11929: Thank you :pray:

Karthik-K-N:
/hold cancel
ready for review
There is a bug in controller which is causing few fields in the status field to be ommited, which needs to be fixed

PR #41, opened 2025-12-18
API changes with go doc and kubebuilder validations

Karthik-K-N:


.golangci.yml
```diff
@@ -4,23 +4,46 @@ run:
 linters:
```
ajaysundark, line 4: Could we move the linter configuration update into a separate PR?

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -21,87 +21,271 @@ import (
...
+	// EnforcementModeBootstrapOnly applies configuration only during the first reconcile.
+	EnforcementModeBootstrapOnly EnforcementMode = "bootstrap-only"
+
+	// EnforcementModeContinuous continuously monitors and enforces the configuration.
+	EnforcementModeContinuous EnforcementMode = "continuous"
+)
+
+// TaintStatus specifies status of the Taint on Node.
+// +kubebuilder:validation:Enum=Present;Absent
+type TaintStatus string
```
ajaysundark, line 38: This looks cleaner, thanks for this api refactoring!

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -21,87 +21,271 @@ import (
...
-	Conditions []ConditionRequirement `json:"conditions"`
+	// conditions contains a list of the Node conditions that defines the specific
+	// criteria that must be met for taints to be managed on the target Node.
+	// The presence or status of these conditions directly triggers the application or removal of Node taints.
+	//
+	// +required
+	// +listType=map
+	// +listMapKey=type
+	// +kubebuilder:validation:MinItems=1
+	// +kubebuilder:validation:MaxItems=32
```
ajaysundark, line 58: Could we document these restrictions? maybe create a doc issue if you think we could follow up separately and figure out where to put it.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -21,87 +21,271 @@ import (
...
+// required to trigger the controller's action.
 type ConditionRequirement struct {
-	Type           string                 `json:"type"`
-	RequiredStatus corev1.ConditionStatus `json:"requiredStatus"`
-}
+	// type of Node condition
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=316
```
ajaysundark, line 98: how did we arrive at this length? if there is a ref in k8s could we capture as a comment here?

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -21,87 +21,271 @@ import (
...
 // NodeReadinessRuleStatus defines the observed state of NodeReadinessRule.
+// +kubebuilder:validation:MinProperties=1
 type NodeReadinessRuleStatus struct {
-	// Keep existing
-	ObservedGeneration int64              `json:"observedGeneration,omitempty"`
-	Conditions         []metav1.Condition `json:"conditions,omitempty"`
-	AppliedNodes       []string           `json:"appliedNodes,omitempty"`
+	// conditions represents the observations of a NodeReadinessRule's current state.
+	//
+	// +optional
```
ajaysundark: I'm not familiar with this, is the expectation on 'status' fields to always be optional? I dont imagine in a real status update the 'generation' to be empty

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -21,87 +21,271 @@ import (
...
+	FailedNodes []NodeFailure `json:"failedNodes,omitempty"`
+
+	// nodeEvaluations provides detailed insight into the rule's assessment for individual Nodes.
+	// This is primarily used for auditing and debugging why specific Nodes were or
+	// were not targeted by the rule.
+	//
+	// +optional
+	// +listType=map
+	// +listMapKey=nodeName
+	// +kubebuilder:validation:MaxItems=1000
```
ajaysundark: Does this limit the status update for a rule to be maximum of 1000 nodes?

As we discussed offline, there's room for improvement for status updates on large scale scenarios. We could skip this validation for now and plan after we get a chance to run some scale tests and observe behavior for 1k+ nodes.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -21,87 +21,271 @@ import (
...
+}
+
+// NodeFailure provides diagnostic details for Nodes that could not be successfully evaluated by the rule.
+type NodeFailure struct {
+	// nodeName is the name of the failed Node.
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
```
ajaysundark, line 163: SAA, please see if you could add a comment reg a ref for this pattern in k8s

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -21,87 +21,271 @@ import (
...
+	// for this Node. This allows for granular auditing of which specific
+	// criteria passed or failed during the rule assessment.
+	//
+	// +required
+	// +listType=map
+	// +listMapKey=type
+	// +kubebuilder:validation:MaxItems=1000
+	ConditionResults []ConditionEvaluationResult `json:"conditionResults,omitempty"`
+
+	// taintStatus represents the taint status on the Node, one of Preset, Absent.
```
ajaysundark: nit: Present

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -21,87 +21,271 @@ import (
...
+	// +optional
+	EnforcementMode EnforcementMode `json:"enforcementMode,omitempty"`
 
-	// Simplify taint specification (remove TaintKey, TaintEffect separation)
-	Taint TaintSpec `json:"taint"`
+	// taint defines the specific Taint (Key, Value, and Effect) to be managed
+	// on Nodes that meet the defined condition criteria.
+	//
+	// +required
+	Taint corev1.Taint `json:"taint,omitempty"`
```
ajaysundark, line 74: 👍

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -21,87 +21,271 @@ import (
...
 
-	// Keep existing fields
+	// nodeSelector limits the scope of this rule to a specific subset of Nodes.
+	// If unspecified, this rule applies to all Nodes in the cluster.
+	//
+	// +optional
```
ajaysundark, line 79: While we refactoring the types, I'm inclined to also make this as required. We can start with 'more restrictive' configuration and could open this if there's any need in the future.

api/v1alpha1/nodereadinessrule_types.go
Karthik-K-N: Done, will follow up with issue https://github.com/kubernetes-sigs/node-readiness-controller/issues/46
I had this open question but forgot to mention it in the PR.

api/v1alpha1/nodereadinessrule_types.go
Karthik-K-N: I referred type from here https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition

api/v1alpha1/nodereadinessrule_types.go
Karthik-K-N: Nothing as such, But for now I dont see anywhere we set condition for status in code, I thought it for future case?

api/v1alpha1/nodereadinessrule_types.go
Karthik-K-N: sure, But can we keep this like this for now and later increase based on the test?

api/v1alpha1/nodereadinessrule_types.go
Karthik-K-N: Sure, Its from here https://github.com/kubernetes/apimachinery/blob/84d740c9e27f3ccc94c8bc4d13f1b17f60f7080b/pkg/util/validation/validation.go#L198

api/v1alpha1/nodereadinessrule_types.go
Karthik-K-N: Can I do that in the follow up PR, It needs some tests fix as well.

api/v1alpha1/nodereadinessrule_types.go
ajaysundark: hmm. you're right Conditions dont belong at rule.status, the results are under NodeEvaluations already. I think we can remove it.

My comment on optional was at 'ObservedGeneration'.

api/v1alpha1/nodereadinessrule_types.go
ajaysundark: Can we keep it as 5k, since many of the distros have it as supported number of nodes?

api/v1alpha1/nodereadinessrule_types.go
ajaysundark: sounds good.

api/v1alpha1/nodereadinessrule_types.go
Karthik-K-N: sure

api/v1alpha1/nodereadinessrule_types.go
Karthik-K-N: Yes, I can make it as required as its should be set by the controller, but we dont set that during the DryRun so we can't make it as required.

ajaysundark: Ah okay.

ajaysundark: Yeah dryrun rule is not considered yet as released. We could keep this as optional for now

api/v1alpha1/nodereadinessrule_types.go
Karthik-K-N: For completeness, Created a follow up PR: https://github.com/kubernetes-sigs/node-readiness-controller/pull/56

Issue #43, opened 2025-12-18
Fix status update inconsistency


Karthik-K-N:
1. The status of NRR(Node ResourceRule) is not being correct updated, Few fields are missing

ObservedGeneration and AppliedNodes not seen in status.
```
status:
  nodeEvaluations:
  - conditionResults:
    - currentStatus: Unknown
      missing: true
      requiredStatus: "True"
      satisfied: false
      type: network.k8s.io/CalicoReady
    lastEvaluated: "2025-12-17T05:40:59Z"
    nodeName: nrg-test-worker
    taintStatus: Present
  - conditionResults:
    - currentStatus: Unknown
      missing: true
      requiredStatus: "True"
      satisfied: false
      type: network.k8s.io/CalicoReady
    lastEvaluated: "2025-12-17T05:41:00Z"
    nodeName: nrg-test-worker2
    taintStatus: Present

```

2. DryRun results are not removed

a. Create NRR with dryRun=true
b. Edit the NRR to remove dryRun - expectation is dryRunResults should be removed from status but its not

```
~ % kubectl get nrr -o yaml
apiVersion: v1
items:
- apiVersion: readiness.node.x-k8s.io/v1alpha1
  kind: NodeReadinessRule
  metadata:
    creationTimestamp: "2025-12-18T10:53:50Z"
    finalizers:
    - readiness.node.x-k8s.io/cleanup-taints
    generation: 2
    name: network-readiness-rule
    resourceVersion: "761"
    uid: fb78c4e6-fe29-40d9-b3bd-055bc3997262
  spec:
    conditions:
    - requiredStatus: "True"
      type: network.k8s.io/CalicoReady
    enforcementMode: continuous
    nodeSelector:
      matchExpressions:
      - key: node-role.kubernetes.io/control-plane
        operator: DoesNotExist
    taint:
      effect: NoSchedule
      key: readiness.k8s.io/NetworkReady
      value: pending
  status:
    dryRunResults:
      affectedNodes: 2
      riskyOperations: 2
      summary: would add 1 taints, 2 nodes have missing conditions
      taintsToAdd: 1
      taintsToRemove: 0
    nodeEvaluations:
    - conditionResults:
      - currentStatus: Unknown
        missing: true
        requiredStatus: "True"
        satisfied: false
        type: network.k8s.io/CalicoReady
      lastEvaluated: "2025-12-18T10:54:48Z"
      nodeName: nrr-test-worker
      taintStatus: Present
[...]

PR #44, opened 2025-12-19
Check rule generation before updating the staus

Karthik-K-N:
Since both node controller and node readiness controller trying to update the status of rule, There may be  conflict where the status updated by the NRR controller overriden by node controller, this tries to match the rule generation and make sure both are at the same level before updating the status,

From Error logs

```
2025-12-17T20:09:42+05:30	DEBUG	Updating rule status	{"controller": "nodereadiness-controller", "object": {"name":"network-readiness-rule"}, "namespace": "", "name": "network-readiness-rule", "reconcileID": "46cd3c25-2e2a-4e51-a39a-5852d051e052", "rule": "network-readiness-rule", "nodeEvaluations": 2, "appliedNodes": 2}
.
.
.
2025-12-17T20:09:53+05:30	DEBUG	Updating rule status	{"controller": "node", "controllerGroup": "", "controllerKind": "Node", "Node": {"name":"nrg-test-worker2"}, "namespace": "", "name": "nrg-test-worker2", "reconcileID": "49acd35b-8ad5-4da8-902f-102b25da14a1", "rule": "network-readiness-rule", "nodeEvaluations": 1, "appliedNodes": 0}
```

Fixes: https://github.com/kubernetes-sigs/node-readiness-controller/issues/43

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -457,6 +457,11 @@ func (r *ReadinessGateController) updateRuleStatus(ctx context.Context, rule *re
 			return err
 		}
 
+		if latestRule.GetGeneration() != rule.GetGeneration() {
+			log.V(4).Info("Rule generation mismatch during status update, avoiding retry to let new reconciliation handle it")
+			return nil
+		}
+
 		// Merge our status updates into fresh version
 		// This ensures we're updating based on the latest resourceVersion
 		latestRule.Status.NodeEvaluations = rule.Status.NodeEvaluations


internal/controller/nodereadinessrule_controller.go
```diff
@@ -457,6 +457,11 @@ func (r *ReadinessGateController) updateRuleStatus(ctx context.Context, rule *re
 			return err
 		}
 
+		if latestRule.GetGeneration() != rule.GetGeneration() {
```
ajaysundark, line 460: The core of the issue seems to me that node-reconciler overwrites correct (recent) values with stale/empty values from cache. 

For instance:
1. T0: NodeReconciler starts rule.gen=1
2. T1: RuleReconciler updates rule.gen=2
3. T2: NodeReconciler updates with copying rule from cache?

I think instead of calling the full updateRuleStatus(), NodeReconciler should use Patch ONLY the fields it modifies.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -457,6 +457,11 @@ func (r *ReadinessGateController) updateRuleStatus(ctx context.Context, rule *re
...
+			log.V(4).Info("Rule generation mismatch during status update, avoiding retry to let new reconciliation handle it")
+			return nil
```
ajaysundark, line 462: I think returning here silently is problematic as it would skip the node update entirely..

internal/controller/nodereadinessrule_controller.go
ajaysundark: Let me try this and see if it helps.

Karthik-K-N: >NodeReconciler should use Patch ONLY the fields it modifies.

This will be the best solution of all.

ajaysundark: My chances here: https://github.com/ajaysundark/node-readiness-controller/commit/0f1b4fc5ca99c317a353e95ea4e06d17033779e9. Haven't gotten a chance to test this fully yet

ajaysundark: https://github.com/kubernetes-sigs/node-readiness-controller/pull/49 for your review fixing the status updates.

rule generation mismatch suggests the node reconciliation is dealing with a older rule and next rule-reconciliation should handle this pending node update anyway; so, including this check makes sense (could be an optimization to save some redundant reconciliation cycles).

Karthik-K-N:
In favor of https://github.com/kubernetes-sigs/node-readiness-controller/pull/49
/close

Issue #48, opened 2025-12-22
Fix make undeploy command


ajaysundark:
"make undeploy" command does not seem to complete but hangs during execution

**Steps to reproduce:**
1. Follow test setup in [TEST_README.md](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/docs/TEST_README.md).
2. As part of cleanup, perform 'make undeploy' before deleting the kind cluster.

The command does not complete to success:

```
$ make undeploy
/usr/local/google/home/ajaysundar/workspace/src/k8s.io/node-readiness-controller/hack/tools/bin/kustomize-v5.7.0 build config/default | kubectl delete --ignore-not-found=true -f -
namespace "nrr-system" deleted
customresourcedefinition.apiextensions.k8s.io "nodereadinessrules.readiness.node.x-k8s.io" deleted
serviceaccount "nrr-controller-manager" deleted
role.rbac.authorization.k8s.io "nrr-leader-election-role" deleted
clusterrole.rbac.authorization.k8s.io "nrr-manager-role" deleted
clusterrole.rbac.authorization.k8s.io "nrr-metrics-auth-role" deleted
clusterrole.rbac.authorization.k8s.io "nrr-metrics-reader" deleted
clusterrole.rbac.authorization.k8s.io "nrr-nodereadinessrule-admin-role" deleted
clusterrole.rbac.authorization.k8s.io "nrr-nodereadinessrule-editor-role" deleted
clusterrole.rbac.authorization.k8s.io "nrr-nodereadinessrule-viewer-role" deleted
rolebinding.rbac.authorization.k8s.io "nrr-leader-election-rolebinding" deleted
clusterrolebinding.rbac.authorization.k8s.io "nrr-manager-rolebinding" deleted
clusterrolebinding.rbac.authorization.k8s.io "nrr-metrics-auth-rolebinding" deleted
deployment.apps "nrr-controller-manager" deleted






^Cmake: *** [Makefile:231: undeploy] Error 1
```

AvineshTripathi:
I was able to reproduce this and found it’s caused by a finalizer on the CR. When the CRD is deleted, a DeleteTimestamp is set on the CR, but the controller (which is responsible for removing node taints and then clearing the finalizer: https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/nodereadinessrule_controller.go#L101) has already been deleted. As a result, the finalizer is never removed and the CR is stuck.

[...]

PR #53, opened 2025-12-25
docs: add uninstallation best practices and separate CRD from deploy

AvineshTripathi:
- Add documentation about proper uninstall order to avoid finalizer deadlocks
- Add recovery steps for stuck resources with manual finalizer removal
- Remove CRDs from config/default since `undeploy` and `uninstall` earlier used to do the same thing.
- CRDs are now managed separately via `make install`/`make uninstall`

Fixes: #48

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -15,7 +15,6 @@ namePrefix: nrr-
 #    someName: someValue
 
 resources:
-- ../crd
 - ../rbac
 - ../manager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -97,15 +97,27 @@ kubectl apply -f examples/network-readiness-rule.yaml
 
 ### Uninstallation
 
+> **Important**: Follow this order to avoid stuck resources due to finalizers.
+
+The controller adds a finalizer (`readiness.node.x-k8s.io/cleanup-taints`) to each `NodeReadinessRule` to ensure node taints are cleaned up before the rule is deleted. This means you must delete CRs **while the controller is still running**.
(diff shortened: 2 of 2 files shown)


config/default/kustomization.yaml
```diff
@@ -15,7 +15,6 @@ namePrefix: nrr-
 #    someName: someValue
 
 resources:
-- ../crd
```
ajaysundark, line 18: This seems like an unintended change. Could you please revert it?

AvineshTripathi: Hey @ajaysundark this is an intended change.

At the moment, **`undeploy` and `uninstall` effectively do the same thing**.

`undeploy`
```makefile
undeploy: $(KUSTOMIZE) ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
````

`undeploy` deletes everything rendered by `config/default`.

`uninstall`

```makefile
uninstall: manifests $(KUSTOMIZE) ## Uninstall CRDs from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
	@out="$$( $(KUSTOMIZE) build config/crd 2>/dev/null || true )"; \
	if [ -n "$$out" ]; then echo "$$out" | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -; else echo "No CRDs to delete; skipping."; fi
```

`uninstall` explicitly targets CRDs via `config/crd`.

`config/crd` is included in `config/default/kustomization.yaml`.
Because of this, running `undeploy` also deletes CRDs, making its behavior overlap with `uninstall`.

PR #56, opened 2026-01-05
Make node selector as required field

Karthik-K-N:
This PR is follow up from previous PR review comment: https://github.com/kubernetes-sigs/node-readiness-controller/pull/41#discussion_r2633794843

1. Made NodeSelector as required field
2. As its a required its on more required to be Pointer as the nil is no more valid value
3. Added corresponding webhook validations
4. Added necessary tests
5. Fixed failing tests as a consequence of this change.


api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -62,7 +62,6 @@ type NodeReadinessRuleSpec struct {
 	// enforcementMode is one of bootstrap-only, continuous.
 	// "bootstrap-only" applies the configuration once during initial setup.
 	// "continuous" ensures the state is monitored and corrected throughout the resource lifecycle.
-	// When omitted, default value will be "continuous".
```
Karthik-K-N, line 65: Noticed while doing this change, there is a hard check in the webhook for it be either bootstrap-only or continuos, should this be made as required?

ajaysundark: thanks for the catch. do we not have a defaulting behavior? I wonder what's the better UX, if certain 'features' can be configurable with an assumed default, or marking it explicit.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -71,13 +70,12 @@ type NodeReadinessRuleSpec struct {
...
-	Taint corev1.Taint `json:"taint,omitempty"`
+	Taint corev1.Taint `json:"taint,omitempty,omitzero"`
 
 	// nodeSelector limits the scope of this rule to a specific subset of Nodes.
-	// If unspecified, this rule applies to all Nodes in the cluster.
 	//
-	// +optional
-	NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
+	// +required
+	NodeSelector metav1.LabelSelector `json:"nodeSelector,omitempty,omitzero"`
```
ajaysundark, line 78: I'm not familiar with this to suggest what's better. but 'required' and 'omitting nil fields' seem to be contradicting. No harm omitting nil fields, but I think you're handling with API server  validation anyway

PR #57, opened 2026-01-07
feat: Implement and integrate a new readiness condition reporter component

ajaysundark:
1. Copy source code from https://github.com/ajaysundark/readiness-condition-reporter/
2.  Add support for building COMPONENT=reporter in build-and-publish-image script
3. Unit-tests and manual verification

### DIFF
--- a/cmd/readiness-condition-reporter/main.go
+++ b/cmd/readiness-condition-reporter/main.go
@@ -25,7 +25,7 @@ const (
 	defaultHTTPTimeout   = 10 * time.Second
 )
 
-// HealthResponse represents the health check response structure
+// HealthResponse represents the health check response structure.
 type HealthResponse struct {
 	Healthy bool   `json:"healthy"`
 	Reason  string `json:"reason"`
@@ -61,7 +61,9 @@ func main() {
 		if err == nil {
 			interval = parsedInterval
 		} else {
-			klog.ErrorS(err, "Failed to parse check interval, using default", "input", checkInterval, "default", defaultCheckInterval)
+			klog.ErrorS(err, "Failed to parse check interval, using default",
+				"input", checkInterval,
+				"default", defaultCheckInterval)
 		}
 	}
 
@@ -87,7 +89,7 @@ func main() {
 	// Main loop to check health and update condition
 	for {
 		// Check health
-		health, err := checkHealth(httpClient, checkEndpoint)
+		health, err := checkHealth(context.TODO(), httpClient, checkEndpoint)
 		if err != nil {
 			klog.ErrorS(err, "Health check failed", "endpoint", checkEndpoint)
 			// Report unhealthy on error
@@ -108,9 +110,18 @@ func main() {
 	}
 }
 
-// checkHealth performs an HTTP request to check component health
-func checkHealth(client *http.Client, endpoint string) (*HealthResponse, error) {
-	resp, err := client.Get(endpoint)
+// checkHealth performs an HTTP request to check component health.
+func checkHealth(ctx context.Context, client *http.Client, endpoint string) (*HealthResponse, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint, nil)
+	if err != nil {
+		return &HealthResponse{
+			Healthy: false,
+			Reason:  "RequestCreationError",
+			Message: fmt.Sprintf("Failed to create request for endpoint %s: %v", endpoint, err),
+		}, nil
+	}
+
+	resp, err := client.Do(req)
 	if err != nil {
 		return &HealthResponse{
 			Healthy: false,
@@ -146,7 +157,7 @@ func checkHealth(client *http.Client, endpoint string) (*HealthResponse, error) 
 	}, nil
 }
 
-// updateNodeCondition updates the node condition based on health check
+// updateNodeCondition updates the node condition based on health check.
 func updateNodeCondition(client kubernetes.Interface, nodeName, conditionType string, health *HealthResponse) error {
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		// Get the node
--- a/cmd/readiness-condition-reporter/main_test.go
+++ b/cmd/readiness-condition-reporter/main_test.go
@@ -50,7 +50,7 @@ func TestCheckHealth(t *testing.T) {
 			}
 
 			httpClient := &http.Client{Timeout: 1 * time.Second}
-			health, err := checkHealth(httpClient, endpoint)
+			health, err := checkHealth(context.TODO(), httpClient, endpoint)
 			if err != nil {
 				if !tt.expectError {
 					t.Errorf("checkHealth() error = %v", err)


cmd/readiness-condition-reporter/main.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	checkEndpoint := os.Getenv("CHECK_ENDPOINT")
+	if checkEndpoint == "" {
+		fmt.Fprintf(os.Stderr, "CHECK_ENDPOINT environment variable not set\n")
+		os.Exit(1)
+	}
+
+	checkInterval := os.Getenv("CHECK_INTERVAL")
+	interval := 30 * time.Second // Default interval
```
AvineshTripathi: we should use const instead of setting variable here

cmd/readiness-condition-reporter/main.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+
+		// Wait for next check
+		time.Sleep(interval)
+	}
+}
+
+// checkHealth performs an HTTP request to check component health
+func checkHealth(endpoint string) (*HealthResponse, error) {
+	resp, err := http.Get(endpoint)
```
AvineshTripathi: probably use timeouts

cmd/readiness-condition-reporter/main.go
```diff
@@ -0,0 +1,181 @@
...
+		return &HealthResponse{
+			Healthy: true,
+			Reason:  "EndpointOK",
+			Message: fmt.Sprintf("Endpoint reports ready at %s", endpoint),
+		}, nil
+	}
+
+	bodyBytes, err := io.ReadAll(resp.Body)
+	bodyString := ""
+	if err == nil {
```
AvineshTripathi, line 146: in case of err != nil message in `healthResponse` would be `Endpoint returned non-2xx status at %s:`

cmd/readiness-condition-reporter/main.go
```diff
@@ -0,0 +1,181 @@
...
+	for i, c := range node.Status.Conditions {
+		if string(c.Type) == conditionType {
+			node.Status.Conditions[i] = condition
+			found = true
+			break
+		}
+	}
+
+	if !found {
+		node.Status.Conditions = append(node.Status.Conditions, condition)
```
AvineshTripathi: rare case but there can be a case where node status is updated while we have a older snapshot here(https://github.com/kubernetes-sigs/node-readiness-controller/pull/57/files#diff-747dc8283e29f76dedbef915ee59d98f24683288bc0d4f270a5c08d7d4bddde2R128) and hence, we do an override. Instead we should use retry mechanism

scripts/build-and-publish-image.sh
Priyankasaggu11929: These TODOs can be removed now that we have `docker buildx` also working in the Prow job.

scripts/build-and-publish-image.sh
```diff
@@ -20,7 +20,19 @@ if [[ -z ${IMG_TAG:-} ]]; then
...
 
-IMG_TAG=${IMG_TAG} IMG_PREFIX=${IMG_PREFIX} make docker-buildx
+COMPONENT=${COMPONENT:-controller}
+
+if [[ "$COMPONENT" == "controller" ]]; then
+  echo "Building controller..."
+  IMG_TAG=${IMG_TAG} IMG_PREFIX=${IMG_PREFIX} make docker-buildx
+elif [[ "$COMPONENT" == "reporter" ]]; then
+  echo "Building reporter..."
+  IMG_TAG=${IMG_TAG} IMG_PREFIX=${IMG_PREFIX} make docker-buildx-reporter
```
Priyankasaggu11929:
```suggestion
  IMG_TAG=${IMG_TAG} IMG_PREFIX=${IMG_PREFIX}/node-readiness-controller make docker-buildx
elif [[ "$COMPONENT" == "reporter" ]]; then
  echo "Building reporter..."
  IMG_TAG=${IMG_TAG} IMG_PREFIX=${IMG_PREFIX}/node-readiness-reporter make docker-buildx-reporter
```

and then we can update cloudbuild.yaml like this:


```
❯ cat cloudbuild.yaml 
# See https://cloud.google.com/cloud-build/docs/build-config

options:
  substitution_option: ALLOW_LOOSE
steps:
  - name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20251211-3eba3d0954
    entrypoint: ./scripts/build-and-publish-image.sh
    env:
    - IMG_PREFIX=us-central1-docker.pkg.dev/k8s-staging-images/node-readiness-controller
    - COMPONENT=controller

  - name: gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20251211-3eba3d0954
    entrypoint: ./scripts/build-and-publish-image.sh
    env:
    - IMG_PREFIX=us-central1-docker.pkg.dev/k8s-staging-images/node-readiness-controller
    - COMPONENT=reporter
```

---

Q: Do we want to build both the controller and reporter image on every PR/commit merge?  
     (like we do with node-readiness-controller image right now)

     If so, should we do that with matching tags (which will basically be removing the if/else conditional in `/scripts/build-and-publish-image.sh` script and keep both image build in serial)?

     If no, I will test how to make cloudbuild.yaml do that (or we create a new cloudbuild-reporter.yaml)

Karthik-K-N: I think building reporter for each commit is not necessary, unless its cheap/fast we can do that, otherwise only on changes to cmd/readiness-condition-reporter directory we should build it.

Priyankasaggu11929: > I think building reporter for each commit is not necessary, unless its cheap/fast we can do that

The process is fast, but I honestly don't know how expensive or cheap it is in terms of resources.

But I agree - my opinion is also not to build it on every PR unless we foresee frequent changes (and if that happens in future, we can always flip it).

So, if we go with this decision, I think best is to make a new cloudbuild yaml for node-readiness-reporter image and trigger it from a postsubmit job on changes in the `cmd/readiness-condition-reporter` directory. 
We can keep the new cloudbuild yaml in some dedicated node-readiness-reporter directory or something.

Karthik-K-N: Yes, I agree to your points.

ajaysundark: IMO, we 'd eventually want to 'release' only artifacts from the release branches, as we dont want adoption from every patch as it'll be difficult to track. Having a smaller finite number of released artifacts could also help us drive bug fixes and set a clear guidance for adopters.

For example, etcd which has multiple utility components follow a manual release model, where the release script(s) are executed only from the release branches - https://goo.gl/Gxwysq.

cmd/readiness-condition-reporter/main.go
ajaysundark: Moved env variable keys and default values as const

cmd/readiness-condition-reporter/main.go
ajaysundark: added a 10s timeout

cmd/readiness-condition-reporter/main.go
ajaysundark: Thanks! improved error messages overall

cmd/readiness-condition-reporter/main.go
ajaysundark: agree conflicts can be possible. added retryonconflct.

scripts/build-and-publish-image.sh
ajaysundark: thanks. removed

scripts/build-and-publish-image.sh
ajaysundark: I accepted @Priyankasaggu11929 's suggestion for now. we could discuss and decide on the release cadence and published artifacts offline at slack

Priyankasaggu11929: > IMO, we 'd eventually want to 'release' only artifacts from the release branches, as we dont want adoption from every patch as it'll be difficult to track. Having a smaller finite number of released artifacts could also help us drive bug fixes and set a clear guidance for adopters.

that would be basically updating the regex entries under `branches:` here - https://github.com/kubernetes/test-infra/blob/26b399caacb20573e770331348fc44f7aaaae589/config/jobs/image-pushing/k8s-staging-node-readiness-controller.yaml#L9-L13

so, yes, we can update it after proper discussion.

PR #66, opened 2026-01-07
Added crd-ref-docs config and generated API docs

### LINKED ISSUE #5: Add auto-generated API reference documentation
As the project evolves, we have to make it accessible for someone wants to discover our CRD fields and understand what they do. Users who want to know "what can I configure?" need not have to dig through the Go code

**Proposed Solution**

Set up auto-generated API docs like **crd-ref-docs** (or other common tooling in kubernetes-sigs projects).

sreeram-venkitesh:
Closes #5 

For writing crd-ref-docs.yaml, I referred the same file in [gateway-api](https://github.com/kubernetes-sigs/gateway-api/blob/main/crd-ref-docs.yaml)

Currently, for running `make crd-ref-docs`, you need to first run
```
go install github.com/elastic/crd-ref-docs@latest
```

@ajaysundark We can discuss the path to which we can write the file. Some projects I referred, like `aws-load-balancer-controller` and `container-object-storage-interface` writes the api docs into the docs directory, while gateway-api has a [script](https://github.com/kubernetes-sigs/gateway-api/blob/03ded6d3983ad6911b76afd7521e15d346c580ba/hack/mkdocs/generate.sh) to install crd-ref-docs and [generate the docs into the website source](https://github.com/kubernetes-sigs/gateway-api/blob/03ded6d3983ad6911b76afd7521e15d346c580ba/Makefile#L202-L215) which can be accessed like [this](https://gateway-api.sigs.k8s.io/reference/1.4/spec/). In my opinion we can have something similar, but we can decide if it can be taken up as a follow up after the initial release once the website is also ready.

Tagging @Karthik-K-N also for API docs review.


docs/spec.md
```diff
@@ -0,0 +1,209 @@
...
+## readiness.node.x-k8s.io/v1alpha1
+
+Package v1alpha1 contains API Schema definitions for the  v1alpha1 API group.
+
+### Resource Types
+- [NodeReadinessRule](#nodereadinessrule)
+
+
+
+#### ConditionEvaluationResult
```
Karthik-K-N, line 16: Can this be reorderd to have SPEC first followed by STATUS?

sreeram-venkitesh: Do we want [NodeReadinessRuleSpec](https://github.com/kubernetes-sigs/node-readiness-controller/pull/66/files/58dae3455c3734f209afe1c52bba0c17dcab2e9d#diff-d7d183da3129291ef1210410b2898ac6a65a4e4de8ede9775601173545570f57R151) (line 151) to come first and then [NodeReadinessRulesStatus](https://github.com/kubernetes-sigs/node-readiness-controller/pull/66/files/58dae3455c3734f209afe1c52bba0c17dcab2e9d#diff-d7d183da3129291ef1210410b2898ac6a65a4e4de8ede9775601173545570f57R171) (line 171) and then the rest everything?

Makefile
```diff
@@ -325,3 +325,12 @@ docs: ## Build the mdBook locally using the same script Netlify uses.
...
 	GO_VERSION=$(GO_VERSION) MDBOOK_VERSION=$(MDBOOK_VERSION) $(MDBOOK_SCRIPT) serve docs/book --open
+
+# generate CRD spec doc
+.PHONY: crd-ref-docs
+crd-ref-docs:
+	crd-ref-docs \
+		--source-path=${PWD}/api/v1alpha1/ \
+		--config=crd-ref-docs.yaml \
+		--renderer=markdown \
+		--output-path=${PWD}/docs/spec.md
```
ajaysundark: Agree to your suggestion for linking this to mdbook directly. how about out to `docs/book/src/reference/api-spec.md`? We already have a (dead) link for this at our site.

docs/spec.md
```diff
@@ -0,0 +1,209 @@
...
+
+_Appears in:_
+- [NodeEvaluation](#nodeevaluation)
+
+| Field | Description | Default | Validation |
+| --- | --- | --- | --- |
+| `type` _string_ | type corresponds to the Node condition type being evaluated. |  | MaxLength: 316 <br />MinLength: 1 <br /> |
+| `currentStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | currentStatus is the actual status value observed on the Node, one of True, False, Unknown. |  | Enum: [True False Unknown] <br /> |
+| `requiredStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | requiredStatus is the status value defined in the rule that must be matched, one of True, False, Unknown. |  | Enum: [True False Unknown] <br /> |
+| `satisfied` _boolean_ | satisfied indicates whether the CurrentStatus matches the RequiredStatus. |  |  |
```
ajaysundark: @Karthik-K-N cleaned up these fields recently, could you sync and recreate this page again?
On that note, how could we maintain the ref-docs to be auto-updated from our changes?

docs/spec.md
ajaysundark: looks like there is no config to declare a field order, but the generator just determines the order from discovered types. Maybe we could define our own order with templating, but dont think we would want to invest more time here. This works.. :)

Makefile
sreeram-venkitesh: Thanks! That path looks good. Tried this out locally and this is how it'll look in the mdbook website. I'll push this soon with the other changes.

image

docs/spec.md
sreeram-venkitesh: I've regenerated the doc and pushed the new changes. For autoupdating changes, we can probably add a CI job that fails if an API change PR doesn't have the updated ref docs. We can compare the existing api-spec.md file with a new run of `make crd-ref-docs` in the CI and block merging if they don't match. What do you think?

ajaysundark: Sounds good. That or a cron action sync to regenerate it at some long enough interval.  its not urgent, we could keep it in our backlog. 

Thanks again for handling this!

PR #68, opened 2026-01-08
Refactor NRR reconciler to simplify adding finalizer to the rule object

Karthik-K-N:
This PR does following things
1. Refactored the NRR reconciler to simplify the finalizer addition to rule object logic
2. Moved the helper functions to helper.go file
3. Added UT to finalizer addition logic.

### DIFF
--- a/internal/controller/helper.go
+++ b/internal/controller/helper.go
@@ -1,0 +1,117 @@
+package controller
+
+import (
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// nodeSelectorChanged checks if nodeSelector has changed.
+func nodeSelectorChanged(current, previous metav1.LabelSelector) bool {
+	// Compare matchLabels
+	if !stringMapEqual(current.MatchLabels, previous.MatchLabels) {
+		return true
+	}
+
+	// Compare matchExpressions
+	if len(current.MatchExpressions) != len(previous.MatchExpressions) {
+		return true
+	}
+
+	// Create maps for comparison
+	currentExprs := make(map[string]metav1.LabelSelectorRequirement)
+	for _, expr := range current.MatchExpressions {
+		key := fmt.Sprintf("%s-%s-%v", expr.Key, expr.Operator, expr.Values)
+		currentExprs[key] = expr
+	}
+
+	previousExprs := make(map[string]metav1.LabelSelectorRequirement)
+	for _, expr := range previous.MatchExpressions {
+		key := fmt.Sprintf("%s-%s-%v", expr.Key, expr.Operator, expr.Values)
+		previousExprs[key] = expr
+	}
+
+	for key := range currentExprs {
+		if _, exists := previousExprs[key]; !exists {
+			return true
+		}
+	}
+
+	return false
+}
+
+// stringMapEqual checks if two string maps are equal.
+func stringMapEqual(a, b map[string]string) bool {
+	if len(a) != len(b) {
+		return false
+	}
+
+	for k, v := range a {
+		if b[k] != v {
+			return false
+		}
+	}
+
+	return true
+}
+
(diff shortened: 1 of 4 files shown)


internal/controller/nodereadinessrule_controller.go
```diff
@@ -181,6 +157,32 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+	log := ctrl.LoggerFrom(ctx)
+
+	log.Info("Cleaning up taints for deleted rule", "rule", rule.Name)
+	if err := r.Controller.cleanupTaintsForRule(ctx, rule); err != nil {
+		log.Error(err, "Failed to cleanup taints for rule", "rule", rule.Name)
+		return ctrl.Result{RequeueAfter: time.Minute}, err
+	}
+
+	log.V(3).Info("Removing the rule from cache")
+	r.Controller.removeRuleFromCache(ctx, rule.Name)
```
ajaysundark, line 184: Should this be moved the last? If patch to remove finalizer fails below, retry wont have this rule in cache for next reconciliation.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -84,52 +86,30 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+	log = log.WithValues("ruleName", rule.Name)
+	ctx = ctrl.LoggerInto(ctx, log)
 
-			// Remove from cache after successful cleanup
-			r.Controller.removeRuleFromCache(ctx, rule.Name)
-		}
-		return ctrl.Result{}, nil
+	// Add finalizer first if not set to avoid the race condition between init and delete.
+	if finalizerAdded, err := ensureFinalizer(ctx, r.Client, rule, finalizerName); err != nil {
+		return ctrl.Result{}, err
```
ajaysundark, line 112: should this have a requeue / retry on error?

internal/controller/helper.go
```diff
@@ -0,0 +1,78 @@
...
+	"fmt"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	readinessv1alpha1 "sigs.k8s.io/node-readiness-controller/api/v1alpha1"
+)
+
+func ensureFinalizer(ctx context.Context, c client.Client, rule *readinessv1alpha1.NodeReadinessRule, finalizer string) (finalizerAdded bool, err error) {
```
ajaysundark: nit: could be a method on RuleReconciler where this is only applicable

internal/controller/nodereadinessrule_controller.go
Karthik-K-N: yeah, but removing a rule from cache is idempotent operation right? so calling it again should cause noharm.
Usually the finalizers should be removed at the end when every cleaning thing wrt to deleting object is done.

internal/controller/nodereadinessrule_controller.go
Karthik-K-N: The reconciler will always reque incase of non-nil error

internal/controller/helper.go
Karthik-K-N: Done, Thanks

internal/controller/nodereadinessrule_controller.go
ajaysundark: yeah, both orderings look fine. I was cautious about possible race-condition if any node-reconciler in flight is handling the rule.

Thinking of which node-reconciliation shouldn't really process a rule that's marked to be deleted. I just handled this here: https://github.com/kubernetes-sigs/node-readiness-controller/pull/70

Karthik-K-N: Yes, Thats perfect. Thanks

PR #79, opened 2026-01-11
refactor: enable metrics manifests

AvineshTripathi:
Enabled Prometheus and Cert-Manager in the Kustomize configuration to support secure metrics scraping and certificate management.

how to test locally:

1. create a kind cluster or any cluster
2. Install prometheus(I used https://github.com/prometheus-operator/kube-prometheus quickstart)
3. Install cert manager crds(kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.19.2/cert-manager.yaml)
4. Run `make deploy ENABLE_METRICS=true ENABLE_TLS=true`
4. expose grafana using (kubectl port-forward svc/grafana -n monitoring 3000)


config/rbac/metrics_reader_role.yaml
```diff
@@ -3,7 +3,19 @@ kind: ClusterRole
 metadata:
   name: metrics-reader
 rules:
-- nonResourceURLs:
-  - "/metrics"
+- apiGroups: [""]
```
ajaysundark: why does metrics-reader RBAC need these resource access?

AvineshTripathi: If we do not give these permissions, Prometheus is not able to configure pods as targets. here were the logs in Prometheus, and something similar happens for other resources

```
time=2026-01-11T11:53:19.320Z level=ERROR source=reflector.go:205 msg="Failed to watch" component=k8s_client_runtime logger=UnhandledError err="failed to list *v1.Pod: pods is forbidden: User \"system:serviceaccount:monitoring:prometheus-k8s\" cannot list resource \"pods\" in API group \"\" in the namespace \"nrr-system\"" reflector=pkg/mod/k8s.io/client-go@v0.34.1/tools/cache/reflector.go:290 type=*v1.Pod
```

internal/metrics/metrics.go
```diff
@@ -0,0 +1,61 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+)
+
+var (
+	// ActiveRules tracks the number of currently active NodeReadinessRules.
```
ajaysundark: Thanks for defining these metrics, they look good. Could you create a follow-up action-item to document these metrics for reference?

config/default/kustomization.yaml
```diff
@@ -18,20 +18,96 @@ resources:
 - ../rbac
 - ../manager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
-#- ../prometheus
+- ../prometheus
 # [METRICS] Expose the controller manager metrics service.
```
ajaysundark: all these dependencies make the install artifact (dist/install.yaml) a "heavy" bundle. I also see your thoughts on including these hard dependencies for readiness-controller which can be something fundamental in a cluster.

While the idea of "all-batteries-included" release sounds nice, it gets into managing this 'ordered' deployments - which is what this project aim to solve.

I agree with your thoughts that we could consider overlays to allow users to generate these manifests as 'opt-in features' to keep the core lean and easy to adapt - especially in managed provider setups.

config/default/kustomization.yaml
```diff
@@ -18,20 +18,96 @@ resources:
...
-# - metrics_service.yaml
+- metrics_service.yaml
 # [NETWORK POLICY] Protect the /metrics endpoint and Webhook Server with NetworkPolicy.
 # Only Pod(s) running a namespace labeled with 'metrics: enabled' will be able to gather the metrics.
 # Only CR(s) which requires webhooks and are applied on namespaces labeled with 'webhooks: enabled' will
 # be able to communicate with the Webhook Server.
 #- ../network-policy
+- ../certmanager
```
ajaysundark: As discussed, this PR will break test-e2e as these dependencies are not handled today.

If we go these in non-default overlays, it would continue to work, and you could add specific tests (or separate e2e suite) for the new functionality.

config/rbac/prometheus_role_binding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: prometheus-metrics-reader
+  namespace: nrr-system
+subjects:
+- kind: ServiceAccount
+  name: prometheus-k8s
```
ajaysundark: This hardcodes the namespace / SA name. Is this standard? I think we could document this as well.

config/rbac/metrics_reader_role.yaml
ajaysundark: I see, this seems to be required for Prometheus to discover the controller and creating service-monitor. Did you verify if `nonResourceURLs: ["/metrics"]` is no-longer needed and this role is sufficient for a scraper without it?

cmd/main.go
```diff
@@ -60,8 +63,15 @@ func main() {
...
 	var probeAddr string
 	var enableWebhook bool
+	var metricsSecure bool
+	var metricsCertDir string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
+	flag.BoolVar(&metricsSecure, "metrics-secure", false,
+		"If set, the metrics endpoint is served securely via HTTPS. "+
+			"Requires certificate and key.")
+	flag.StringVar(&metricsCertDir, "metrics-cert-dir", "",
```
ajaysundark, line 73: so does this default work when cert-manager isnt there? likely empty is fine, but see if you could verify behavior for the default flag.

config/rbac/metrics_reader_role.yaml
AvineshTripathi: Yes, it works without that because Prometheus is hitting the controller pod api and not kube api server

cmd/main.go
AvineshTripathi: yes it works fine

Makefile
```diff
@@ -223,6 +226,32 @@ ifndef ignore-not-found
...
+# Temporary directory for building manifests
+BUILD_DIR := $(ROOT_DIR)/bin/build
+
+# Internal target to build manifests in a temporary directory to keep the source config clean.
+# This prevents 'kustomize edit' from modifying your local git state.
+# Features (Metrics, TLS) are enabled by adding Kustomize Components to the temporary copy.
+# TODO: we can do better for prometheus metrics ports that are added by manager_prometheus_metrics.yaml
+.PHONY: build-manifests-temp
+build-manifests-temp: manifests $(KUSTOMIZE)
+	@mkdir -p $(BUILD_DIR)
```
ajaysundark, line 244: should we add `@set -e` for safety?

Makefile
```diff
@@ -208,10 +212,9 @@ docker-buildx-reporter: ## Build and push docker image for the reporter for cros
 	- $(CONTAINER_TOOL) buildx rm reporter-builder
 
 .PHONY: build-installer
-build-installer: manifests generate $(KUSTOMIZE) ## Generate a consolidated YAML with CRDs and deployment.
+build-installer: ## Generate a consolidated YAML with CRDs and deployment.
 	mkdir -p dist
-	cd config/manager && $(KUSTOMIZE) edit set image controller=${IMG_PREFIX}:${IMG_TAG}
-	$(KUSTOMIZE) build config/default > dist/install.yaml
+	$(MAKE) -s build-manifests-temp > dist/install.yaml
```
ajaysundark: Should this redirection be handled within the call?

Makefile
```diff
@@ -223,6 +226,32 @@ ifndef ignore-not-found
...
+	@if [ "$(ENABLE_METRICS)" = "true" ]; then \
+		cd $(BUILD_DIR)/config/default && $(KUSTOMIZE) edit add component ../prometheus; \
+		if [ "$(ENABLE_TLS)" = "true" ]; then \
+			cd $(BUILD_DIR)/config/default && $(KUSTOMIZE) edit add component ../certmanager && \
+			$(KUSTOMIZE) edit add component ../prometheus/tls; \
+		else \
+			cd $(BUILD_DIR)/config/prometheus && $(KUSTOMIZE) edit add patch --path manager_prometheus_metrics.yaml --kind Deployment --name controller-manager; \
+		fi; \
+	fi
+	@$(KUSTOMIZE) build $(BUILD_DIR)/config/default
```
ajaysundark: We dont need to club this in this PR, but I realized while handling release that our previous cleanup removed 'crds' from the manifest, and we are handling creating crds.yaml manually today.

We will need to refactor this a bit to create multiple artifacts as a followup.

internal/metrics/metrics.go
AvineshTripathi: ref : https://github.com/kubernetes-sigs/node-readiness-controller/issues/94

Makefile
AvineshTripathi: i think we set it globally(https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/Makefile#L78) so it's not needed

Makefile
AvineshTripathi: you mean we should add crds as part of https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R215 ?

ajaysundark: I just sent a short PR  https://github.com/kubernetes-sigs/node-readiness-controller/pull/95.

crds are also part of the artifacts bundle we publish as part of release. build-installer used to generate a single manifest including the crds and the deployment, but this was removed to favor cleaner install/uninstall paths.

ajaysundark: I just sent a short PR  https://github.com/kubernetes-sigs/node-readiness-controller/pull/95.

crds are also part of the artifacts bundle we publish as part of release. build-installer used to generate a single manifest including the crds and the deployment, but this was removed to favor cleaner install/uninstall paths.

Makefile
```diff
@@ -351,7 +384,6 @@ docs-serve: ## Serve mdBook locally.
 .PHONY: crd-ref-docs
 crd-ref-docs:
 	crd-ref-docs \
-		--source-path=${PWD}/api/v1alpha1/ \
```
ajaysundark, line 354: is this change intended?

AvineshTripathi: go catch thank you!

Makefile
```diff
@@ -240,13 +269,17 @@ uninstall: manifests $(KUSTOMIZE) ## Uninstall CRDs from the K8s cluster specifi
...
+deploy: build-manifests-temp ## Deploy controller to the K8s cluster. Use ENABLE_METRICS=true and ENABLE_TLS=true to enable features.
+	$(KUBECTL) apply -f $(BUILD_DIR)/manifests.yaml
 
 .PHONY: undeploy
-undeploy: $(KUSTOMIZE) ## Undeploy controller from the K8s cluster specified in ~/.kube/config. Call with ignore-not-found=true to ignore resource not found errors during deletion.
-	$(KUSTOMIZE) build config/default | $(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f -
+undeploy: build-manifests-temp ## Undeploy controller from the K8s cluster. Use ENABLE_METRICS=true and ENABLE_TLS=true if they were enabled during deploy.
+	$(KUBECTL) delete --ignore-not-found=$(ignore-not-found) -f $(BUILD_DIR)/manifests.yaml
+
+.PHONY: debug-deploy
```
ajaysundark: is this for investigating what did build-manifests-temp do?
IMHO, this target is not required as build-installer also copies the manifests out?
and, also think debug-deploy is a misleading name.

cmd/main.go
```diff
@@ -60,8 +63,15 @@ func main() {
 	var enableLeaderElection bool
 	var probeAddr string
 	var enableWebhook bool
+	var metricsSecure bool
+	var metricsCertDir string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
+	flag.BoolVar(&metricsSecure, "metrics-secure", false,
```
Karthik-K-N, line 70: How did you install the certificates? to test this flow.

config/certmanager/certificate.yaml
```diff
@@ -0,0 +1,24 @@
...
+  namespace: system
+spec:
+  commonName: nrr-metrics
+  dnsNames:
+  - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc
+  - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc.cluster.local
+  issuerRef:
+    kind: Issuer
+    name: selfsigned-issuer
+  secretName: metrics-server-cert
```
Karthik-K-N: Lets add the new line here and whichever file its missing at the end to keep it consistent across other files.

internal/metrics/metrics.go
```diff
@@ -0,0 +1,61 @@
...
+	ActiveRules = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "node_readiness_active_rules",
```
Karthik-K-N: happy with the name, but whats does active rule means? is it all the rules except dryRun rules? as of now we dont any way to have active and inactive rules. It can be `node_readiness_rules_total`

AvineshTripathi:
Hey @Karthik-K-N I have added steps to test locally in the description. Once the setup is done on the Grafana dashboard, you should be able to see the metrics from the controller. Let me know if something is missing.

internal/metrics/metrics.go
AvineshTripathi: I agree the name can be confusing; for now the metric is basically the length on the cacheMap (here: https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/changes#diff-b857080def3b2fecd3967ca35b1bdc29564dda0b064fa902fa24efcd55471899R441) so it includes all the rules, including the dryRun

AvineshTripathi: I should probably rename it to `node_readiness_rules_total`

Karthik-K-N: Yep, That should be good, Lets also add a comment to state it includes all the rules, Including rules with dry-runs

Karthik-K-N:
> Hey @Karthik-K-N I have added steps to test locally in the description. Once the setup is done on the Grafana dashboard, you should be able to see the metrics from the controller. Let me know if something is missing.

Thanks for the quick update, I will give a try .

cmd/main.go
AvineshTripathi: You dont have to install it manually; they are installed as part of the `make deploy` command with flags enabled. 

ref: 
- https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/changes#diff-4427771506cc17d573c7f557ec2f3dc01860e2da78be5daa0ecb8ad88ffc7aafR9
- https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/changes#diff-029a68426e5aab9ab474867d91fedbe5c9d494c0dca8a2fcb0274cc19172e452R5

config/default/kustomization.yaml
Priyankasaggu11929: +1 to keep monitoring stack as prometheus. 

Controller bundle should be lean as possible (contained to its own business logic), IMO

config/rbac/metrics_reader_role.yaml
Priyankasaggu11929: I find the existing `nonResourceURLs: ["/metrics"]` to be the case in most other projects.

@AvineshTripathi, do you have see examples of projects giving `metrics-reader` these wide permissions?

Priyankasaggu11929: > If we do not give these permissions, Prometheus is not able to configure pods as targets. here were the logs in Prometheus, and something similar happens for other resources
> 
> ```
> time=2026-01-11T11:53:19.320Z level=ERROR source=reflector.go:205 msg="Failed to watch" component=k8s_client_runtime logger=UnhandledError err="failed to list *v1.Pod: pods is forbidden: User \"system:serviceaccount:monitoring:prometheus-k8s\" cannot list resource \"pods\" in API group \"\" in the namespace \"nrr-system\"" reflector=pkg/mod/k8s.io/client-go@v0.34.1/tools/cache/reflector.go:290 type=*v1.Pod
> ```

I think create a `Role` rather than a `ClusterRole` would be sufficient here?

Like for reference - https://github.com/kubernetes/ingress-nginx/blob/main/deploy/prometheus/role.yaml

AvineshTripathi: While the ingress-nginx example uses a Role because Prometheus is deployed locally within their own namespace, our architecture requires a ClusterRole.

Since we follow the standard practice of deploying Prometheus into a dedicated monitoring namespace, it needs to 'reach across' boundaries to discover and scrape targets in other namespaces. A standard Role is restricted to its own namespace therefore, a ClusterRole is essential to provide the cross-namespace visibility required for cluster-wide monitoring.

Also to scope down the access I have create a roleBinding instead of ClusterRoleBinding(https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/changes/BASE..50a3a97c07670077ed72db8713346096a39429b0#diff-20ea6f9012b8991ff298f25bb01a912482502e674e540b55b6117ce920eb6990R2)

also wrt `nonResourceURLs` I think this is for API server endpoints, but in our case our pod exposes the metrics endpoints, so we do not need it. I couldn't find any official documentation for this, though. 

p.s. happy to be corrected here

Priyankasaggu11929: > Also to scope down the access I have create a roleBinding instead of ClusterRoleBinding(https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/changes/BASE..50a3a97c07670077ed72db8713346096a39429b0#diff-20ea6f9012b8991ff298f25bb01a912482502e674e540b55b6117ce920eb6990R2)

Ok, creating a RoleBinding makes sense. (Thanks for pointing it. I didn't notice it before.)

But I think it's not the responsibility of the controller to provide this ClusterRole here. It should be provided by the Monitoring stack itself (which actually is the case).

So, I followed the quickstart guide from https://github.com/prometheus-operator/kube-prometheus (from the local testing instructions in the PR description)

The kube-prometheus stack there, creates the following ClusterRole(s) by itself. I think we should use them instead

```
❯ kubectl get clusterrole prometheus-adapter -o yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"rbac.authorization.k8s.io/v1","kind":"ClusterRole","metadata":{"annotations":{},"labels":{"app.kubernetes.io/component":"metrics-adapter","app.kubernetes.io/name":"prometheus-adapter","app.kubernetes.io/part-of":"kube-prometheus","app.kubernetes.io/version":"0.12.0"},"name":"prometheus-adapter"},"rules":[{"apiGroups":[""],"resources":["nodes","namespaces","pods","services"],"verbs":["get","list","watch"]}]}
  creationTimestamp: "2026-01-24T07:28:34Z"
  labels:
    app.kubernetes.io/component: metrics-adapter
    app.kubernetes.io/name: prometheus-adapter
    app.kubernetes.io/part-of: kube-prometheus
    app.kubernetes.io/version: 0.12.0
  name: prometheus-adapter
  resourceVersion: "759"
  uid: 6d482752-0088-49d7-9aa3-42be3db6e4a7
rules:
- apiGroups:
  - ""
  resources:
  - nodes
  - namespaces
  - pods
  - services
  verbs:
  - get
  - list
  - watch
```

comes from - https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/prometheusAdapter-clusterRole.yaml

There's another one for RBAC around endpointslices as well -
https://github.com/prometheus-operator/kube-prometheus/blob/be3a410867333faf4de99cb9b2196c0e1a21b9cb/manifests/prometheusOperator-clusterRole.yaml#L114-L123
(which actually is more destructive in terms of permissions, but I didn't check other manifests, maybe something else might be there more suited).

Also, there's `prometheus-k8s` clusterrole as well that takes care of `nonResourceURLs` as well -
https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/prometheus-clusterRole.yaml
But you pointed correctly, as far as the controller is concerned, it doesn't need to provide RBAC for scraping api-server components raw metric endpoints.

So, IMO, controller code should not provide a Cluster level RBAC for use by an external component. 

It's supposed to be handled by the component itself that needs those permissions (in this case, kube-prometheus stack takes care of it, for example.)

So, I think we can drop this RBAC entirely from controller base config?

AvineshTripathi: I think we can remove the dependency of RBAC entirely and let the user configure it. Thank you, @Priyankasaggu11929, for pointing this out

AvineshTripathi: maybe for now keep the files there but disable it in kustomization.yaml

config/rbac/prometheus_role_binding.yaml
```diff
@@ -0,0 +1,13 @@
...
+  namespace: monitoring
+roleRef:
+  kind: ClusterRole
+  name: metrics-reader
```
Priyankasaggu11929: Based on https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/changes#r2723899802

ClusterRole reference will need an update here.

AvineshTripathi: do you think disabling it here(https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/changes#diff-3fd064f2f1027a954b035a6d5885b96cc5b483f04889d14fa840ace5e9c29c55R20) will be better for now just to keep some ref in the codebase for now to revisit if needed

config/rbac/metrics_reader_role.yaml
Priyankasaggu11929: If it is not needed, I think we should clean up.

But since @ajaysundark added the existing file, so, maybe let's wait for a final check from him.

@ajaysundark, what do you think?

config/rbac/prometheus_role_binding.yaml
Priyankasaggu11929: responded to above comment here - https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/changes#r2724047539

---

Also -

Just for my understanding (I don't know what's the best way to structure Kustomization files for each component) - would it be better to move this file under `config/prometheus/` just to keep all prometheus manifests in one place?

AvineshTripathi: will do once we decide if are keeping it or removing

config/rbac/metrics_reader_role.yaml
ajaysundark: sorry folks, I'm catching up. What's the `existing file` we are talking about?

Priyankasaggu11929: File - [config/rbac/metrics_reader_role.yaml](https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/files/50a3a97c07670077ed72db8713346096a39429b0#diff-f584cdb0d5f2ae8793279088ff8028e4007a9147c52a21874092025adf8dd71a)
(basically the one this ongoing thread is about)

context here - https://github.com/kubernetes-sigs/node-readiness-controller/pull/79#discussion_r2723863256 - basically this RBAC is already provided by kube-prometheus stack and so from controller code, we ahould clean it since its not needed.

ajaysundark: got it, thanks for the context, @Priyankasaggu11929.  This file comes from the default scaffolding of kubebuilder - https://github.com/kubernetes-sigs/kubebuilder/blob/ccb3bd37277f9e6db2cdca3a6b553994275db42e/pkg/plugins/common/kustomize/v2/scaffolds/internal/templates/config/rbac/metrics_reader_role.go#L41.

ajaysundark: You are setting up authn/authz with controller-runtime's `WithAuthenticationAndAuthorization` feature. This seems to require this ClusterRole with `nonResourceUrls: /metrics` permissions. [ref](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/metrics/filters#WithAuthenticationAndAuthorization)

more details here: http://book.kubebuilder.io/reference/metrics

config/rbac/prometheus_role_binding.yaml
ajaysundark: Looks like the best practice is to not provide the role-binding - https://book.kubebuilder.io/reference/metrics#granting-permissions-to-access-metrics.

I think we should not prefer to include this either as we dont know the service-account or namespace, and by providing a default role-binding we could also risk accidentally binding to the wrong / restricted environments. Morever the consumer may not even be using prom (can be using Datadog or other custom agents as well)

config/rbac/prometheus_role_binding.yaml
ajaysundark: replied to [this at the wrong discussion](https://github.com/kubernetes-sigs/node-readiness-controller/pull/79/changes/BASE..50a3a97c07670077ed72db8713346096a39429b0#r2726461043) - could we not provide the role-binding? this has risks of user accidentally deploying the role into a wrong environment / namespace as this file has assumptions on the consumer.

ajaysundark: could we rather have this documented on 'how to install with secure metrics'?

config/rbac/metrics_reader_role.yaml
AvineshTripathi: > You are setting up authn/authz with controller-runtime's WithAuthenticationAndAuthorization feature.
we already have that here https://github.com/kubernetes-sigs/node-readiness-controller/blob/f579d8221beeaefa6668c1dd23723464086135a9/config/rbac/metrics_auth_role.yaml#L9

wrt to `nonResourceUrls`  i am still confused. My understanding about that endpoint is that it is apiServer endpoint and not controller endpoint

config/prometheus/kustomization.yaml
```diff
@@ -5,7 +5,7 @@ resources:
...
 # Additionally, ensure that you uncomment the [METRICS WITH CERTMANAGER] patch under config/default/kustomization.yaml
 # to mount the "metrics-server-cert" secret in the Manager Deployment.
-#patches:
-#  - path: monitor_tls_patch.yaml
-#    target:
-#      kind: ServiceMonitor
+patches:
+  - path: monitor_tls_patch.yaml
+    target:
+      kind: ServiceMonitor
```
ajaysundark: Is this intended or slipped from your testing? seems like there is no such patch at this path?

AvineshTripathi: it is in older commits

config/rbac/metrics_reader_role.yaml
ajaysundark: > kube-prometheus-stack takes care of it

I agree to @Priyankasaggu11929 here. Service discovery is provided by prometheus installation. For eg: https://github.com/prometheus-community/helm-charts/blob/main/charts%2Fkube-prometheus-stack%2Ftemplates%2Fprometheus%2Fclusterrole.yaml#L13 sets up these permissions.

We should provide only rbac for our controller requirements

config/certmanager/certificate.yaml
```diff
@@ -0,0 +1,24 @@
+# The following manifests contain a self-signed issuer CR and a certificate CR.
+# More document can be found at https://docs.cert-manager.io
+apiVersion: cert-manager.io/v1
+kind: Issuer
+metadata:
+  name: selfsigned-issuer
+  namespace: system
```
ajaysundark, line 7: is this suppose to be system namespace?

config/prometheus/monitor.yaml
ajaysundark, line 3: IMO, this could also be excluded as this adds dependency to prometheus operator. I expect the node-readiness-controller to be one of the initial pieces the cluster admin will be installing during cluster bootstrap, and adding dependencies will make it a bad UX.

config/certmanager/certificate.yaml
ajaysundark: this aligns with kubebuilder conventions.

config/prometheus/monitor.yaml
AvineshTripathi: @ajaysundark this is also part of the default kubebuilder layout

Issue #83, opened 2026-01-13
delete rule for concurrent nodes can end up with stuck taint


ajaysundark:
1. Create nodes with multiple worker nodes (3) 
2. Rule adds readiness-taint on worker nodes.
3. Delete Rule triggered - finalizer starts cleanup of taints
4. Finalizer removes taints from both nodes (nrr-test-worker2 and nrr-test-worker3) 
5. Node controller reconciles nrr-test-worker2 (triggered by the taint removal)
> Processing node against rules ... node: nrr-test-worker2, ruleCount: 1
6. Since the rule is still in cache, the node reconciliation sees the rule (ruleCount: 1), evaluates it, finds conditions not satisfied, and adds the taint back:
>  Adding taint ... node: nrr-test-worker2
7. There was a second reconciliation, but rule was removed from the cache, and previous added taint was stuck.
> Processing node against rules ... ruleCount: 0

---
**Bug**: The issue was that concurrent node-reconciliation during rule-deletion should see the 'deletionTimestamp' on the rule and skip processing it. We added fix for this in https://github.com/kubernetes-sigs/node-readiness-controller/pull/70, but since delete-rule flow doesnt update the cache, the parallel node-reconciliation escaped this check.

PR #84, opened 2026-01-13
fix rule deletion race condition with node-reconciliation

ajaysundark:
If there's a delete rule reconciliation, the finalizer triggers taint removal. This node update triggered node-reconciliation, which sometimes tried to re-add the taint when working with stale cache.

The patch here is that rule cache update should happen as the first step in `reconcileDelete` before `cleanupTaintsForRule`, so any further node reconciliation works with the latest update. 

Fixes https://github.com/kubernetes-sigs/node-readiness-controller/issues/83


Karthik-K-N (commented):
/lgtm

Thanks

PR #95, opened 2026-01-20
build-installer to create crd and install manifests

ajaysundark:
include manifest for crds.yaml


Karthik-K-N (commented):
/lgtm

Thank you.

PR #112, opened 2026-01-26
feat: Restrict NodeReadinessRuleSpec.Taint to "readiness.k8s.io/" prefix

### LINKED ISSUE #106: Restrict NodeReadinessRuleSpec.Taint to "readiness.k8s.io/" prefix
Allow only rule spec creation with prefix "readiness.k8s.io/".
This is needed to have separation of controller managed taints from other taints in the cluster.

This can be enforced with kubebuilder:validation:pattern. ref: https://book-v3.book.kubebuilder.io/reference/markers/crd-validation

sats-23:
## Description
-Enforces a naming convention for taints managed by the NodeReadinessRule
-`spec.taint.key` values must now be prefixed with readiness.k8s.io/
-Added a CEL validation rule to the CRD schema to enforce prefix
-Updated unit tests to reflect new naming standard 
-Updated use guide to document this prefix 

## Related Issue
Fixes #106 

## Type of Change
/kind feature

## Testing
-Create a rule with invalid key - fails validation (without prefix)
-Create a rule with a valid key - passes validation (with prefix)

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
spec.taint.key values must now be prefixed with readiness.k8s.io/
```
Doc #(issue)

### DIFF
--- a/docs/book/src/user-guide/concepts.md
+++ b/docs/book/src/user-guide/concepts.md
@@ -18,9 +18,9 @@ When a rule is created, the controller continuously watches all matching nodes. 
 Node Readiness Controller uses the `readiness.k8s.io` domain for taints and annotations that it owns. All `spec.taint.key` values in `NodeReadinessRule` must start with the `readiness.k8s.io/` prefix; this is enforced by the CRD schema.
 
 Typical taint keys look like:
-- `readiness.k8s.io/NetworkReady`
-- `readiness.k8s.io/StorageReady`
-- `readiness.k8s.io/<ComponentName>`
+- `readiness.k8s.io/cni.example.net/network-not-ready`
+- `readiness.k8s.io/csi.vendor.com/storage-driver-not-ready`
+- `readiness.k8s.io/<dns.subdomain>/<component-name>`
 
 The segment after `readiness.k8s.io/` should describe the dependency or subsystem whose readiness is being guarded (for example, a CNI plugin, storage backend, or security agent). Treat this domain as reserved for the controller and closely related components, and avoid reusing it for unrelated taints.
 


ajaysundark:
@Karthik-K-N Do you have time for this review?

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -70,7 +101,7 @@ type NodeReadinessRuleSpec struct {
 	// on Nodes that meet the defined condition criteria.
 	//
 	// +required
-	Taint corev1.Taint `json:"taint,omitempty,omitzero"`
```
Karthik-K-N, line 77: I think you have created local copy of corev1.Taint just to add the validation? Instead can we use the CEL validation and still use corev1.Taint?

Also lets update the validation webhook to have this check.

sats-23: I have migrated the validation logic to CEL rules that enforce directly on CRD. PTAL!

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -70,6 +70,10 @@ type NodeReadinessRuleSpec struct {
 	// on Nodes that meet the defined condition criteria.
 	//
 	// +required
+	// +kubebuilder:validation:XValidation:rule="self.key.startsWith('readiness.k8s.io/')",message="taint key must start with 'readiness.k8s.io/'"
+	// +kubebuilder:validation:XValidation:rule="self.key.size() >= 17 && self.key.size() <= 253",message="taint key length must be between 17 and 253 characters"
+	// +kubebuilder:validation:XValidation:rule="!has(self.value) || self.value.size() <= 63",message="taint value length must be at most 63 characters"
+	// +kubebuilder:validation:XValidation:rule="self.effect in ['NoSchedule', 'PreferNoSchedule', 'NoExecute']",message="taint effect must be one of 'NoSchedule', 'PreferNoSchedule', 'NoExecute'"
```
Karthik-K-N, line 80: This looks good. Thanks

internal/controller/node_controller_test.go
```diff
@@ -566,9 +567,9 @@ var _ = Describe("Node Controller", func() {
 						{Type: "StatusTestCondition", RequiredStatus: corev1.ConditionTrue},
 					},
 					Taint: corev1.Taint{
-						Key:    "status-test-taint",
+						Key:    "readiness.k8s.io/status-test-taint",
 						Effect: corev1.TaintEffectNoSchedule,
-						Value:  "pending",
+						Value:  val,
```
Karthik-K-N: why this change?

internal/webhook/nodereadinessgaterule_webhook.go
```diff
@@ -93,9 +95,36 @@ func (w *NodeReadinessRuleWebhook) validateSpec(spec readinessv1alpha1.NodeReadi
 	taintField := specField.Child("taint")
 	if spec.Taint.Key == "" {
 		allErrs = append(allErrs, field.Required(taintField.Child("key"), "taint key cannot be empty"))
+	} else {
```
Karthik-K-N: Apologies, I think with CEL, Nowhere it could skip the validation there so webhook might not be needed. We can remove.

api/v1alpha1/nodereadinessrule_types.go
Karthik-K-N: Also lets document it somewhere in the book or doc that the `readiness.k8s.io/` should be the prefix for taints, User should be aware of it.

sats-23: Sure, will do

internal/webhook/nodereadinessgaterule_webhook.go
sats-23: I was thinking what in case of conflict detection, like, if a rule A already exists and rule B is applied on the same set of nodes (same taint), then individually with CEL, rule B may pass but cause conflict in the controller? 
In case, this is invalid, I can remove it

internal/controller/node_controller_test.go
sats-23: Just extracted it for consistency

internal/webhook/nodereadinessgaterule_webhook.go
```diff
@@ -93,9 +95,36 @@ func (w *NodeReadinessRuleWebhook) validateSpec(spec readinessv1alpha1.NodeReadi
...
+		// Validate key prefix
+		if !strings.HasPrefix(spec.Taint.Key, "readiness.k8s.io/") {
+			allErrs = append(allErrs, field.Invalid(taintField.Child("key"), spec.Taint.Key, "taint key must start with 'readiness.k8s.io/'"))
+		}
+		// Validate key length
+		if len(spec.Taint.Key) < 17 || len(spec.Taint.Key) > 253 {
```
ajaysundark: nit: <17 length check feels redundant as you already have the prefix check above.

internal/webhook/nodereadinessgaterule_webhook.go
ajaysundark: IIUC, @Karthik-K-N's suggestion was that field validations in webhook is redundant as you already added the CEL check at API. The conflict check for rules are already in place and they need not be updated for the scope of this PR.

sats-23: Done, PTAL

internal/webhook/nodereadinessgaterule_webhook.go
sats-23: Done, PTAL

internal/controller/nodereadinessrule_controller.go
```diff
@@ -281,6 +280,30 @@ func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, ru
 	rule.Status.ObservedGeneration = rule.Generation
 	rule.Status.AppliedNodes = appliedNodes
 
+	// Create a map for efficient lookup of applied nodes
```
ajaysundark: @sats-23 This doesnt seem related to this PR.

If this intended change for a different need, could you create a separate PR with context?

ajaysundark:
> Fixes #106

@sats-23 We are now trying PR template for all the pull-requests.
Could you also update the description to abide the template when you get a chance? Thanks!

Karthik-K-N (commented):
Changes LGTM, Thanks for working on this.
When you get time, Please fix the failing tests.

docs/book/src/user-guide/concepts.md
```diff
@@ -13,6 +13,17 @@ A rule specifies:
...
 When a rule is created, the controller continuously watches all matching nodes. If a node does not satisfy the required conditions, the controller ensures the configured taint is present, preventing the scheduler from assigning new pods to that node.
 
+### Readiness Domain and Taint Keys
+
+Node Readiness Controller uses the `readiness.k8s.io` domain for taints and annotations that it owns. All `spec.taint.key` values in `NodeReadinessRule` must start with the `readiness.k8s.io/` prefix; this is enforced by the CRD schema.
+
+Typical taint keys look like:
+- `readiness.k8s.io/NetworkReady`
+- `readiness.k8s.io/StorageReady`
+- `readiness.k8s.io/<ComponentName>`
```
ajaysundark: Few suggestions on this:
1. we could use small cases for taint-keys. Though there's no API restriction for this, it seems general practice to use lower case here. eg: https://kubernetes.io/docs/reference/labels-annotations-taints/
2. As best-practices (and not confuse with 'well-defined' taints in K8s), I think we should start using a sub-domain in the managed taints. I know there are existing usage in the project that conflicts with this, and that's something we need to find time to clean up 
3. nit: taint api gives the 'repel' semantic in K8s, so I realized it reads better as "negative constraint" (what is not ready vs ready)

```suggestion
- `readiness.k8s.io/cni.example.net/network-not-ready`
- `readiness.k8s.io/csi.vendor.com/storage-driver-not-ready`
- `readiness.k8s.io/<dns.subdomain>/<component-name>`
```

PR #113, opened 2026-01-26
enhancement: Dockerfiles and Go Mod version

sats-23:
What's included?

-Updated base image to latest Go 1.25
-Optimized copy steps into single layer
-Added required DOCKER_BUILDKIT=1 flag for builds

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,14 +13,13 @@
 # limitations under the License.
 
 # Build the manager binary
-FROM --platform=$BUILDPLATFORM golang:1.24 AS builder
+FROM --platform=$BUILDPLATFORM golang:1.25 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /workspace
 # Copy the Go Modules manifests
-COPY go.mod go.mod
-COPY go.sum go.sum
+COPY go.mod go.sum ./
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 RUN go mod download
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/node-readiness-controller
 
-go 1.24.0
+go 1.25
 
 require (
 	github.com/onsi/ginkgo/v2 v2.27.2
--- a/Dockerfile.reporter
+++ b/Dockerfile.reporter
@@ -1,11 +1,10 @@
-FROM --platform=$BUILDPLATFORM golang:1.24 AS builder
+FROM --platform=$BUILDPLATFORM golang:1.25 AS builder
 ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /workspace
 
-COPY go.mod go.mod
-COPY go.sum go.sum
+COPY go.mod go.sum ./
 
 RUN go mod download
 
--- a/Makefile
+++ b/Makefile
@@ -172,7 +172,7 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
-	$(CONTAINER_TOOL) build -t ${IMG_PREFIX}:${IMG_TAG} .
+	DOCKER_BUILDKIT=1 $(CONTAINER_TOOL) build -t ${IMG_PREFIX}:${IMG_TAG} .
 
 .PHONY: docker-push
 docker-push: ## Push docker image with the manager.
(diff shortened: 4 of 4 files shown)


Dockerfile
```diff
@@ -13,14 +13,13 @@
 # limitations under the License.
 
 # Build the manager binary
-FROM --platform=$BUILDPLATFORM golang:1.24 AS builder
```
ajaysundark, line 16: This is needed to support multi-platform builds.

sats-23: image
The platform var is not injected as the docker build is not using the buildkit

https://github.com/kubernetes-sigs/node-readiness-controller/blob/21d133a8ed94ac18512da2578e48484b5a816f67/Makefile#L175

Hence, I thought having the docker-buildx target can be used for building multi-platform images and for standard docker-build, let docker assume the machine's platform.

Do let me know your thoughts!

ajaysundark: I think I made these changes [commit#](https://github.com/kubernetes-sigs/node-readiness-controller/commit/c518d9d491f9b9ddecbe262e4001deb870a9c42b) as the CI job was timing out.
AIUI, Go's cross-compilation is way faster and preferred for multiplatform builds. 

Instead of removing it, can we rather create the Dockerfile.cross on the fly and use it for `docker-buildx` as originally handled by @Priyankasaggu11929?
Alternatively, we can enforce buildkit as a requirement with DOCKER_BUILDKIT=1 as hinted by the docker error.

sats-23: Sure, let's stick to enforcing the buildkit

Priyankasaggu11929: @sats-23, just confirming, with the current changes in the PR - are you able to fix the errors you were getting in the standard `docker-build` make target?

If yes, then LGTM.

Otherwise, let's address it?

sats-23: Yep, absolutely, with DOCKER_BUILDKIT=1 , things work fine!

Priyankasaggu11929: got it, Though, I don't see in the code where we enforced using `DOCKER_BUILDKIT=1` in the PR? Should we add it?

sats-23: I suggest yes, it makes sense as it is mandatory to get it build process going

sats-23: I have added the enforcement into the Makefile

.dockerignore
```diff
@@ -1,11 +1,18 @@
...
 !go.mod
 !go.sum
+
+# Re-include source directories
+!api/
+!api/**
+!cmd/
+!cmd/**
+!internal/
+!internal/**
```
Priyankasaggu11929: I feel either we should document the contents of the  `.dockerignore` file somewhere properly or revert this block to how it was done before.

Because, let's say in future if we have more code directories that we want to be included in, that will require updating `.dockerignore` file as well and I feel it is very easy to miss that (and could be time consuming to debug because of being a hidden file)

sats-23: Right, I shall document it and simplify the .dockerignore

sats-23: Done

README.md
```diff
@@ -72,6 +72,26 @@ Find a more detailed walkthrough of setting up Node Readiness Controller in your
...
+
+**How it works:**
+The Docker CLI automatically checks for `.dockerignore` in the root directory. Files matching the patterns in this file are excluded from the context sent to the Docker daemon.
+
+Commonly ignored files in this project include:
+- `bin/`, `config/`, `dist/`: Local build artifacts.
+- `docs/`, `examples/`: Documentation and examples not needed in the runtime image.
+- `test/`: Test code.
+- Hidden files like `.git`, `.devcontainer`.
+
```
Priyankasaggu11929: @ajaysundark, where do you suggest this content goes? 

Do we have some contributing guide somewhere for this repo? I think this should be part of that and not README

.dockerignore
```diff
@@ -1,11 +1,27 @@
...
+OWNERS
+PROJECT
+SECURITY*
+demo.tape
 
-# Re-include Go module files
-!go.mod
-!go.sum
+# Tests
+**/*_test.go
```
Priyankasaggu11929: @ajaysundark, could you please review this part? I haven't used dockerignore files until now.

IMO, we don't need this much pruning here, because our Dockerfile is defined as multi-stage and so, the final container only gets the built manager binary, and that's it.

So, even if this ends up in the builder stage, it doesn't affect (unless I am missing something)

.dockerignore
ajaysundark: I agree with @Priyankasaggu11929  on maintainability part. If this change is not required, we could leave this as-is. Even if these aspects are documented, .dockerignore is easy to miss.

README.md
ajaysundark: I agree with the need for a contributing guide*, we have a Contributing.md in the root of the repo.

But thinking about this change, now I feel like reverting the dockerignore is safer for long-run. @sats-23 can we scope this PR for fixing only your docker-build and the go update?

*I'm pretty sure we also documented go updates somewhere as well! :!

.dockerignore
ajaysundark: These just came from kubebuilder https://github.com/kubernetes-sigs/kubebuilder/blob/master/docs/book/src/cronjob-tutorial/testdata/project/Dockerfile#L14.

I agree with optimizing these is not worth the effort.

sats-23: Done, PTAL

README.md
sats-23: Done PTAL

.dockerignore
sats-23: Done, PTAL

go.mod
```diff
@@ -1,11 +1,10 @@
 module sigs.k8s.io/node-readiness-controller
 
-go 1.24.0
+go 1.25
```
ajaysundark, line 3: @sats-23 could you also check and bump the golang version for devcontainer and golangci-kal for consistency? Thx

PR #117, opened 2026-02-02
feat: auto tolerate daemonsets with MAP

### LINKED ISSUE #7: feat: Auto tolerate DaemonSets with mutating admission controller
When requirements change or  a new NodeReadinessRule with a new taint need to be managed, existing components (DaemonSets) don't tolerate it. This require the admin to manually update every DaemonSet manifest to add the toleration. This is not great for operational ergonomics as in a typical enterprise setup the ownership are spread across different teams.

**Proposed Solution**

For better UX, an *optional* mutating admission policy could automatically inject tolerations for `readiness.k8s.io/*` taints into DaemonSets. 

**How it works**

- Watch DaemonSet create/update operations
- Automatically add tolerations for all `NoSchedule` taints with `readiness.k8s.io/` prefix 
- Disabled by default (separate deployment from main controller)

This will ensure no manual manifest updates are required when adding new readiness rules, and guarantee safer operations of critical daemon-sets during project evolution

pehlicd:
## Description

This PR implements automatic DaemonSet toleration injection for `readiness.k8s.io/*` taints using MutatingAdmissionPolicy with ConfigMap parameter resource and automated sync via the existing controller.

**Key features:**
- Zero additional infrastructure (no webhook server needed)
- ConfigMap-based parameter resource for dynamic toleration data
- Extended existing [RuleReconciler](cci:2://file:///Users/pehli/projects/node-readiness-controller/internal/controller/nodereadinessrule_controller.go:62:0-66:1) to automatically sync taints → ConfigMap
- CEL-based mutation logic
- Opt-out support via annotation

## Related Issue
Fixes #7 

## Type of Change

/kind feature

## Testing
Tested in local kind cluster.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Added optional MutatingAdmissionPolicy for automatic DaemonSet toleration injection. When enabled, DaemonSets automatically receive tolerations for all readiness.k8s.io/* taints, ensuring they can schedule on nodes during readiness transitions. Requires Kubernetes 1.32+ with MutatingAdmissionPolicy feature gate enabled.
```
Doc #7


config/admission-policy/binding.yaml
```diff
@@ -0,0 +1,15 @@
+---
+# MutatingAdmissionPolicyBinding binds the policy to the ConfigMap parameter
+apiVersion: admissionregistration.k8s.io/v1alpha1
```
Priyankasaggu11929:
```suggestion
apiVersion: admissionregistration.k8s.io/v1beta1
```

Ref: https://kubernetes.io/docs/reference/access-authn-authz/mutating-admission-policy/

Since Kubernetes 1.34, `MutatingAdmissionPolicy|MutatingAdmissionPolicyBinding` are available in v1beta1 API

config/admission-policy/policy.yaml
```diff
@@ -0,0 +1,71 @@
+---
+# MutatingAdmissionPolicy for automatic DaemonSet toleration injection
+# Reads taint keys from a ConfigMap parameter resource
+# Requires: MutatingAdmissionPolicy feature enabled in the cluster
+apiVersion: admissionregistration.k8s.io/v1alpha1
```
Priyankasaggu11929:
```suggestion
apiVersion: admissionregistration.k8s.io/v1beta1
```

Same as https://github.com/kubernetes-sigs/node-readiness-controller/pull/117/changes#r2779309765

docs/admission-policy.md
```diff
@@ -0,0 +1,54 @@
...
+
+The MutatingAdmissionPolicy approach uses Kubernetes's native admission control mechanism with CEL (Common Expression Language) to inject tolerations **without running a webhook server**. This provides a simpler, more declarative alternative to the webhook-based approach.
+
+## Requirements
+
+> [!IMPORTANT]
+> MutatingAdmissionPolicy is needed to be enabled in the cluster.
+
+- Feature gate: `MutatingAdmissionPolicy=true`
+- Runtime config: `admissionregistration.k8s.io/v1alpha1=true`
```
Priyankasaggu11929:
```suggestion
- Runtime config: `admissionregistration.k8s.io/v1beta1=true`
```
same, ref: https://kubernetes.io/docs/reference/access-authn-authz/mutating-admission-policy/

internal/controller/nodereadinessrule_controller.go
```diff
@@ -166,6 +166,12 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{RequeueAfter: time.Minute}, err
 	}
 
+	// Sync taints to ConfigMap for MutatingAdmissionPolicy
+	if err := r.Controller.syncTaintsConfigMap(ctx); err != nil {
+		log.Error(err, "Failed to sync taints configmap", "rule", rule.Name)
+		// Don't fail reconciliation for this - log and continue
+	}
```
Priyankasaggu11929, line 182: we need to add a matching block in `reconcileDelete` func as well, like below:

```
+       // Sync taints to ConfigMap after removal
+       if err := r.Controller.syncTaintsConfigMap(ctx); err != nil {
+               log.Error(err, "Failed to sync taints configmap after deletion", "rule", rule.Name)
+               // Don't fail reconciliation for this - log and continue
+       }
```

Currently, if I create a node-readiness-rule, it updates the `taints-keys` data in `readiness-taints` configmap.

But If I delete a node-readiness-rule, the configmap is not being updated - which should be the case.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -671,6 +677,75 @@ func (r *RuleReadinessController) cleanupNodesAfterSelectorChange(ctx context.Co
...
+	// List all NodeReadinessRules
+	var ruleList readinessv1alpha1.NodeReadinessRuleList
+	if err := r.List(ctx, &ruleList); err != nil {
+		return fmt.Errorf("failed to list NodeReadinessRules: %w", err)
+	}
+
+	// Extract unique taint keys with readiness.k8s.io/ prefix and NoSchedule effect
+	taintKeysSet := make(map[string]struct{})
+	for _, rule := range ruleList.Items {
+		if rule.Spec.Taint.Key != "" &&
```
Priyankasaggu11929, line 783:
```suggestion
	for _, rule := range ruleList.Items {
        // Skip rules that are being deleted
        if !rule.DeletionTimestamp.IsZero() {
                   continue
        }	
		if rule.Spec.Taint.Key != "" &&
```

we need to skip node readiness rules that have a DeletionTimestamp set during ConfigMap sync^ 

part of what I suggested here - https://github.com/kubernetes-sigs/node-readiness-controller/pull/117/changes#r2779315589

internal/controller/nodereadinessrule_controller.go
```diff
@@ -671,6 +677,75 @@ func (r *RuleReadinessController) cleanupNodesAfterSelectorChange(ctx context.Co
...
+	if apierrors.IsNotFound(err) {
+		// Create new ConfigMap
+		log.Info("Creating readiness-taints ConfigMap", "taintCount", len(taintKeys))
+		if err := r.Create(ctx, cm); err != nil {
+			return fmt.Errorf("failed to create configmap: %w", err)
+		}
+	} else {
+		// Update existing ConfigMap
+		log.V(1).Info("Updating readiness-taints ConfigMap", "taintCount", len(taintKeys))
+		patch := client.MergeFrom(existingCM)
```
Priyankasaggu11929:
```suggestion
		patch := client.MergeFrom(existingCM.DeepCopy())
```
we need to copy the configmap before creating a patch, otherwise currently we get an empty `taints-key` data in the configmap 

(I tested it locally)

This is a bug

pehlicd: Thanks for finding out, it is now fixed.

internal/controller/nodereadinessrule_controller.go
pehlicd: done!

internal/controller/nodereadinessrule_controller.go
pehlicd: yes definitely agree, sync added to the `reconcileDelete` func.

docs/admission-policy.md
pehlicd: done

config/admission-policy/policy.yaml
pehlicd: done

config/admission-policy/binding.yaml
pehlicd: done

internal/controller/nodereadinessrule_controller.go
```diff
@@ -90,6 +90,8 @@ func (r *RuleReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager)
 // +kubebuilder:rbac:groups=readiness.node.x-k8s.io,resources=nodereadinessrules,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=readiness.node.x-k8s.io,resources=nodereadinessrules/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=readiness.node.x-k8s.io,resources=nodereadinessrules/finalizers,verbs=update
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups="",resources=nodes,verbs=get;list;watch;create;update;patch;delete
```
pehlicd: Although we had nodes rbac roles, it was missing in the inline comments so I added them.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -90,6 +90,8 @@ func (r *RuleReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager)
 // +kubebuilder:rbac:groups=readiness.node.x-k8s.io,resources=nodereadinessrules,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=readiness.node.x-k8s.io,resources=nodereadinessrules/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=readiness.node.x-k8s.io,resources=nodereadinessrules/finalizers,verbs=update
+// +kubebuilder:rbac:groups="",resources=configmaps,verbs=get;list;watch;create;update;patch;delete
```
pehlicd: > In addition to those, how do you suggest handling RBAC for the controller to be able to update taints-key data in the readiness-taints configmap?
The PR currently miss that piece.

@Priyankasaggu11929 regarding to your comment above I added this inline kubebuilder comments and regenerated manifests.

internal/controller/nodereadinessrule_controller.go
Priyankasaggu11929: @pehlicd, I looked into it further.

Firstly, the node resource RBAC  added above - ~~I don't think we need them at all for this PR's purpose.~~ (correction - they're already added in the node controller here - https://github.com/kubernetes-sigs/node-readiness-controller/blob/354b1510874b623bc31a1f9ad75ebf04ced7fd0b/internal/controller/node_controller.go#L84-L86).
So, Let's remove the redundant markers from here.

What I realised is we don't need the above cluster-scoped configmap RBAC as well -
because, we already have a namespace-scoped RBAC role for configmaps (leader-election-role) here which is better IMO in terms of surface area of permissions and serves our purpose - https://github.com/kubernetes-sigs/node-readiness-controller/blob/354b1510874b623bc31a1f9ad75ebf04ced7fd0b/config/rbac/leader_election_role.yaml#L9-L21

The only problem is that currently the (controller-runtime's informer) cache is trying to watch configmaps at the cluster scope (see the error below again) but our namespaced role only provide permissions for nrr-system namespace, so I suggest we configure the cache to only watch configmaps in our required namespace.

```
❯ kubectl logs -n nrr-system 
   deployment/nrr-controller-manager --tail=30 | grep -i configmap
   2026-02-09T09:46:35Z    ERROR    controller-runtime.cache.UnhandledError    Failed 
   to watch{"reflector": 
   "pkg/mod/k8s.io/client-go@v0.34.0/tools/cache/reflector.go:290", "type": 
   "*v1.ConfigMap", "error": "failed to list *v1.ConfigMap: configmaps is forbidden: 
   User \"system:serviceaccount:nrr-system:nrr-controller-manager\" cannot list 
   resource \"configmaps\" in API group \"\" at the cluster scope"}
```

I tried like following, it worked -

```patch
diff --git a/cmd/main.go b/cmd/main.go
index 34652e9..1527845 100644
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -28,11 +28,14 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/rest"
 
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/filters"
@@ -107,6 +110,15 @@ func main() {
 		HealthProbeBindAddress: probeAddr,
 		LeaderElection:         enableLeaderElection,
 		LeaderElectionID:       "ba65f13e.readiness.node.x-k8s.io",
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&corev1.ConfigMap{}: {
+					Namespaces: map[string]cache.Config{
+						"nrr-system": {},
+					},
+				},
+			},
+		},
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
```

Priyankasaggu11929: @ajaysundark, for your review as well ^

config/admission-policy/policy.yaml
```diff
@@ -0,0 +1,71 @@
...
+  - name: existingTolerations
+    expression: |
+      has(object.spec.template.spec.tolerations) ? 
+        object.spec.template.spec.tolerations : []
+  
+  # Get taint keys from ConfigMap and parse to array
+  - name: taintKeys
+    expression: |
+      ("taint-keys" in params.data) && params.data["taint-keys"] != "" ?
+        params.data["taint-keys"].split(",") : []
```
Priyankasaggu11929:
```suggestion
        params.data["taint-keys"].split(",").filter(key, key != "") : []
```

currently, if someone intentionally add a empty taint key to the configmap 

our MAP policy will create a matching empty-key wildcard toleration

For example

```bash
❯ kubectl patch configmap readiness-taints -n nrr-system --type=merge   -p '{"data":{"taint-keys":"readiness.k8s.io/NetworkReady,,readiness.k8s.io/StorageReady"}}'
configmap/readiness-taints patched

// notice the two commas which basically mean an empty taint key
❯ kubectl get configmap -n nrr-system readiness-taints -o jsonpath='{.data.taint-keys}'
readiness.k8s.io/NetworkReady,,readiness.k8s.io/StorageReady


// the above empty taint, injected an empty-key wildcard toleration on the ds
❯ kubectl get ds test-ds -o jsonpath='{.spec.template.spec.tolerations[*]}' | jq 
{
  "effect": "NoSchedule",
  "key": "readiness.k8s.io/NetworkReady",
  "operator": "Exists"
}
{
  "effect": "NoSchedule",
  "operator": "Exists"
}
{
  "effect": "NoSchedule",
  "key": "readiness.k8s.io/StorageReady",
  "operator": "Exists"
}
```

config/admission-policy/binding.yaml
```diff
@@ -0,0 +1,15 @@
...
+kind: MutatingAdmissionPolicyBinding
+metadata:
+  name: inject-daemonset-readiness-tolerations-binding
+spec:
+  policyName: inject-daemonset-readiness-tolerations
+  # Reference the ConfigMap containing toleration data
+  paramRef:
+    name: readiness-taints
+    namespace: nrr-system
+    parameterNotFoundAction: Deny
```
ajaysundark: Does `parameterNotFoundAction:Deny` blocks daemonset creations in the absence of this config-map param?

config/rbac/role.yaml
ajaysundark: not from your changes, but dont think manager needs create or delete for node objects.

docs/admission-policy.md
```diff
@@ -0,0 +1,54 @@
...
+### Option 2: Direct kubectl apply
+
+```bash
+# Install CRDs first
+make install
+
+# Deploy policy and binding
+kubectl apply -f config/admission-policy/policy.yaml
+kubectl apply -f config/admission-policy/binding.yaml
+```
```
ajaysundark, line 74: hmm. this looks redundant. also, dont we need to show updating the config-map and how we maintain it with readiness-taints introduced later?

docs/admission-policy.md
```diff
@@ -0,0 +1,54 @@
...
+## Architecture
+
+```
+User applies DaemonSet
+    ↓
+API Server evaluates CEL policy
+    ↓
+Fetches Tolerations ConfigMap which contains the tolerations to be injected
+    ↓
+Injects tolerations (if applicable)
```
ajaysundark, line 28: This doesnt tell what is applicable. could we also add details on the annotation requirement?

config/admission-policy/binding.yaml
Priyankasaggu11929: yes, in my local testing with the PR changes, it is blocking creation of new daemonsets.

It works if a user creates node-readiness-rules first and then daemonsets, but the other way around blocks the creation of new daemonsets.

I tested the behaviour by disabling the empty configmap file in the kustomization.yaml, because I was thinking why do we need to have an empty configmap file, if the configmap creation is taken care of by the controller itself - https://github.com/kubernetes-sigs/node-readiness-controller/pull/117/changes#diff-b857080def3b2fecd3967ca35b1bdc29564dda0b064fa902fa24efcd55471899R721-R727

But the thing is - the controller logic only kicks in when a user create a new node-readiness-rule.
 So if a user created a daemonset first before creating the node-readiness-rules, that daemonset object creation is blocked.

The presence of the empty configmap file at the same time of creating MAP Policy/Policy-Binding fixes it.

```bash
❯ kubectl get node nrr-test-worker -o jsonpath={.spec.taints} | jq .
[
  {
    "effect": "NoSchedule",
    "key": "readiness.k8s.io/NetworkReady",
    "value": "pending"
  }
]

// applying MAP policy/policy-binding without empty configmap
❯ kubectl apply -k config/admission-policy
mutatingadmissionpolicy.admissionregistration.k8s.io/inject-daemonset-readiness-tolerations created
mutatingadmissionpolicybinding.admissionregistration.k8s.io/inject-daemonset-readiness-tolerations-binding created


❯ kubectl get cm -n nrr-system
NAME               DATA   AGE
kube-root-ca.crt   1      41s

// daemonset creation is blocked
❯ kubectl apply -f test-daemonset.yaml 
Error from server (Forbidden): error when creating "test-daemonset.yaml": policy 'inject-daemonset-readiness-tolerations' with binding 'inject-daemonset-readiness-tolerations-binding' denied request: failed to configure binding: no params found for policy binding with `Deny` parameterNotFoundAction

// creating a new nrr rule, which will trigger in turn configmap creation/patch per PR changes
❯ kubectl apply -f examples/cni-readiness/network-readiness-rule.yaml 
nodereadinessrule.readiness.node.x-k8s.io/network-readiness-rule created

❯ kubectl get cm readiness-taints -n nrr-system -o jsonpath={.data}
{"taint-keys":"readiness.k8s.io/NetworkReady"}


// now daemonset creation works
❯ kubectl apply -f test-daemonset.yaml 
daemonset.apps/test-ds created
```

ajaysundark: It feels like using Deny is high-risk to me.  If the controller is healthy and the config map exists, the tolerations are injected. Say if something is wrong with the controller, the cluster should remain functional, and users can still manually add tolerations to their DaemonSets if needed.

Auto-tolerations given an optional feature in node-readiness-controller shouldn't be fail-closed. It can be "best-effort".

config/admission-policy/policy.yaml
```diff
@@ -0,0 +1,71 @@
+---
+# MutatingAdmissionPolicy for automatic DaemonSet toleration injection
+# Reads taint keys from a ConfigMap parameter resource
+# Requires: MutatingAdmissionPolicy feature enabled in the cluster
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: MutatingAdmissionPolicy
+metadata:
+  name: inject-daemonset-readiness-tolerations
+spec:
+  failurePolicy: Fail
```
ajaysundark:
```suggestion
  failurePolicy: Ignore
```

same as discussed above, we shouldn't risk controller becoming a single point of failure for the entire cluster's DaemonSet management.

ajaysundark (changes requested):

config/rbac/role.yaml
pehlicd: removed them.

TomWis97:
What's the status of this PR? It would be really nice to have this functionality be part of the NRC! 🙂
I was watching the talk about NRC from KubeCon 2026 and this PR was mentioned.

Priyankasaggu11929:
Hello @TomWis97, thanks for the ping.

Apologies @pehlicd, I wasn't able to get to it after your last rebase.

Do you still have the bandwidth to rebase and refresh the changes in the PR once more? I'll work on getting it merged on priority. 

Thanks so much again!

pehlicd:
> Hello @TomWis97, thanks for the ping.
> 
> Apologies @pehlicd, I wasn't able to get to it after your last rebase.
> 
> Do you still have the bandwidth to rebase and refresh the changes in the PR once more? I'll work on getting it merged on priority.
> 
> Thanks so much again!

Hey @Priyankasaggu11929, no need to apologies, it is actually my fault I couldn't find enough capacity to continue working on it till now. I will rebase the PR and address the review comments. Once again sorry for inconvenience.

Priyankasaggu11929:
> I will rebase the PR and address the review comments. Once again sorry for inconvenience.

Thank you so much. (Truly appreciate your help and time on this. :pray: )

config/rbac/role.yaml
```diff
@@ -4,6 +4,16 @@ kind: ClusterRole
...
+- apiGroups:
+  - ""
+  resources:
+  - configmaps
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - watch
```
Priyankasaggu11929, line 16: These needs to be in a Role (not a ClusterRole) scoped to the `nrr-system` namespace.

The current `ClusterRole` gives the controller access to all configmaps across all namespaces (including kube-system), which is not needed at all.

config/admission-policy/policy.yaml
```diff
@@ -0,0 +1,73 @@
+---
+# MutatingAdmissionPolicy for automatic DaemonSet toleration injection
+# Reads taint keys from a ConfigMap parameter resource
+# Requires: MutatingAdmissionPolicy feature enabled in the cluster
+apiVersion: admissionregistration.k8s.io/v1beta1
```
Priyankasaggu11929, line 5:
```suggestion
apiVersion: admissionregistration.k8s.io/v1
```

MAP has graduated to stable `v1` starting 1.36.

Although let's document as a note in the accompanying markdown that if someone is on a cluster version < 1.36, then they need to use v1beta1 API

config/admission-policy/binding.yaml
```diff
@@ -0,0 +1,15 @@
+---
+# MutatingAdmissionPolicyBinding binds the policy to the ConfigMap parameter
+apiVersion: admissionregistration.k8s.io/v1beta1
```
Priyankasaggu11929, line 3:
```suggestion
apiVersion: admissionregistration.k8s.io/v1
```

same as https://github.com/kubernetes-sigs/node-readiness-controller/pull/117/changes#r3767735422

internal/controller/nodereadinessrule_controller.go
```diff
@@ -696,6 +710,132 @@ func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule
...
+	cm.Data = map[string]string{
+		"taint-keys": taintKeysStr,
+	}
+
+	if apierrors.IsNotFound(err) {
+		// Create new ConfigMap
+		log.Info("Creating readiness-taints ConfigMap", "taintCount", len(taintKeys))
+		if err := r.Create(ctx, cm); err != nil {
+			return fmt.Errorf("failed to create configmap: %w", err)
+		}
```
Priyankasaggu11929, line 822: we should move the configmap creation to the controller manager startup point itself

because I realised the configmap (post this PR changes merges) will be created on every NRR create/update/delete, regardless of whether the MAP policy/bindings are created or not. 

So it's better IMO to create it at the controller manager startup time only, once.
And that way, we also don't need to maintain the seed yaml manifest.

```go
if err := mgr.Add(manager.RunnableFunc(func(ctx context.Context) error {
    cm := &corev1.ConfigMap{
        ObjectMeta: metav1.ObjectMeta{
            Name:      "readiness-taints",
            Namespace: "nrr-system",
        },
    }
    err := mgr.GetClient().Create(ctx, cm)
    if err != nil && !apierrors.IsAlreadyExists(err) {
        return fmt.Errorf("failed to ensure readiness-taints configmap: %w", err)
    }
    return nil
})); err != nil {
    setupLog.Error(err, "unable to add configmap-init runnable")
    os.Exit(1)
}
```

internal/controller/nodereadinessrule_controller.go
```diff
@@ -696,6 +710,132 @@ func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule
...
+	} else {
+		// Update existing ConfigMap
+		log.V(1).Info("Updating readiness-taints ConfigMap", "taintCount", len(taintKeys))
+		patch := client.MergeFrom(existingCM.DeepCopy())
+		existingCM.Data = cm.Data
+		if err := r.Patch(ctx, existingCM, patch); err != nil {
+			return fmt.Errorf("failed to update configmap: %w", err)
+		}
```
Priyankasaggu11929, line 830: we should add a check before the patch

so, that we only do a Patch when `data.taint-keys` actually differs.
This will help reduce the API writes in cases like any reconcile that doesn't change the taint set or delete of the last NRR rule


```
	if existingCM.Data["taint-keys"] == taintKeysStr {
		log.V(2).Info("readiness-taints ConfigMap already up to date, skipping patch")
		return nil
	}

	patch := client.MergeFrom(existingCM.DeepCopy())
	existingCM.Data = cm.Data
	if err := r.Patch(ctx, existingCM, patch); err != nil {
		return fmt.Errorf("failed to update configmap: %w", err)
	}
```

internal/controller/nodereadinessrule_controller.go
```diff
@@ -696,6 +710,132 @@ func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule
...
+			}
+		}
+	}
+
+	if len(errors) > 0 {
+		return fmt.Errorf("failed to cleanup taints on some nodes: %s", strings.Join(errors, "; "))
+	}
+
+	return nil
+}
```
Priyankasaggu11929, line 764: this block can be removed. 

remnants from earlier. It was later removed as part of PR https://github.com/kubernetes-sigs/node-readiness-controller/pull/250

charts/nrr-controller/templates/rbac.yaml
```diff
@@ -46,6 +46,9 @@ rules:
   - apiGroups: [""]
     resources: ["nodes/status"]
     verbs: ["get"]
+  - apiGroups: [""]
+    resources: ["configmaps"]
+    verbs: ["create", "get", "list", "patch", "watch"]
```
Priyankasaggu11929, line 51: same as https://github.com/kubernetes-sigs/node-readiness-controller/pull/117/changes#r3767731277

docs/admission-policy.md
Priyankasaggu11929: @pehlicd, could we add a standalone para about the annotation  `readiness.k8s.io/auto-tolerate`  for opt out behavior? 

especially document the part that a typo like  "False", "no", etc  will silently be treated as opted-in with no error or warning anywhere

docs/admission-policy.md
Priyankasaggu11929: @pehlicd, option 2 can be removed. 

option 1 is same as option 2 essentially.

Priyankasaggu11929 (commented):
@pehlicd, thank you so much again!

I have left some in-line comments.

After that, i think we should be good to merge it, and follow up for anything else in new PRs.

Priyankasaggu11929:
@pehlicd - also please rebase your PR commits on top of the commits in the main branch. That will help remove the `needs-rebase` label hopefully.

config/admission-policy/policy.yaml
```diff
@@ -0,0 +1,73 @@
...
+kind: MutatingAdmissionPolicy
+metadata:
+  name: inject-daemonset-readiness-tolerations
+spec:
+  failurePolicy: Ignore
```
ajaysundark, line 10: can we add a comment above this:

> if there is a typo in the CEL logic or a runtime error during execution that causes the script to crash, this wont block the DaemonSets.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -696,6 +710,132 @@ func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule
 	return nil
 }
 
+// cleanupNodesAfterSelectorChange cleans up nodes that matched old selector but not new one.
+func (r *RuleReadinessController) cleanupNodesAfterSelectorChange(ctx context.Context, oldRule, newRule *readinessv1alpha1.NodeReadinessRule) error {
```
ajaysundark, line 714: This function doesnt have any reference. is this a missed cleanup?

cmd/main.go
```diff
@@ -142,6 +145,15 @@ func main() {
 		LeaderElection:          enableLeaderElection,
 		LeaderElectionID:        "ba65f13e.readiness.node.x-k8s.io",
 		LeaderElectionNamespace: leaderElectionNamespace,
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&corev1.ConfigMap{}: {
+					Namespaces: map[string]cache.Config{
+						"nrr-system": {},
```
ajaysundark, line 152: This namespace is hardcoded in code at few places. Is it possible for us to inject it via flag? helm-chart install flow now allows installation into other namespaces and this could break if nrr-system doesnt exist.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -696,6 +710,132 @@ func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule
...
+			taintKeysSet[rule.Spec.Taint.Key] = struct{}{}
+		}
+	}
+
+	// Convert set to comma-separated string
+	taintKeys := make([]string, 0, len(taintKeysSet))
+	for key := range taintKeysSet {
+		taintKeys = append(taintKeys, key)
+	}
+	taintKeysStr := strings.Join(taintKeys, ",")
```
ajaysundark, line 795: should we sort it first to make this deterministic and testable?

internal/controller/nodereadinessrule_controller.go
```diff
@@ -174,13 +175,20 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		r.Controller.SyncNodeStateMetrics(ctx, rule)
 	}
 
+	// Sync taints to ConfigMap for MutatingAdmissionPolicy
+	if err := r.Controller.syncTaintsConfigMap(ctx); err != nil {
```
ajaysundark, line 179: I'd prefer this feature behind a feature-flag like `--enable-auto-toleration-sync` and default off for a safety net. This will help in clusters where MAP is not yet available.

config/rbac/role.yaml
ajaysundark: I agree to this.  I'd also suggest adding a resource-name scope, as the controller touches exactly one configmap in one namespace.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -696,6 +710,132 @@ func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule
...
+
+	if len(errors) > 0 {
+		return fmt.Errorf("failed to cleanup taints on some nodes: %s", strings.Join(errors, "; "))
+	}
+
+	return nil
+}
+
+// syncTaintsConfigMap synchronizes readiness taints to a ConfigMap for admission policy.
+func (r *RuleReadinessController) syncTaintsConfigMap(ctx context.Context) error {
```
ajaysundark, line 767: can we please add some unit-tests coverage?

docs/admission-policy.md
```diff
@@ -0,0 +1,74 @@
...
+    ↓
+DaemonSet created with tolerations
+```
+
+## Failure behavior
```
ajaysundark, line 33: Thanks for documenting this behavior!

docs/admission-policy.md
```diff
@@ -0,0 +1,74 @@
...
+exists is admitted without readiness tolerations. Because the policy also matches
+`UPDATE`, the tolerations are injected on the next update to the DaemonSet;
+otherwise add them manually.
+
+Opt out per DaemonSet with the annotation
+`readiness.k8s.io/auto-tolerate: "false"`.
+
+## Deployment
+
+### Option 1: Using kustomize
```
ajaysundark, line 55: could we add helm option here as well?

docs/admission-policy.md
```diff
@@ -0,0 +1,74 @@
...
+This document describes how to deploy and use the MutatingAdmissionPolicy-based approach for automatically injecting readiness tolerations into DaemonSets.
+
+## Overview
+
+The MutatingAdmissionPolicy approach uses Kubernetes's native admission control mechanism with CEL (Common Expression Language) to inject tolerations **without running a webhook server**. This provides a simpler, more declarative alternative to the webhook-based approach.
+
+## Requirements
+
+> [!IMPORTANT]
+> MutatingAdmissionPolicy is needed to be enabled in the cluster.
```
ajaysundark, line 12: Can we also mention the version constraints (for MAP) here.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -696,6 +710,132 @@ func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule
...
+	// Extract unique taint keys with readiness.k8s.io/ prefix and NoSchedule effect
+	taintKeysSet := make(map[string]struct{})
+	for _, rule := range ruleList.Items {
+		// Skip rules that are being deleted
+		if !rule.DeletionTimestamp.IsZero() {
+			continue
+		}
+		if rule.Spec.Taint.Key != "" &&
+			strings.HasPrefix(rule.Spec.Taint.Key, "readiness.k8s.io/") &&
+			rule.Spec.Taint.Effect == corev1.TaintEffectNoSchedule {
```
ajaysundark, line 785: We are filtering only NoSchedule taints here. I think we should either sync all rule taint-effect keys or add  document the limitation that this will handle only NoSchedule keys, as NRC doesnt restrict such configuration.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -696,6 +710,132 @@ func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule
...
+	// Convert set to comma-separated string
+	taintKeys := make([]string, 0, len(taintKeysSet))
+	for key := range taintKeysSet {
+		taintKeys = append(taintKeys, key)
+	}
+	taintKeysStr := strings.Join(taintKeys, ",")
+
+	// Update or create ConfigMap
+	cm := &corev1.ConfigMap{
+		ObjectMeta: metav1.ObjectMeta{
```
ajaysundark, line 799: Could we align on the name as 'node-readiness' to think this beyond just the taints?

I think this config-map could be evolving into a cluster-wide "node-readiness info" object in the future, something like the cluster-info today, with information such as controller version etc. I could imagine deployment scenarios where name-spaced component owners or end-users may lack access to nrr-system namespace, but would still benefit from node-readiness milestones configured in the cluster. 

For the same reason this namespace should be configurable, so if operators want to have a readonly 'discovery contract' visible in kube-public, they can achieve it.

ajaysundark: This is another resource where managed labels (xref- https://github.com/kubernetes-sigs/node-readiness-controller/issues/438) would be valuable. cc @yindia

Karthik-K-N:
Hi @pehlicd do you have the bandwidth to take this to closure or guide if any other contributor is interested to drive this. Thanks

PR #122, opened 2026-02-07
feat(config): add webhook as kustomize component

AvineshTripathi:
## Description
This PR converts the webhook config to a component like metrics and cert-manager and enables it. It also removes dependency of service monitor from the controller.

NOTE: webhooks require TLS, so cert manager crds installation is mandatory and `ENABLE_TLS` needs to be `true`.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ CONTROLLER_GEN_PKG := sigs.k8s.io/controller-tools/cmd/controller-gen
 IMG_PREFIX ?= controller
 IMG_TAG ?= latest
 
-# ENABLE_METRICS: If set to true, includes Prometheus Service and ServiceMonitor resources.
+# ENABLE_METRICS: If set to true, includes Prometheus Service resources.
 ENABLE_METRICS ?= false
 ENABLE_TLS ?= false
 # ENABLE_WEBHOOK: If set to true, includes validating webhook. Requires ENABLE_TLS=true.


config/webhook/webhook_patch.yaml
```diff
@@ -3,11 +3,11 @@ kind: ValidatingWebhookConfiguration
 metadata:
   name: validating-webhook-configuration
   annotations:
-    cert-manager.io/inject-ca-from: $(CERTIFICATE_NAMESPACE)/$(CERTIFICATE_NAME)
+    cert-manager.io/inject-ca-from: nrr-system/nrr-serving-cert
```
AvineshTripathi, line 6: There is a reason these values are hardcoded here: they were previously populated using vars. However, vars is now deprecated (https://github.com/kubernetes-sigs/kustomize/discussions/5046
), and we should move to using `replacements`. When running Kustomize, it also throws the following warning indicating this deprecation.

```
# Warning: 'vars' is deprecated. Please use 'replacements' instead. [EXPERIMENTAL] Run 'kustomize edit fix' to update your Kustomization automatically.
```

If we switch to replacements, we run into a dependency issue. Both the webhook and metrics services need these replacements, but they are individual components and may or may not be deployed together. Because of this, we cannot keep the replacements in `config/default`. Placing them in individual components also does not work, as it fails to populate the `nrr-` prefix in the DNS names and annotations.

So I thought a better solution would be to hardcode it. Open for suggestions


Other places:
- https://github.com/kubernetes-sigs/node-readiness-controller/pull/122/changes#diff-5d354d4779d9351c4d8f5226dfaa9e85430effa32a48718a10eaf522a1056c89R8
- https://github.com/kubernetes-sigs/node-readiness-controller/pull/122/changes#diff-cf7731a38a1c580bd6f4538056f1a4ee90dd1f14515e61c6cdfc9e51a9a055d4R9

config/certmanager/certificate.yaml
```diff
@@ -8,17 +8,3 @@ metadata:
...
-  namespace: system
-spec:
-  commonName: nrr-metrics
-  dnsNames:
-  - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc
-  - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc.cluster.local
-  issuerRef:
-    kind: Issuer
-    name: selfsigned-issuer
-  secretName: metrics-server-cert
```
Priyankasaggu11929, line 24: (confirming as I read the changes)

Is this change is essentially a cleanup? Basically moving the certificate creation under the component that needs the certificate right?  

- so for prometheus (metrics-cert) is defined here now - `config/prometheus/tls/certificate.yaml`
- and same, webhook (serving-cert) is defined here - `config/webhook/certificate.yaml`

config/prometheus/tls/certificate.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: cert-manager.io/v1
+kind: Certificate
+metadata:
+  name: metrics-certs
+  namespace: system
+spec:
+  commonName: metrics
```
Priyankasaggu11929: just wondering, is there any use of `commonName` now? 
(because IIUC, hostname verification will happen with SAN entries below)

If it is for just completion sake in the certificate, let's add some `commonName` to the webhook certificate as well? 
Maybe it's a nit, but then atleast both our certificates will have common meta

config/prometheus/tls/kustomization.yaml
```diff
@@ -1,15 +1,14 @@
...
+resources:
+- certificate.yaml
+
 patches:
 # Enable HTTPS args in Deployment
 - path: manager_prometheus_metrics_tls.yaml
   target:
     kind: Deployment
-# Configure ServiceMonitor for TLS
-- path: monitor_tls_patch.yaml
```
Priyankasaggu11929, line 10: we delete the `monitor_tls_patch.yaml` file if not in use?

config/prometheus/tls/kustomization.yaml
```diff
@@ -18,63 +17,3 @@ patches:
 - path: metrics_service_tls_patch.yaml
   target:
     kind: Service
-
-replacements:
```
Priyankasaggu11929, line 22: was there any benefit of not hardcoding these values earlier?

And just to confirm - this `replacements` section is not required now because you've hardcoded the `dnsNames` in the new `config/prometheus/tls/certificate.yaml`?

```
  dnsNames:
  - nrr-controller-manager-metrics-service.nrr-system.svc
  - nrr-controller-manager-metrics-service.nrr-system.svc.cluster.local
```

config/prometheus/tls/certificate.yaml
```diff
@@ -0,0 +1,14 @@
...
+  dnsNames:
+  - nrr-controller-manager-metrics-service.nrr-system.svc
+  - nrr-controller-manager-metrics-service.nrr-system.svc.cluster.local
```
Priyankasaggu11929: Nit:

 I see in the webhook certificate below, the dnsNames entries are shorter:

```
  dnsNames:
  - nrr-webhook-service.nrr-system.svc
  - nrr-webhook-service.nrr-system.svc.cluster.local
```

should we trim the service name for metrics as well? to follow a pattern?
(remove the `controller-manager` bit?)

config/prometheus/kustomization.yaml
```diff
@@ -1,22 +1,5 @@
 apiVersion: kustomize.config.k8s.io/v1alpha1
 kind: Component
 resources:
-- monitor.yaml
```
Priyankasaggu11929, line 4: Delete the `monitor.yaml` file as well if not in use?

config/webhook/webhook_patch.yaml
```diff
@@ -3,11 +3,11 @@ kind: ValidatingWebhookConfiguration
...
 webhooks:
 - name: vnodereadinessrule.kb.io
   clientConfig:
     service:
       name: webhook-service
       namespace: system
-      path: /validate-nodereadiness-io-v1alpha1-nodereadinessrule
+      path: /validate-readiness-node-x-k8s-io-v1alpha1-nodereadinessrule
```
Priyankasaggu11929: why are we correcting the path here with the patch and not directly correcting the marker itself (which IMO is the source of truth) - https://github.com/kubernetes-sigs/node-readiness-controller/blob/354b1510874b623bc31a1f9ad75ebf04ced7fd0b/internal/webhook/nodereadinessgaterule_webhook.go#L46

I tried correcting the marker and removing this patch - it works.

So, I suggest we drop patching the path here and only keep this patch to add the required cert-manager annotation?

More so, because if we don't update the marker, any subsequent `make manifests` and `make deploy` will go back to using the old path defined in the markers, and that will break

config/default/manager_webhook_patch.yaml
```diff
@@ -1,27 +0,0 @@
...
-          protocol: TCP
-        volumeMounts:
-        - mountPath: /tmp/k8s-webhook-server/serving-certs
-          name: cert
-          readOnly: true
-      volumes:
-      - name: cert
-        secret:
-          defaultMode: 420
-          secretName: webhook-server-certs
```
Priyankasaggu11929, line 27: Again, only for confirmation -

This change is just moving the patch to `config/webhook/manager_webhook_patch.yaml` and using JSON Patch format?

config/webhook/webhook_patch.yaml
AvineshTripathi: Thanks, @Priyankasaggu11929. I totally agree on updating the marker itself

config/certmanager/certificate.yaml
AvineshTripathi: that is correct!

config/prometheus/tls/kustomization.yaml
AvineshTripathi: Earlier too, the `dnsName` field was not propagating correctly; however, it seems Prometheus was still allowing that, but in the case of a webhook, the API server does not allow a mismatch in dnsName.

Since I was doing the changes in webhook certificate, I corrected it as well.

config/default/manager_webhook_patch.yaml
AvineshTripathi: yes, that is correct

Makefile
```diff
@@ -278,21 +288,52 @@ undeploy: build-manifests-temp ## Undeploy controller from the K8s cluster. Use
...
+
+# Deploy with all features: metrics, TLS, webhook.
+.PHONY: deploy-full
+deploy-full: ENABLE_METRICS=true
+deploy-full: ENABLE_TLS=true
+deploy-full: ENABLE_WEBHOOK=true
+deploy-full: deploy ## Deploy with all features: metrics, TLS, webhook.
+
+.PHONY: undeploy-full
+undeploy-full: ENABLE_METRICS=true
```
ajaysundark, line 333: looks like undeploy target already ignores if resources are not found? would it then suffice to just keep one undeploy?

Priyankasaggu11929:
LGTM from my side. Thanks!

config/prometheus/tls/certificate.yaml
AvineshTripathi: i have removed the commonName field as the [doc](https://cert-manager.io/docs/usage/certificate/) suggest to avoid using it

config/webhook/webhook_patch.yaml
ajaysundark: Non blocking comment-

I need to make myself familiarize with the `replacements` field. but from the [description](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/replacements/) seems it could work? -

> Substitute field(s) in N target(s) with a field from a source.

we are looking to invest in helm for long-term so it's not urgent to handle this right away. maybe create a follow up backlog?

ajaysundark (commented):
/lgtm

Makefile
```diff
@@ -64,8 +64,9 @@ IMG_TAG ?= latest
 
 # ENABLE_METRICS: If set to true, includes Prometheus Service and ServiceMonitor resources.
```
ajaysundark:
```suggestion
# ENABLE_METRICS: If set to true, includes Prometheus Service resources.
```

nit: I think this is removed now?

PR #130, opened 2026-02-14
cleanup: Remove unused globalDryRun feature (#123)

### LINKED ISSUE #123: cleanup: Remove unused globalDryRun feature
The codebase has a hanging `globalDryRun` field in the `RuleReadinessController`. This global off-switch feature was never realized or used.
This task is to remove this dead code from the controllers.

https://github.com/kubernetes-sigs/node-readiness-controller/blob/354b1510874b623bc31a1f9ad75ebf04ced7fd0b/internal/controller/node_controller.go#L136

GGh41th:
## Description
Removing dead code for the GlobalDryRun feature , since it was never used.
## Related Issue
Fixes #123 

## Type of Change
/kind cleanup

## Checklist
- [ ✓ ] `make test` passes
- [ ✓ ] `make lint` passes

```release-note

```
Doc #(issue)

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -56,9 +56,6 @@ type RuleReadinessController struct {
 	// Cache for efficient rule lookup
 	ruleCacheMutex sync.RWMutex
 	ruleCache      map[string]*readinessv1alpha1.NodeReadinessRule // ruleName -> rule
-
-	// Global dry run mode (emergency off-switch)
-	globalDryRun bool
 }
 
 // RuleReconciler handles NodeReadinessRule reconciliation.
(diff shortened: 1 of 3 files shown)


internal/controller/nodereadinessrule_controller.go
```diff
@@ -56,9 +56,6 @@ type RuleReadinessController struct {
 	// Cache for efficient rule lookup
 	ruleCacheMutex sync.RWMutex
 	ruleCache      map[string]*readinessv1alpha1.NodeReadinessRule // ruleName -> rule
-
-	// Global dry run mode (emergency off-switch)
-	globalDryRun bool
```
AvineshTripathi, line 61: while i agree here that it is not implemented in all required places, I think we still should keep this field or at least make a comment to use it whenever we are implementing it across the project. Thoughts? @ajaysundark

ajaysundark: @AvineshTripathi I added this originally for the CUJ: cluster-admin realized some misconfiguration widely affecting cluster and want to turn-off NRC to avoid further damage. 

But then came to realization that an 'off-switch' has more risks for misconfiguration (affecting rules globally for nodes that would need it) and has cluster-wide impact. 

It could be better to handle the specific rules than a global 'just watch' change. If we hear feedback/ experience that this is needed, we could add this as a feature later on. wdyt?

AvineshTripathi: that make sense! Thanks for explaining

PR #131, opened 2026-02-19
:warning: Remove per node-data from rule status

### LINKED ISSUE #111: Remove per node-data from rule.status
Following up on https://github.com/kubernetes-sigs/node-readiness-controller/issues/89, we likely are shooting ourselves in the foot, with concurrent node-updates in the node.

We are planning for scalability tests #(TODO: capture issue here) for upto 1k, 2k and 5k nodes with [kwok](https://kwok.sigs.k8s.io/), to assess our goals for scale. This issue is to track removing node-references from rule.status object following sig-arch [api-conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)

Goal: 

1. Remove per node-data from status -- delete AppliedNodes, FailedNodes, and NodeEvaluations from NodeReadinessRuleStatus.
2. Instead of NodeEvaluations, emit K8s Events / metrics when significant actions occur for history / audit.

/kind api-change

Karthik-K-N:
## Description

This PR updates the NRR CRD status filed to remove the per-node related data.
## Related Issue

Fixes https://github.com/kubernetes-sigs/node-readiness-controller/issues/111

## Type of Change
/kind api-change

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
scalability: Update NodeReadinessRule Status object to remove the node related data.
```
Doc #(issue)


internal/controller/node_controller.go
```diff
@@ -139,85 +139,18 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
...
-			"rule", rule.Name,
-			"ruleResourceVersion", rule.ResourceVersion)
-
 		if err := r.evaluateRuleForNode(ctx, rule, node); err != nil {
+			// Log error and record event but continue with other rules.
 			log.Error(err, "Failed to evaluate rule for node",
 				"node", node.Name, "rule", rule.Name)
-			// Continue with other rules even if one fails
-			r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
+			r.recorder.Eventf(node, corev1.EventTypeWarning, "RuleEvaluationError", "Failed to evaluate rule for node: %v", err)
```
ajaysundark, line 146: cc @Priyankasaggu11929 This PR introduces event recorder and some node eventing which may be related to https://github.com/kubernetes-sigs/node-readiness-controller/issues/134

Priyankasaggu11929: Ack-ing ^ @ajaysundark

ajaysundark:
I ran the scale experiments from #185, and noted below.

We see below costs at scale:

1.  The rule / status object grew up to 1.01 MB (at 5K limit); extrapolating the data suggests we'd hit max limits at 7.5K nodes until the etcd per-object limit (1.5 MB). Though we currently set `+kubebuilder:validation:MaxItems=5000` for our controller, this shouldn't be a blocking factor and should be scalable limit.

2. any single node's change rewrites the whole list so write cost grows with fleet size. This is not ideal as a node condition change (eg: a DS rollout) on a large cluster creates sweeping update wave to overwhelm KCP.

3.  A really large status array is operationally unusable; requires parsing for simple / single-node drill down analysis.

However I also think 'removing' doesnt have to be 'dropping' the fields. We should plan to handle graceful migration (with internal wrapping at controller) from v1alpha1 to v1alpha2 shape so it'll continue work for existing users. 

cc @Karthik-K-N

Karthik-K-N:
> However I also think 'removing' doesnt have to be 'dropping' the fields. We should plan to handle graceful migration (with internal wrapping at controller) from v1alpha1 to v1alpha2 shape so it'll continue work for existing users.

Thank you for validating and providing the data.
Yes, with the help of conversion webhooks we can make sure we don't loose data across round trips with new API version.

ajaysundark:
Thanks @Karthik-K-N , I'm looking at the proposed shape https://github.com/kubernetes-sigs/node-readiness-controller/pull/133, and it feels like the changes are mostly additive and are in controller managed objects.

we should also consider evolving v1alpha1 vs introducing v1alpha2. Evaluate the pros and cons in the proposal. We could start this doc and get community feedback to decide next steps.

Karthik-K-N:
> Thanks @Karthik-K-N , I'm looking at the proposed shape #133, and it feels like the changes are mostly additive and are in controller managed objects.
> 
> we should also consider evolving v1alpha1 vs introducing v1alpha2. Evaluate the pros and cons in the proposal. We could start this doc and get community feedback to decide next steps.

Sure, Completely agree, I will do that and share asap. Thanks

ajaysundark:
Thanks for all your effort driving these refactors!

This is obsolete. We aligned on design in #345, we will slowly deprecate the rule.status node-only entries with a migration plan to the node-evaluation #345 design.

PR #140, opened 2026-02-27
Add link checker to fix broken links in markdown

### LINKED ISSUE #137: feat: Add automated link checker to 'make verify'
Our documentation contains many links to internal pages and external websites. To ensure a good user experience, we need to regularly check that these links are not broken. 

This task is to create a script (`hack/verify-links.sh`) that automatically checks for broken links in all Markdown (`.md`) files and integrate it into our existing verification pipeline. The main verification script (`hack/verify-all.sh`) could then be updated to execute `hack/verify-links.sh`. Running `make verify` should successfully run the link checker over all `.md` files in the repository.

We can use `docs/book/install-and-build-mdbook.sh` as a reference for how to download a tool based on OS and architecture. The `lychee` tool is a good option for this purpose.
 
example, ref: containerd project [has this workflow](https://github.com/containerd/containerd/blob/589bf746c5f7e364783defc88c7edd14a158091d/.github/workflows/links.yml#L17)

farazmd:
## Description
Added link checker (lychee) to check for broken links in markdown files as part of the verify process.
Also fixed broken links in `README.md` and `docs/book/src/introduction.md`

## Related Issue

Fixes #137  

## Type of Change

/kind feature

## Testing

Running make verify - runs the verify-links.sh script
## Checklist
- [ x ] `make test` passes
- [ x ] `make lint` passes

```release-note
NONE
```

### DIFF
--- a/hack/verify-links.sh
+++ b/hack/verify-links.sh
@@ -26,6 +26,10 @@ ROOT_DIR=$(dirname "${BASH_SOURCE[0]}")/..
 
 cd "$ROOT_DIR"
 
+# Create temp directory to download lychee binary
+TEMP_DIR=$(mktemp -d)
+trap 'rm -rf "${TEMP_DIR}"' EXIT
+
 # Detect OS and architecture
 os=$(go env GOOS)
 arch=$(go env GOARCH)
@@ -34,7 +38,7 @@ arch=$(go env GOARCH)
 LYCHEE_VERSION="${LYCHEE_VERSION:-0.23.0}"
 LYCHEE_BINARY="lychee"
 EXT="tar.gz"
-EXTRACT_CMD="tar -C /tmp -xzf"
+EXTRACT_CMD="tar -C ${TEMP_DIR} -xzf"
 case "${os}-${arch}" in
 linux-amd64)
     LYCHEE_BASENAME="${LYCHEE_BINARY}-x86_64-unknown-linux-gnu"
@@ -45,10 +49,6 @@ linux-arm64)
 darwin-arm64)
     LYCHEE_BASENAME="${LYCHEE_BINARY}-arm64-macos"
     ;;
-windows-amd64)
-    echo "skipping verify-links for windows"
-    exit 0
-    ;;
 *)
     echo "Unsupported platform: ${os}-${arch}" >&2
     exit 1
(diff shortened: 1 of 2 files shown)


hack/verify-links.sh
```diff
@@ -0,0 +1,78 @@
...
+linux-amd64)
+    LYCHEE_BASENAME="${LYCHEE_BINARY}-x86_64-unknown-linux-gnu"
+    ;;
+linux-arm64)
+    LYCHEE_BASENAME="${LYCHEE_BINARY}-aarch64-unknown-linux-gnu"
+    ;;
+darwin-arm64)
+    LYCHEE_BASENAME="${LYCHEE_BINARY}-arm64-macos"
+    ;;
+windows-amd64)
```
Priyankasaggu11929: Is there a static lychee binary (built with musl maybe) that we can use for `windows-amd64` case (if we are adding it?) 
There's also ` lychee-x86_64-windows.exe ` in the artifacts. 

But overall, I feel like, rather than skipping check for `windows-amd64`, it should be flagged as unsupported platform in our controller case.

hack/verify-links.sh
```diff
@@ -0,0 +1,78 @@
...
+set -x
+curl -fL -o /tmp/${LYCHEE_BINARY}.${EXT} "${LYCHEE_URL}"
+
+${EXTRACT_CMD} "/tmp/${LYCHEE_BINARY}.${EXT}"
+chmod +x /tmp/${LYCHEE_BINARY}
+
+
+# Remove artifacts
+rm -rf /tmp/${LYCHEE_BINARY}.${EXT}
+rm -rf /tmp/docs
```
Priyankasaggu11929:
```
rm -rf /tmp/docs
``` 

where are these coming from ^?

hack/verify-links.sh
```diff
@@ -0,0 +1,78 @@
...
+
+# Change back to root directory
+cd "$ROOT_DIR"
```
Priyankasaggu11929: duplicate?

You're already in `ROOT_DIR` no?

https://github.com/kubernetes-sigs/node-readiness-controller/pull/140/changes#diff-e6c42564ed69f2e6c8632e691e010b21644bea4cda9c196b15ebe5fb4920f5ceR27

hack/verify-links.sh
```diff
@@ -0,0 +1,78 @@
...
+    ;;
+esac
+
+LYCHEE_URL="https://github.com/lycheeverse/lychee/releases/download/lychee-v${LYCHEE_VERSION}/${LYCHEE_BASENAME}.${EXT}"
+
+echo "downloading ${LYCHEE_BASENAME}.${EXT} from ${LYCHEE_URL}"
+set -x
+curl -fL -o /tmp/${LYCHEE_BINARY}.${EXT} "${LYCHEE_URL}"
+
+${EXTRACT_CMD} "/tmp/${LYCHEE_BINARY}.${EXT}"
```
Priyankasaggu11929: let's wrap the binary downloading and extracting with error handling

hack/verify-links.sh
```diff
@@ -0,0 +1,78 @@
...
+*)
+    echo "Unsupported platform: ${os}-${arch}" >&2
+    exit 1
+    ;;
+esac
+
+LYCHEE_URL="https://github.com/lycheeverse/lychee/releases/download/lychee-v${LYCHEE_VERSION}/${LYCHEE_BASENAME}.${EXT}"
+
+echo "downloading ${LYCHEE_BASENAME}.${EXT} from ${LYCHEE_URL}"
+set -x
```
Priyankasaggu11929: I think `set -x` is not needed in the middle of the script

Priyankasaggu11929:
/ok-to-test

@farazmd, thanks for the PR. I have left a few in-line comments, and dropping some suggestions below as well.

Let's create a temp dir in the beginning of the  `verify-link` script and do a cleanup with `trap` rather than us manually catching and cleaning up every single artifact

```
TEMP_DIR=$(mktemp -d)
trap 'rm -rf "${TEMP_DIR}"' EXIT
```

---

Also, let's add a `.lycheeignore` file to ignore links like  `example.com` or `localhost` etc

hack/verify-links.sh
farazmd: Those are from the lychee tarball. Just removing those artifacts since we don't need them

hack/verify-links.sh
farazmd: When I extract the lychee tarball, I am passing in the change directory option to move into /tmp before extract. Otherwise it would extract them in the current working directory. Hence the change back to root_dir

hack/verify-links.sh
Priyankasaggu11929: got it.

```
TEMP_DIR=$(mktemp -d)
trap 'rm -rf "${TEMP_DIR}"' EXIT
```

This will help.

hack/verify-links.sh
Priyankasaggu11929: let's use a `TEMP_DIR` to point to the required tmp path while extracting the tarball, rather than changing directory.

And then the trap will help with the cleanup at the end all together.

```
TEMP_DIR=$(mktemp -d)
trap 'rm -rf "${TEMP_DIR}"' EXIT
```

PR #142, opened 2026-02-28
feat/static-pod-initialization (#125)

### LINKED ISSUE #125: static-pod initialization for proper resource accounting
**Static Pod Readiness**

Static pods are managed directly by the Kubelet on a specific node, without the API server's involvement in their creation. To make them visible to the control plane, the Kubelet creates a Mirror Pod for each static pod in the API server. However, if there's a delay before Mirror Pods are created, the scheduler is unaware of the resources (cpu/memory) consumed by these static pods and could potentially over-commit on the node[1].

This race condition could be avoided with NRC to ensure a node only accepts workloads after all static pods are successfully synchronized with the API server.

This doc issue is to demonstrate how a [npd](https://github.com/kubernetes/node-problem-detector) custom-plugin script could be utilized to work with NRC using a 'StaticPodsSynced' node condition.

[1] Ref:

More details here: https://github.com/kubernetes/kubernetes/issues/115325, https://github.com/kubernetes/website/issues/47264 and https://github.com/kubernetes/kubernetes/pull/126870

GGh41th:
## Description

Integrate with npd using a custom problem daemon to ensure that static pods are mirrored before scheduling on the node.

## Related Issue

Fixes #125 

## Type of Change

/kind feature

## Checklist
- [*] `make test` passes
- [*] `make lint` passes

```release-note

```
Doc #(issue)

### DIFF
--- a/examples/staticpods-readiness/README.md
+++ b/examples/staticpods-readiness/README.md
@@ -1,0 +1,42 @@
+# Node Readiness Examples (Static Pods)
+
+This example demonstrates how NRC can be used alongside [NPD (Node Problem Detector)](https://github.com/kubernetes/node-problem-detector) to make sure that all static pods are mirrored in the API server to avoid over-committing issues. See [#115325](https://github.com/kubernetes/kubernetes/issues/115325), [#47264](https://github.com/kubernetes/website/issues/47264), and [#126870](https://github.com/kubernetes/kubernetes/pull/126870) for reference.
+
+## Deployment Steps
+
+1. Deploy a testing cluster. For this example we use Kind with a pre-tainted worker node and a mounted static pod manifest:
+
+   ```bash
+   kind create cluster --config examples/staticpods-readiness/kind-cluster-config.yaml
+   ```
+
+2. Install NRC:
+
+   ```bash
+   VERSION=v0.2.0
+   kubectl apply -f https://github.com/kubernetes-sigs/node-readiness-controller/releases/download/${VERSION}/crds.yaml
+   kubectl apply -f https://github.com/kubernetes-sigs/node-readiness-controller/releases/download/${VERSION}/install.yaml
+   ```
+
+3. Deploy NPD as a DaemonSet with the static pods readiness configuration:
+
+   ```bash
+   ./examples/staticpods-readiness/setup-staticpods-readiness.sh
+   ```
+
+   This deploys NPD with:
+   - Init container that downloads `kubectl`, `yq`, and `jq`
+   - Custom plugin that checks if static pods are mirrored
+   - RBAC permissions to read pods and nodes
+
+4. Apply the node readiness rule:
+   ```bash
+      kubectl apply -f examples/staticpods-readiness/staticpods-readiness-rule.yaml
+   ```
+
+## How It Works
+
+The NPD problem daemon script runs every 30s:
+- Finds manifests in `/etc/kubernetes/manifests`
+- Parses the manifests and checks if mirror pod `{pod-name}-{node-name}` exists
+- Sets `StaticPodsMissing=False` if all pods are mirrored
--- a/examples/staticpods-readiness/kind-cluster-config.yaml
+++ b/examples/staticpods-readiness/kind-cluster-config.yaml
@@ -1,0 +1,17 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+nodes:
+- role: control-plane
+- role: worker
+  extraMounts:
+  - hostPath: examples/staticpods-readiness/test-staticpod.yaml
+    containerPath: /etc/kubernetes/manifests/test-staticpod.yaml
+    readOnly: true
+  kubeadmConfigPatches:
+  - |
+    kind: JoinConfiguration
+    nodeRegistration:
+      taints:
+      - key: "readiness.k8s.io/StaticPodsMissing"
+        value: "pending"
+        effect: "NoSchedule"  
--- a/examples/staticpods-readiness/npd/npd-ds.yaml
+++ b/examples/staticpods-readiness/npd/npd-ds.yaml
@@ -1,0 +1,113 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: node-problem-detector-staticpods
+  namespace: kube-system
+  labels:
+    app: node-problem-detector
+    component: staticpods-monitor
+spec:
+  selector:
+    matchLabels:
+      app: node-problem-detector
+      component: staticpods-monitor
+  template:
+    metadata:
+      labels:
+        app: node-problem-detector
+        component: staticpods-monitor
+    spec:
+      serviceAccountName: node-problem-detector-staticpods
+      tolerations:
+      - operator: Exists
+        effect: NoSchedule
+      - operator: Exists
+        effect: NoExecute
+      initContainers:
+      # Download tools (kubectl, yq, jq) for the check script
+      - name: install-tools
+        image: busybox:1.36
+        command:
+        - sh
+        - -c
+        - |
+          set -e
+          ARCH=$(uname -m)
+          case $ARCH in
+            x86_64) ARCH="amd64" ;;
+            aarch64) ARCH="arm64" ;;
+          esac
+          
+          # Install kubectl
+          KUBECTL_VERSION="v1.35.0"
+          wget -O /tools-bin/kubectl "https://dl.k8s.io/release/${KUBECTL_VERSION}/bin/linux/${ARCH}/kubectl"
+          chmod +x /tools-bin/kubectl
+          
+          # Install yq
+          YQ_VERSION="v4.44.1"
+          wget -O /tools-bin/yq "https://github.com/mikefarah/yq/releases/download/${YQ_VERSION}/yq_linux_${ARCH}"
+          chmod +x /tools-bin/yq
+          
+          # Install jq
+          JQ_VERSION="1.7.1"
+          wget -O /tools-bin/jq "https://github.com/jqlang/jq/releases/download/jq-${JQ_VERSION}/jq-linux-${ARCH}"
+          chmod +x /tools-bin/jq
+        volumeMounts:
+        - name: tools-bin
+          mountPath: /tools-bin
+      containers:
+      - name: node-problem-detector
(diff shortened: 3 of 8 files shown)


ajaysundark:
Thanks for the PR, @GGh41th , I'll get to it this week.

examples/staticpods-readiness/check-staticpods-synced.sh
```diff
@@ -0,0 +1,184 @@
...
+# limitations under the License.
+
+
+# This script will check that all static pods are mirrored in the
+# API server.
+
+# Exit code constants
+readonly OK=0
+readonly NONOK=1
+readonly UNKOWN=2
```
AvineshTripathi:
```suggestion
readonly UNKNOWN=2
```

examples/staticpods-readiness/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+   ```bash
+   kubectl apply -f nrr.yaml
+   ```
+
+3. Deploy NPD, either as a DaemonSet using the official Helm chart, or in standalone mode. Keep in mind that if you are going to go the Helm way, you need to modify the NPD image to include the binaries that your script depends on (in our case, we need `curl` and `kubectl`), or download them in your script (this is not recommended since you'll have to set a high timeout in the custom plugin monitor config, at least for the initial script run).
+
+   ### Standalone Mode
+
+   In this example, and since I have a 1-node Kind cluster, I will go the standalone way. Note that this is the default shipping method in GKE and AKS.
```
AvineshTripathi:
```suggestion
   In this example, and since we have a 1-node Kind cluster, we will go the standalone way. Note that this is the default shipping method in GKE and AKS.
```

AvineshTripathi: > Note that this is the default shipping method in GKE and AKS.

Little confused here, do we know why this is a default method for GKE and AKS and why not Helm? Just out of curiosity

examples/staticpods-readiness/check-staticpods-synced.sh
```diff
@@ -0,0 +1,184 @@
...
+YQ_PATH="/opt/yq"
+
+JQ_VERSION="jq-1.8.1"
+JQ_PATH="/opt/jq"
+
+
+if [ ! -f "$YQ_PATH" ]; then 
+
+    echo "yq not found at $YQ_PATH, downloading..."
+    OS=$(uname -s | tr [:upper:] [:lower:])
```
AvineshTripathi: this part looks redundant; can we make a common function and pass arguments to get similar behavior

examples/staticpods-readiness/check-staticpods-synced.sh
```diff
@@ -0,0 +1,184 @@
...
+        echo "Static pod $FNAME is missing a mirror pod"
+        exit "$NONOK"
+    fi
+}
+
+
+parse_yaml_manifest() {
+
+    NAME=$($YQ_PATH -r '.metadata.name' $1 | head -1) 
+    NAMESPACE=$($YQ_PATH -r '.metadata.namespace // "default"' $1 | head -1)
```
AvineshTripathi:
```suggestion
    NAMESPACE=$($YQ_PATH -r '.metadata.namespace // "kube-system"' $1 | head -1)
```

examples/staticpods-readiness/check-staticpods-synced.sh
```diff
@@ -0,0 +1,184 @@
...
+    chmod +x "$JQ_PATH"
+
+fi
+
+parse_json_manifest() {
+
+    # The yaml file might be a multi-document yaml , but given that kubelet
+    # only reads the first document we will skip the rest.
+    NAME=$($JQ_PATH -r '.metadata.name' $1) 
+    NAMESPACE=$($JQ_PATH -r '.metadata.namespace // "default"' $1)
```
AvineshTripathi:
```suggestion
    NAMESPACE=$($JQ_PATH -r '.metadata.namespace // "kube-system"' $1) 
```

examples/staticpods-readiness/staticpods-syncer.json
```diff
@@ -0,0 +1,25 @@
...
+    "timeout": "1m",
+    "max_output_length": 80,
+    "concurrency": 1
+  },
+  "source": "staticpods-custom-plugin-monitor",
+  "conditions": [
+    {
+      "type": "StaticPodsMissing",
+      "reason": "mirrorpodsfound",
+      "message": "all mirros pods were created"
```
AvineshTripathi:
```suggestion
      "message": "all mirror pods were created"
```

examples/staticpods-readiness/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+          "timeout": "1m",
+          "max_output_length": 80,
+          "concurrency": 1
+        },
+        "source": "staticpods-custom-plugin-monitor",
+        "conditions": [
+          {
+            "type": "StaticPodsMissing",
+            "reason": "mirrorpodsfound",
+            "message": "all mirros pods were created"
```
AvineshTripathi:
```suggestion
            "message": "all mirror pods were created"
```

AvineshTripathi (changes requested):

examples/staticpods-readiness/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+   Use the `values.yaml` file to override the default Helm values, and don't forget to add the required binaries to the NPD image (`curl`, `kubectl`).
+
+   ```bash
+   helm repo add deliveryhero https://charts.deliveryhero.io/
+   helm repo update
+   helm install --generate-name deliveryhero/node-problem-detector -f values.yaml
+   ```
+
+   > **Note:** For a more robust and reliable version of the shell script, you can check this [standalone binary](https://github.com/GGh41th/NPD-staticpods-syncer).
```
Priyankasaggu11929: I suggest that we deploy NPD as a daemonset.

And store - (i) the NPD custom plugin (`staticpods-syncer.json`) content and (ii) the script with your actual logic for checking static/mirror pods, both inside a configmap maybe.

Right now, I see in the README documentation, you have suggested 2 ways of deploying NPD - IMO, we leave that to the NPD project documentation itself, and we can just include the above 2 "ready-to-use" yaml files in our example (will make it easier for the user)

examples/staticpods-readiness/check-staticpods-synced.sh
```diff
@@ -0,0 +1,184 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# This script will check that all static pods are mirrored in the
+# API server.
```
Priyankasaggu11929: Similar to what I suggest here - https://github.com/kubernetes-sigs/node-readiness-controller/pull/142/changes#r2912619130, much of the logic of this script can go in the configmap yaml

And whatever extra tools (like from your example - kubectl, yq, jq) we need on top of the NPD container image, let's install them in an init container inside the daemonset.

examples/staticpods-readiness/values.yaml
```diff
@@ -0,0 +1,47 @@
...
+    path: /opt/check-staticpods-synced.sh
+    type: File
+
+extraVolumeMounts:
+- name: static-pod-syncer
+  mountPath: /opt/check-staticpods-synced.sh
+
+settings:
+  custom_monitor_definitions:
+    staticpods-syncer.json: |
```
Priyankasaggu11929: This looks like repetitive to your changes here - https://github.com/GGh41th/node-readiness-controller/blob/e0463ce0bce4dab8499e48e4f8e65eb9c55f5de2/examples/staticpods-readiness/staticpods-syncer.json

same suggestion - https://github.com/kubernetes-sigs/node-readiness-controller/pull/142/changes#r2912619130

this can go in the configmap yaml

Priyankasaggu11929 (changes requested):
Thanks for the PR @GGh41th.

I have left a few in-line suggestions to make the example a bit more easy to test and consume (more like automating the many manual steps we have right now).

I was trying to test the PR locally, and was playing around with what I suggested in my comments.  I've pushed them here (if you want to reference)- https://github.com/Priyankasaggu11929/node-readiness-controller/commit/6f0e78811a698813b56608444f4e47a7013061ab

GGh41th:
Thanks for the comments everyone, I will try to make the changes as soon as I can.

examples/staticpods-readiness/README.md
GGh41th: It's not explicitly stated, but from my understanding they deploy it that way because you'll need to bake whatever dependencies your problem daemons might need into the npd image, which is inconvenient for a program you intend to ship as a default add on. So instead they deploy it as a systemd service and it's up to users to install whatever they need on the node.
If you feel that the comment is unecessary I can remove it.

examples/staticpods-readiness/check-staticpods-synced.sh
GGh41th: Here I'm just parsing the static pods manifests, if a namespace wasn't specified then the pod will be created in the default namespace, that's why m defaulting to that.

examples/staticpods-readiness/check-staticpods-synced.sh
GGh41th: Same as above

examples/staticpods-readiness/values.yaml
GGh41th: Thanks for the suggestions, we surely can automate a lot of this stuff, I'll use your commit as a reference.

examples/staticpods-readiness/README.md
ajaysundark: yes, many managed k8s-provider setup has npd as a default addon (documented here: https://github.com/kubernetes/node-problem-detector/blob/master/README.md).

ajaysundark:
@GGh41th Are you still looking into this? It seems like the PR is mostly in good shape, but there are still pending comments you're looking to address.

Karthik-K-N:
@GGh41th gentle ping, If you are occupied please let us know or you can guide the other contributor to take this forward. Thanks.

Issue #139, opened 2026-02-26
feat: Constrained impersonation for secure node status updates


ajaysundark:
The node-readiness-controller relies on external components (reporters) to update Node conditions. Because these reporters often run on the Nodes themselves, securing the communication between the reporter and the Kubernetes API is critical.

In our current [examples](https://github.com/kubernetes-sigs/node-readiness-controller/blob/6fa1e9351fafdca862a5629de41580eabf645320/examples/security-agent-readiness/falco-rbac-node-status-rbac.yaml#L10-L11), the reporter might require broad permissions to patch or update any Node in the cluster. This creates a security risk: if a single Node is compromised, the attacker could theoretically manipulate the status of any other Node.

By leveraging Constrained User Impersonation [KEP-5284](https://github.com/kubernetes/enhancements/issues/5284), we can significantly harden this posture by impersonating the identity of the specific node it is scheduled on and nothing else.

Example RBAC:

```
# 1. Role will allow impersonating the local node identity
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: node-readiness-impersonator
rules:
- apiGroups: ["authentication.k8s.io"]
  resources: ["nodes"]
  verbs: ["impersonate:scheduled-node"]     # can only impersonate the Node name provided by API verified spec.nodeName
---
# 2. Define only node/status actions are allowed during impersonation
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: node-status-patcher-constrained
rules:
- apiGroups: [""]
  resources: ["nodes/status"]
  verbs: ["impersonate-on:patch"]         #  limits the actions the reporter can perform

```

This will ensure that the readiness-reporter components are restricted to the `nodes/status` subresource to prevent modification of the spec (e.g., labels or taints), also while restricting it to the specific node.

**Tasks:**
[ ] Verify the behavior (as the K8s feature is currently Alpha, proposed Beta in 1.36) in Kind cluster 
[ ] Update the examples used in node-readiness-controller to use constrained impersonation
[ ] Document  at docs/book/src/operations/security.md

PR #143, opened 2026-03-02
feat: constrained impersonation for secure node status updates

ali-a-a:
## Description

This PR adds support for Constrained Impersonation so reporters can only update the status of the node they run on, instead of any node in the cluster.

### Changes

- Reporter reads `IMPERSONATE_NODE` env var to send `Impersonate-User: system:node:<nodeName>` headers when enabled. It is an opt-in feature as older Kubernetes versions don't support it.
- A new example (`examples/constrained-impersonation/`) with RBAC, DaemonSet, and readiness rule is added.
- New docs (`operations/security.md` and `examples/constrained-impersonation.md`) are added.

### Notes

- This feature requires Kubernetes v1.35+ (Alpha) or v1.36+ (Beta, enabled by default).
- Existing examples are unchanged, and impersonation is opt-in via a new example.

## Related Issue

Fixes https://github.com/kubernetes-sigs/node-readiness-controller/issues/139

## Type of Change

/kind feature

## Testing

You can follow along this [document](https://github.com/ali-a-a/node-readiness-controller/blob/f437da759e782b584f0073d0c6c4878ed5588365/docs/book/src/examples/constrained-impersonation.md) to test it locally.

## Checklist
- [X] `make test` passes
- [X] `make lint` passes

```release-note
Add support for Constrained Impersonation (KEP-5284) in the readiness condition reporter. Requires Kubernetes v1.35+.
```

### DIFF
--- a/docs/book/src/examples/constrained-impersonation.md
+++ b/docs/book/src/examples/constrained-impersonation.md
@@ -1,0 +1,110 @@
+# Secure Status Reporting
+
+By default, a readiness reporter's ServiceAccount is granted broad permissions to update the status of **any** Node in the cluster. If a node is compromised, an attacker can manipulate the readiness status of every other node.
+
+[Constrained Impersonation (KEP-5284)](https://github.com/kubernetes/enhancements/issues/5284) solves this by allowing the reporter to impersonate only the Node it runs on. The API server enforces this at the authorization layer, so that no other node's status can be touched.
+
+This guide walks through a CNI readiness example that uses constrained impersonation instead of broad RBAC. It is a hardened variant of the [CNI Readiness](./cni-readiness.md) example.
+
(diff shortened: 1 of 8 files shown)


docs/book/src/examples/constrained-impersonation.md
```diff
@@ -0,0 +1,110 @@
+# Constrained Impersonation
```
ajaysundark:
```suggestion
# Secure Status Reporting
```

ali-a-a:
> @ali-a-a Thanks, this looks great and detailed. I have a minor suggestion for updating the title for the docs, to show the value than the technical (KEP) detail. Consider renaming it to "Secure Status Reporting".

Thanks for the review, @ajaysundark. That makes sense. I updated the title.

PR #157, opened 2026-03-09
:sparkles: Add Podman support

Karthik-K-N:
## Description

Add support for Podman

## Type of Change
/kind feature

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Add Podman support
```
Doc #(issue)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,8 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore build artifacts and unnecessary files
+# Compatible with both Docker and Podman
+# Use explicit exclusions (not negation patterns) for better compatibility
+
+# Build artifacts and unnecessary files
 bin/
 dist/
 .git/
@@ -15,8 +18,19 @@ test/
 *.svg
 *.png
 
-# Keep test files out of the image
+# Test files
 **/*_test.go
 
-# Keep Go module cache out
+# Cache directories
 .cache/
+
+# IDE and editor files
+.vscode/
+.idea/
+*.swp
+*.swo
+*~
+
+# OS files
+.DS_Store
+Thumbs.db
--- a/Makefile
+++ b/Makefile
@@ -63,7 +63,7 @@ IMG_PREFIX ?= controller
 IMG_TAG ?= latest
 
 # Kind cluster name for loading images
-KIND_CLUSTER ?= nrr-local
+KIND_CLUSTER ?= nrr-test
 
 # ENABLE_METRICS: If set to true, includes Prometheus Service and ServiceMonitor resources.
 ENABLE_METRICS ?= false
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -177,7 +181,11 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
 docker-build: ## Build docker image with the manager.
```
Priyankasaggu11929:
```suggestion
container-build: ## Build docker image with the manager.
```
`docker-build` for building with podman feels a bit confusing

Makefile
```diff
@@ -197,9 +205,29 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
...
+	@echo "Cleaning up temporary tar archive..."
+	@rm /tmp/controller-image.tar
+	@echo "Image loaded successfully!"
+else
+	@echo "Loading Docker image into kind cluster: $(KIND_CLUSTER)"
+	@kind load docker-image $(IMG_PREFIX):$(IMG_TAG) --name $(KIND_CLUSTER)
+endif
+
 .PHONY: docker-build-reporter
 docker-build-reporter: ## Build docker image with the reporter.
```
Priyankasaggu11929:
```suggestion
.PHONY: container-build-reporter
container-build-reporter: ## Build container image with the reporter.
```

Makefile
Karthik-K-N: Yeah, Correct, I just update it. Thanks.

docs/TEST_README.md
```diff
@@ -38,21 +38,44 @@ make install
...
+make container-build CONTAINER_TOOL=podman IMG_PREFIX=controller IMG_TAG=latest
+
+# Load the image into the kind cluster
+make kind-load CONTAINER_TOOL=podman IMG_PREFIX=controller IMG_TAG=latest KIND_CLUSTER=nrr-test
+```
+
+**Verify the image is loaded:**
+```bash
+# Check images in the Kind cluster
+docker exec -it nrr-test-control-plane crictl images | grep controller
```
Priyankasaggu11929: we add a `podman exec ...` version too, since we have a podman variant for all other commands above

docs/TEST_README.md
```diff
@@ -38,21 +38,44 @@ make install
...
+Deploy the controller to the cluster.
+
+**Using Docker:**
 ```bash
 make deploy IMG_PREFIX=controller IMG_TAG=latest
 ```
 
+**Using Podman:**
+```bash
+make deploy IMG_PREFIX=localhost/controller IMG_TAG=latest
```
Priyankasaggu11929, line 76: TIL, that podman forces the `localhost` prefix.

docs/TEST_README.md
Karthik-K-N: Sure, Done.

docs/TEST_README.md
Karthik-K-N: Yes, Thats correct.

Makefile
ajaysundark: I understood `docker-build` as more of a standard target name in Make due to conventions.

Not to sound prudent, and this may be fine to make such changes as we are still in early releases -- but I think we should be careful about changing existing names / fields as it may be breaking change for users.

Karthik-K-N: In that case should I keep the docker build target, Only specific to building docker.

ajaysundark: I think I'm okay with either- using a variable override or a podman-target

Karthik-K-N: Added separate target for docker and podman. Please take a look. Thanks.

.dockerignore
```diff
@@ -1,11 +1,22 @@
 # More info: https://docs.docker.com/engine/reference/builder/#dockerignore-file
-# Ignore everything by default and re-include only needed files
```
ajaysundark, line 2: Is there a reason why including files individually is preferred? 

I remember we had this discussion earlier as well (https://github.com/kubernetes-sigs/node-readiness-controller/pull/113#discussion_r2782559307) and decided current model is more maintainable.

Karthik-K-N: while building for podman, it was failing to consider the files.

Karthik-K-N: I read that the dockerignore parsing for docker and podman is bit different.

ajaysundark: ah okay. we should probably document these development quirks somewhere like a contributing guide.

Karthik-K-N: sure will do

Issue #161, opened 2026-03-11
Install controller in control plane via static pod


ajaysundark:
Support installing the node-readiness-controller via static pods

PR #162, opened 2026-03-11
feat: support static pod installation flow for control-plane nodes

ajaysundark:
## Description
Support installing node-readiness-controller in control-plane via static-pods

## Related Issue
Fixes https://github.com/kubernetes-sigs/node-readiness-controller/issues/161

## Type of Change

/kind documentation
/kind feature

## Testing
Verified in kind cluster using static manifest

## Checklist
- [X] `make test` passes
- [X] `make lint` passes

```release-note
support installing node-readiness-controller via static pods
```

### DIFF
--- a/examples/static-pod/kind-static-pod.yaml
+++ b/examples/static-pod/kind-static-pod.yaml
@@ -1,6 +1,6 @@
 kind: Cluster
 apiVersion: kind.x-k8s.io/v1alpha4
-name: nrr-static-test
+name: nrr-static
 nodes:
 - role: control-plane
   extraMounts:


examples/static-pod/kind-static-pod.yaml
```diff
@@ -0,0 +1,35 @@
+kind: Cluster
+apiVersion: kind.x-k8s.io/v1alpha4
+name: nrr-static-test
```
Karthik-K-N: can we avoid test in the name?

ajaysundark: done

PR #196, opened 2026-05-02
fix: replace hardcoded arch with env detection in scripts

vitorfloriano:
This PR adds a small fix to make install scripts a bit more environment-agnostic. 

The default is a hardcoded amd64, but that can be problematic on runners with different architecture (e.g. arm). 

/kind cleanup


ajaysundark:
Thanks for cleaning this up!

/lgtm
/approve

PR #197, opened 2026-05-02
Add Minikube addon support: labeled CRD and static pod deployment for NRC

### LINKED ISSUE #188: feat: minikube addon for NRC
Node Readiness Controller supports a control-plane co-located / static-pod deployment model that could be preferred for non-disruptive deployments. A [minikube addon](https://minikube.sigs.k8s.io/docs/commands/addons/) will give us a way to test this as a lifecycle-managed addon path where NRC runs before user workloads after the control plane.

Minikube also will help our developer environments eg. for features in our roadmaps such as GPU/TPU node simulation for device-driver readiness related testing is very enabling with minikube vs kind. It'll also help with testing more agentic use-cases in the future with support for running gVisor / kata containers.

This feature will give a `minikube start --addons=nrc` as a one-liner for anyone looking to test / consume NRC.

ref:
- How to add a new addon: https://minikube.sigs.k8s.io/docs/contrib/addons/

Prtm2110:
## Description
This change makes it easy to install the Node Readiness Controller in Minikube. It adds the needed files so you can enable the controller as a Minikube addon with a single command. The update includes a special label for Minikube and  a static pod template.

## Related Issue
Fixes: #188

## Testing

Waiting for minikube#22924 to be merged then this can be tested properly. 

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

Yes. 

```release-note
This PR lets users deploy the Node Readiness Controller as a Minikube addon with a single command, simplifying installation and integration.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,6 +268,13 @@ build-installer: build-manifests-temp ## Generate CRDs and deployment manifests 
 	@echo "Generated dist/install-full.yaml (Features: Metrics, TLS, Webhook - Requires cert-manager)"
 	@echo "Check https://node-readiness-controller.sigs.k8s.io/user-guide/installation.html for installation instructions."
 
+.PHONY: minikube-addon
+minikube-addon: manifests ## Regenerate minikube addon CRD from source.
+	@cp config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml deploy/minikube/crds.yaml
+	@# Patch in the minikube addon label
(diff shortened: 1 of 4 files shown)


AvineshTripathi:
hey @Prtm2110 We already have a PR https://github.com/kubernetes/minikube/pull/22924. I can see deploying YAMLs to be part of the minikube project. Any particular reason we want it in NRC?

prathamhole14:
@AvineshTripathi I completely missed https://github.com/kubernetes/minikube/pull/22924, You are right since a native addon actually needs to be registered directly inside the Minikube codebase for the CLI to recognize it, having it there makes much more sense. Closing this it in favor of `#22924` thanks for the review.

PR #199, opened 2026-05-03
ci: fix lint issues for golangci-lint v2.12.1

### LINKED ISSUE #195: ci: golangci-lint v2.12.1 detected issues
When running the latest version of `golangci-lint` on `main`, the following issues are detected:

```
$ golangci-lint version
golangci-lint has version 2.12.1 built with go1.26.2 from 9aa24e94 on 2026-05-02T10:12:50Z

$ golangci-lint run ./...
cmd/readiness-condition-reporter/main.go:155:40: G704: SSRF via taint analysis (gosec)
        req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint, nil)
                                              ^
cmd/readiness-condition-reporter/main.go:164:24: G704: SSRF via taint analysis (gosec)
        resp, err := client.Do(req)
                              ^
internal/controller/suite_test.go:155:23: G703: Path traversal via taint analysis (gosec)
                if _, err := os.Stat(assets); err == nil {
                                    ^
internal/webhook/nodereadinessgaterule_webhook.go:49:6: Consider preallocating allErrs (prealloc)
        var allErrs field.ErrorList
            ^
4 issues:
* gosec: 3
* prealloc: 1
```
We need to either fix those issues or disable those checks, be it via config or directives in the code.

These issues block upgrading to the latest version of `golangci-lint`, since it will break in CI.

rawadhossain:
This pr resolves the issues reported by `golangci-lint v2.12.1` that block upgrading the linter version in CI.

### Changes
- **G704 (SSRF) — readiness-condition-reporter/main.go**  
  Excluded via `.golangci.yml`. The endpoint comes from the `CHECK_ENDPOINT` env var (deployment config), not user input.

- **G703 (path traversal) — suite_test.go**  
  Excluded via `.golangci.yml`. This is test only code using `KUBEBUILDER_ASSETS` from the envtest setup.

- **prealloc — nodereadinessgaterule_webhook.go**  
  Preallocated `allErrs` (`make(..., 0, 4)`) to address the prealloc linter.

### Why config exclusions instead of `//nolint`
Inline `//nolint:gosec` triggers `nolintlint` errors on v2.9.0 (since the checks don’t exist there). Using `.golangci.yml` keeps things compatible across both versions. So clean results on v2.12.1.

With these fixes, `golangci-lint` can be upgraded in CI without breaking.

### Testing
**Before:**
image

**After:**
image
`make lint` (v2.9.0) and `golangci-lint run ./...` (v2.12.1) are both clean.

### Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/cmd/readiness-condition-reporter/main.go
+++ b/cmd/readiness-condition-reporter/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"os/signal"
 	"syscall"
@@ -58,18 +59,22 @@ func main() {
 	nodeName := os.Getenv(envNodeName)
 	if nodeName == "" {
 		klog.ErrorS(nil, "Environment variable not set", "variable", envNodeName)
+		klog.Flush()
 		os.Exit(1)
 	}
 
 	conditionType := os.Getenv(envConditionType)
 	if conditionType == "" {
 		klog.ErrorS(nil, "Environment variable not set", "variable", envConditionType)
+		klog.Flush()
 		os.Exit(1)
 	}
 
-	checkEndpoint := os.Getenv(envCheckEndpoint)
-	if checkEndpoint == "" {
-		klog.ErrorS(nil, "Environment variable not set", "variable", envCheckEndpoint)
+	checkEndpoint, err := validateCheckEndpoint(os.Getenv(envCheckEndpoint))
+
+	if err != nil {
+		klog.ErrorS(err, "Invalid check endpoint configuration", "variable", envCheckEndpoint)
+		klog.Flush()
 		os.Exit(1)
 	}
 
@@ -90,6 +95,7 @@ func main() {
 	config, err := rest.InClusterConfig()
 	if err != nil {
 		klog.ErrorS(err, "Failed to create in-cluster config")
+		klog.Flush()
 		os.Exit(1)
 	}
 
@@ -104,6 +110,7 @@ func main() {
 	clientset, err := kubernetes.NewForConfig(config)
 	if err != nil {
 		klog.ErrorS(err, "Failed to create client")
+		klog.Flush()
 		os.Exit(1)
 	}
 
@@ -150,9 +157,29 @@ func runCheck(ctx context.Context, httpClient *http.Client, clientset kubernetes
 	}
 }
 
+// validateCheckEndpoint ensures the health check endpoint is a well-formed HTTP(S) URL.
+func validateCheckEndpoint(endpoint string) (string, error) {
+	if endpoint == "" {
+		return "", fmt.Errorf("endpoint cannot be empty")
+	}
+
+	parsed, err := url.Parse(endpoint)
+	if err != nil {
+		return "", fmt.Errorf("invalid endpoint URL: %w", err)
+	}
+	if parsed.Scheme != "http" && parsed.Scheme != "https" {
+		return "", fmt.Errorf("unsupported scheme %q: must be http or https", parsed.Scheme)
+	}
+	if parsed.Host == "" {
+		return "", fmt.Errorf("missing host in endpoint URL")
+	}
+
+	return endpoint, nil
+}
+
 // checkHealth performs an HTTP request to check component health.
 func checkHealth(ctx context.Context, client *http.Client, endpoint string) (*HealthResponse, error) {
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint, nil)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, endpoint, nil) //nolint:gosec // endpoint validated at startup
 	if err != nil {
 		return &HealthResponse{
 			Healthy: false,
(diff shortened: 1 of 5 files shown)


.golangci.yml
```diff
@@ -66,6 +66,14 @@ linters:
           - gosec
         path: "test/utils/"
         text: "G204"
+      - linters:
```
Karthik-K-N: Ins't there any better way apart from excluding it?

rawadhossain: Hey @Karthik-K-N, thanks for reviewing the PR.

I looked into fixing this properly instead of excluding it.
- For G703 `suite_test.go)`, using `filepath.Clean` works fine - no suppression needed.
- For G704 `(main.go)`, I added URL validation (scheme + host), so we’re not accepting arbitrary inputs anymore. But gosec still flags the HTTP calls. looks like its taint analysis doesn’t treat this validation as a sanitizer.

Because of that, some form of suppression is still needed for those two lines.
The tricky part is that the repo is currently using golangci-lint v2.9.0, which doesn’t include G704. So if I add inline `//nolint:gosec`, nolintlint flags it as unused and CI fails.

That’s why I initially went with the `.golangci.yml` exclusion, else the warnings can’t be resolved cleanly with the current lint setup (shared the lint output if helpful).

From here, I see two options:
1. Keep the validation in code and use a scoped `.golangci.yml` exclusion just for G704
2. Upgrade golangci-lint to a version that includes G704, then use inline `//nolint:gosec` with the validation in place

Happy to go with whichever direction you prefer.

image

Karthik-K-N: I think in near future we need to update the golang lint, so I think that would be better, wdyt?

rawadhossain: Yeah I think that sounds better for the long term. 
I can update the lint version in CI and switch to inline suppression using `//nolint:gosec`. That should keep things cleaner going forward.

Let me know if you want me to go ahead with that.

rawadhossain:
Hi @Karthik-K-N, just following up on this. Could you take a look when you get time?

.golangci.yml
Karthik-K-N: apologies for missing it, Yes, I think you can go ahead with that. I feel thats the best approach.

rawadhossain:
Hi @Karthik-K-N, I have made the changes. PTAL. 
Thanks.

cmd/readiness-condition-reporter/main.go
```diff
@@ -67,9 +68,9 @@ func main() {
 		os.Exit(1)
 	}
 
-	checkEndpoint := os.Getenv(envCheckEndpoint)
-	if checkEndpoint == "" {
-		klog.ErrorS(nil, "Environment variable not set", "variable", envCheckEndpoint)
+	checkEndpoint, err := validateCheckEndpoint(os.Getenv(envCheckEndpoint))
+	if err != nil {
+		klog.ErrorS(err, "Invalid check endpoint configuration", "variable", envCheckEndpoint)
 		os.Exit(1)
```
Karthik-K-N, line 78: Do you think we need to klog.Flush() before exit for safer side?

cmd/readiness-condition-reporter/main.go
```diff
@@ -150,9 +151,29 @@ func runCheck(ctx context.Context, httpClient *http.Client, clientset kubernetes
 	}
 }
 
+// validateCheckEndpoint ensures the health check endpoint is a well-formed HTTP(S) URL.
+func validateCheckEndpoint(endpoint string) (string, error) {
+	if endpoint == "" {
+		return "", fmt.Errorf("environment variable not set")
```
Karthik-K-N:
```suggestion
		return "", fmt.Errorf("endpoint cannot be empty")
```

cmd/readiness-condition-reporter/main.go
```diff
@@ -150,9 +151,29 @@ func runCheck(ctx context.Context, httpClient *http.Client, clientset kubernetes
...
+	}
+
+	parsed, err := url.Parse(endpoint)
+	if err != nil {
+		return "", fmt.Errorf("parse URL: %w", err)
```
Karthik-K-N:
```suggestion
		return "", fmt.Errorf("invalid endpoint URL: %w", err)
```

cmd/readiness-condition-reporter/main.go
```diff
@@ -150,9 +151,29 @@ func runCheck(ctx context.Context, httpClient *http.Client, clientset kubernetes
...
+	}
+	if parsed.Scheme != "http" && parsed.Scheme != "https" {
+		return "", fmt.Errorf("unsupported scheme %q, must be http or https", parsed.Scheme)
```
Karthik-K-N:
```suggestion
		return "", fmt.Errorf("unsupported scheme %q: must be http or https", parsed.Scheme)
```

Karthik-K-N (commented):
Changes LGTM, Thanks for working on this
Few suggestion for improving error messages.

cmd/readiness-condition-reporter/main.go
rawadhossain: yes, good point. I think it would be better. I'll then add klog.Flush() before the os.Exit(1) calls.

Karthik-K-N (commented):
LGTM
Please squash the commits to keep the history clean.
Thank you.

rawadhossain:
Squashed the commits. PTAL. 
Thanks.

Karthik-K-N:
Thank you
@ajaysundark please take a look. LGTM from my side.

PR #204, opened 2026-05-04
Remove stale failedNodes entries for deleted nodes

### LINKED ISSUE #203: Stale failedNodes entries remain after node deletion
Noticed that controller logic has a small inconsistency in how deleted nodes are handled in rule status.

When a node is removed from the cluster, `status.nodeEvaluations` gets cleaned up correctly, but `status.failedNodes` can still retain entries for that deleted node. This leads to situations where the rule status continues to report failures for nodes that no longer exist.

From what I can see:
- `cleanupDeletedNodes` filters `nodeEvaluations`
- but `failedNodes` isn’t updated accordingly
- and `updateRuleStatus` persists those stale entries

So over time, failedNodes can accumulate stale entries.

Ideally, both `nodeEvaluations` and `failedNodes` should stay in sync, so once a node is gone, it shouldn’t appear in either.

rawadhossain:
## Description
Prunes stale entries from `status.failedNodes` when nodes are deleted.

Currently, `cleanupDeletedNodes` only filters `nodeEvaluations`, so `failedNodes` can retain entries for nodes that no longer exist. This can cause the rule status to report failures for deleted nodes.

This ensures both `nodeEvaluations` and `failedNodes` stay in sync with the current cluster state.

## Type of Change
/kind bug
/kind cleanup

## Testing
- Added a test to verify `failedNodes` entries are removed after a node is deleted
- Ensured no unintended entries are introduced during cleanup
- Ran:
    - `make test`
    - `make lint`

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

Fixes an issue where `status.failedNodes` could include nodes that were already deleted.

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -215,19 +215,11 @@ func (r *RuleReadinessController) cleanupDeletedNodes(ctx context.Context, rule 
 	}
 
 	// Filter out deleted nodes
-	var newNodeEvaluations []readinessv1alpha1.NodeEvaluation
-	for _, evaluation := range rule.Status.NodeEvaluations {
-		if existingNodes[evaluation.NodeName] {
-			newNodeEvaluations = append(newNodeEvaluations, evaluation)
-		}
-	}
-
-	var newFailedNodes []readinessv1alpha1.NodeFailure
-	for _, failure := range rule.Status.FailedNodes {
-		if existingNodes[failure.NodeName] {
-			newFailedNodes = append(newFailedNodes, failure)
-		}
-	}
+	newNodeEvaluations, newFailedNodes := filterStatusForExistingNodes(
+		existingNodes,
+		rule.Status.NodeEvaluations,
+		rule.Status.FailedNodes,
+	)
 
 	if len(newNodeEvaluations) == len(rule.Status.NodeEvaluations) &&
 		len(newFailedNodes) == len(rule.Status.FailedNodes) {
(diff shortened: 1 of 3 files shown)


internal/controller/nodereadinessrule_controller.go
```diff
@@ -246,12 +254,21 @@ func (r *RuleReadinessController) cleanupDeletedNodes(ctx context.Context, rule
 			}
 		}
 
-		if len(freshNodeEvaluations) == len(fresh.Status.NodeEvaluations) {
+		var freshFailedNodes []readinessv1alpha1.NodeFailure
```
AvineshTripathi: Can we make a utility function for this? we do a check before retry too [here](https://github.com/kubernetes-sigs/node-readiness-controller/pull/204/changes#diff-b857080def3b2fecd3967ca35b1bdc29564dda0b064fa902fa24efcd55471899R232)

rawadhossain: Yes, we can, would be even better.
I initially kept it inline to avoid touching other files. But yeah, you’re right, same pattern is used before retry as well.

Committing the changes now.

rawadhossain:
Hi @AvineshTripathi, pushed the changes and extracted it into a utility function. Could you take a look when you get time?

ajaysundark:
/lgtm

@rawadhossain Thanks for raising + fixing these stale status issues in NRC.
JFYI, we are looking into separating 'node evaluation status' from Rule object for scalability. So likely will need to revisit these in the future.

/cc @Karthik-K-N

ajaysundark:
/hold

will wait for @AvineshTripathi  approval to merge

rawadhossain:
> /lgtm
> 
> @rawadhossain Thanks for raising + fixing these stale status issues in NRC. JFYI, we are looking into separating 'node evaluation status' from Rule object for scalability. So likely will need to revisit these in the future.
> 
> /cc @Karthik-K-N

Sure, I understand.

ajaysundark:
/lgtm 

@rawadhossain sorry for the delay on this one. can you rebase?

rawadhossain:
@ajaysundark rebased and resolved the conflicts. Should be good now.

ajaysundark (approved):

PR #206, opened 2026-05-04
fix: avoid double-counting bootstrap completion metric

### LINKED ISSUE #205: Bootstrap completed metric can be overcounted
## What happened?

`node_readiness_bootstrap_completed_total` can be incremented even when the controller does not newly mark a node as bootstrap completed.

The metric is declared in `internal/metrics/metrics.go` and incremented from `markBootstrapCompleted()` in `internal/controller/node_controller.go`.

`markBootstrapCompleted()` already has an idempotency check that returns early from the retry function when the node has the `readiness.k8s.io/bootstrap-completed-<rule>` annotation. However, after the retry block succeeds, the function increments `metrics.BootstrapCompleted` regardless of whether the annotation was newly written or already present.

That means repeated calls for the same node/rule can overcount `node_readiness_bootstrap_completed_total`.

## What did you expect to happen?

The bootstrap completion metric should increment only when the controller newly marks a node as bootstrap completed for a rule.

If the bootstrap-completed annotation already exists and the function performs no patch, the metric should not increment again.

## Why does this matter?

The monitoring docs describe `node_readiness_bootstrap_completed_total` as being recorded when the controller marks bootstrap as completed for a node under a bootstrap-only rule. Counting no-op calls can make the metric higher than the number of actual bootstrap completions observed by the controller.

## Suggested fix

Track whether `markBootstrapCompleted()` actually wrote the bootstrap-completed annotation. Increment `metrics.BootstrapCompleted` only after a successful patch that newly adds the annotation.

A regression test can call `markBootstrapCompleted()` twice for the same node/rule and verify that the metric increases by one, not two.

/kind bug

sahitya-chandra:
## Description

Fixes overcounting of `node_readiness_bootstrap_completed_total`.

`markBootstrapCompleted()` already skips patching a node when the bootstrap-completed annotation exists, but the metric was still incremented after that no-op path. This change tracks whether the annotation was newly written and increments `metrics.BootstrapCompleted` only after a successful new mark.

Adds a regression test that calls `markBootstrapCompleted()` twice for the same node/rule and verifies the metric increases only once.

## Related Issue

Fixes #205

## Type of Change

/kind bug

## Testing

```bash
make test
make lint
```

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/onsi/ginkgo/v2 v2.27.2
 	github.com/onsi/gomega v1.38.2
 	github.com/prometheus/client_golang v1.22.0
+	github.com/prometheus/client_model v0.6.1
 	go.uber.org/zap v1.27.0
 	k8s.io/api v0.34.0
 	k8s.io/apimachinery v0.34.0
@@ -52,7 +53,6 @@ require (
 	github.com/munnerz/goautoneg v0.0.0-20191010083416-a7dc8b61c822 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/prometheus/client_model v0.6.1 // indirect
 	github.com/prometheus/common v0.62.0 // indirect
 	github.com/prometheus/procfs v0.15.1 // indirect
 	github.com/spf13/cobra v1.9.1 // indirect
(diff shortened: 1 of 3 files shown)


AvineshTripathi:
/lgtm
Thanks for fixing this!

ajaysundark (approved):
Thanks!

PR #208, opened 2026-05-04
Fix re-taint in bootstrap-only mode when annotation fails

### LINKED ISSUE #207: Bootstrap-only mode may reapply taint if bootstrap annotation fails
Noticed an issue with bootstrap-only mode when the bootstrap completion annotation fails to be written.

When conditions are satisfied, the controller removes the taint and marks bootstrap as completed via a node annotation. If that annotation write fails, the taint is already removed but completion is never recorded.

Looking at the flow:

- `markBootstrapCompleted` is only called during taint removal
- if it fails, later reconciles may not call it again while the taint stays absent
- so the controller has no completion marker on the node

Because of that, if conditions change later, the controller can reapply the taint — even though bootstrap-only mode should not reapply after the first success.

Once the taint has been removed in bootstrap-only mode, it shouldn’t be added again just because the annotation step didn’t succeed.

rawadhossain:
## Description
Fixes an issue where nodes can get re-tainted in bootstrap-only mode if the bootstrap completion annotation fails to be written.

Currently, the controller removes the taint but doesn’t record completion if the annotation write fails. This can cause the controller to treat the node as incomplete and reapply the taint on later reconciles.

To fix this:
- made sure errors from the annotation write aren’t ignored
- kept NodeEvaluation in sync even if the annotation step fails
- skipped nodes that are already marked as bootstrap-completed during rule reconciliation

This keeps bootstrap-only behavior consistent after the first successful taint removal.

## Type of Change
/kind bug
/kind cleanup

## Testing
- Added tests to cover:
  - annotation patch failure after taint removal
  - skipping evaluation when bootstrap is already completed
- Ran:
    - `make test`
    - `make lint`

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -316,6 +316,8 @@ func (r *RuleReadinessController) removeTaintBySpec(ctx context.Context, node *c
 	})
 }
 
+const bootstrapAnnotationKeyPrefix = "readiness.k8s.io/bootstrap-completed-"
+
 // Bootstrap completion tracking.
 func (r *RuleReadinessController) isBootstrapCompleted(ctx context.Context, nodeName, ruleName string) bool {
 	// Check node annotation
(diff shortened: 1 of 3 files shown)


ajaysundark:
@rawadhossain Could you re-evaluate this patch on whether this fix is still relevant after https://github.com/kubernetes-sigs/node-readiness-controller/pull/417 (now merged)?

This also needs rebase.

rawadhossain:
Yeah, took a look, looks like #417 already covers this. Since the taint removal and annotation write are now done atomically in a single patch, the case this PR was handling shouldn’t happen anymore.

I’ll close this one out.

PR #200, opened 2026-05-03
CI: harden GitHub Actions workflows security

vitorfloriano:
This PR hardens the security on GitHub Actions workflows.

Summary of changes:

- Add `persist-credentials: false` to checkout action to remediate [artipacked](https://docs.zizmor.sh/audits/#artipacked).
- Change top-level permissions to `{}` (none) and grant necessary permissions in job-level, as needed, to remediate [excessive permissions](https://docs.zizmor.sh/audits/#excessive-permissions).
- Pin `govulncheck` install to commit hash.

Issue #151, opened 2026-03-05
Automate manual steps in release with a repeatable process


ajaysundark:
Currently release process (RELEASE.md) requires the Release Shepherd to generate deployment manifests and then verify them with tests before uploading them to the GitHub Release.

We could automate possible repeatable sections in this process using a post-submit GitHub Action that triggers when a new version tag is pushed.

Some ideas for exploring:

A scheduled cronjob that -
1. checks the latest VERSION tag available in k8s-registry 
2. generate the release manifests
3. run test-e2e with prepared manifests
4. create gh action release (in draft)

This could be further utilized with [krel](https://github.com/kubernetes/release/blob/master/docs/krel/README.md) to add some structure to release automation by auto-generating release notes to a `previous-releases/vX.Y.Z/release-notes.md` target. This could be a `create-new-minor-release.sh` script to make changes to create the release-cut PR. The automation will utilize the metadata for release.  

This would reduce the release toil while also ensuring a repeatable controlled process for the nrc releases.

cc @Priyankasaggu11929 we could break it down to subtasks and see how we could create good-first-issues for the project

Yuan-prog:
Hi @ajaysundark, this looks like a great direction.

I’m interested in helping with this. If this gets broken down into smaller pieces, I’d be happy to take one of the subtasks and contribute step by step.

Let me know how you’d like to structure it.

Karman580:
Hi @ajaysundark, thanks for the clarification regarding the LFX mentorship scope.

I’m Karman Singh Talwar, and I’ve been going through the release workflow and related discussions/PRs around artifact generation, E2E validation, release metadata, and automation opportunities. The direction for building a more repeatable and structured release process is very interesting, especially the potential integration with tooling like krel.

I’m highly interested in contributing to this area as part of the LFX mentorship cycle and will continue exploring the repository, release flow, and adjacent issues to better understand the ecosystem and contribute where appropriate.

Looking forward to learning from the discussions and engaging further with the project.

PR #210, opened 2026-05-04
ci: add zizmor for linting gha workflows

vitorfloriano:
This PR adds [Zizmor](https://github.com/zizmorcore/zizmor) to CI.

In short, Zizmor is a Static Analysis Tool that scans GitHub Actions workflows to detect security issues like unpinned version usage, excessive permissions and [many other issues](https://docs.zizmor.sh/audits/).

It will help us ensure existent and added workflows adhere to security best practices, including Kubernetes new GitHub Actions Security Policy.

About the chosen config:

- We set `advanced-security: false` to disable uploading results to GitHub's Advanced Security tab but instead print to the console and fail the run on CI if issues are found. See: https://github.com/zizmorcore/zizmor-action#advanced-security
- We set `min-confidence: medium` to ignore findings that are merely informational due to low confidence of the audit, just so they don't fail the run. See: https://github.com/zizmorcore/zizmor-action#min-confidence

Blocked by #200

### DIFF
--- a/.github/workflows/lint-workflows.yml
+++ b/.github/workflows/lint-workflows.yml
@@ -1,0 +1,26 @@
+name: Lint GitHub Actions Workflows
+on:
+  pull_request:
+    paths: 
+    - '.github/workflows/*.yml'
+    - '.github/workflows/*.yaml'
+
+permissions: {}
+
+jobs:
+  zizmor:
+    permissions:
+      contents: read
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@b1d7e1fb5de872772f31590499237e7cce841e8e # v0.5.3
+        with:
+          advanced-security: false
+          min-confidence: medium
+          


ajaysundark:
/cc @Priyankasaggu11929 

Could you help take a look at this? Thanks!

.github/workflows/lint-workflows.yml
```diff
@@ -0,0 +1,25 @@
+name: Lint GitHub Actions Workflows
+on:
+  pull_request:
+    paths: 
+     - '**/workflows/*.yml'
```
ajaysundark: could this be restricted to `.github/workflows/**` instead?

ajaysundark: I think we could prefer path filter to be `.github/workflows/**/*.{yml, yaml}`

vitorfloriano: Yeah, `**/workflows/` works best for monorepos, but that's not the case here. I'll change that.

vitorfloriano: I think GitHub Actions only recognizes the files directly under `workflows/` (without additional nesting), so I believe it should be like this:

```suggestion
     - '.github/workflows/*.{yml,yaml}'
```
<br>

See: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#about-yaml-syntax-for-workflows 

>You must store workflow files in the .github/workflows directory of your repository.

I'll update based on that.

vitorfloriano:
Hm...the workflow linter did not run after the latest change. Need to investigate.

/hold

.github/workflows/lint-workflows.yml
```diff
@@ -0,0 +1,26 @@
+name: Lint GitHub Actions Workflows
+on:
+  pull_request:
+    paths: 
+    - '.github/workflows/*.yml'
+    - '.github/workflows/*.yaml'
```
vitorfloriano, line 6: @ajaysundark Apparently, `*.{yml, yaml}` isn't valid syntax in workflows, so we'll need to be a bit ~~redundant~~ explicit here.

See: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet

Priyankasaggu11929:
Folks, just a thought - rather than introducing more new tools to ensure our GH actions are safe, could we try moving our existing GitHub Action workflows to Prow (I can help guide here)?

Priyankasaggu11929:
> Folks, just a thought - rather than introducing more new tools to ensure our GH actions are safe, could we try moving our existing GitHub Action workflows to Prow (I can help guide here)?

Hello @vitorfloriano, following up from our discussion from yesterday's call.

For learning about Prow/Prowjobs, here are some resources:
- prow job docs - https://docs.prow.k8s.io/docs/jobs/
- and this talk - https://www.youtube.com/watch?v=gKKfXTZd7VU

---

We already have a directory setup for defining Prow jobs for the Node Readiness Controller (NRC) repo here - https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/node-readiness-controller

This is the PR that added the first (presubmit) job for the NRC project - https://github.com/kubernetes/test-infra/pull/36261.  
The review comments might be helpful.


---

And for some guidance/reference on migrating our existing NRC GH action worfklow jobs, we can maybe look at the prow jobs definition for NPD project - https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes/node-problem-detector

We can start by migrating one of the GH jobs to Prow.
Please feel free to start a draft PR and use that for any queries/doubts, I can help. Thanks!

vitorfloriano:
>We can start by migrating one of the GH jobs to Prow.
Please feel free to start a draft PR and use that for any queries/doubts, I can help. Thanks!

@Priyankasaggu11929 I drafted https://github.com/kubernetes/test-infra/pull/37035.

I decided to start with migrating the [test workflow](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/.github/workflows/test.yml) because it seems to be the easiest. 

Also, I think that migrating the [release workflow](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/.github/workflows/release-automation.yml) would enter the scope of the upcoming mentorship project (see https://github.com/kubernetes-sigs/node-readiness-controller/issues/151), so I'll leave it for now. 

Considering we are moving towards a Prow-centered CI, I'll close this PR. 

We can continue the conversation on the prowjob PR and in future meetings.

Thanks for the guidance!

PR #212, opened 2026-05-05
webhook: improve nodeSelectorsOverlap to detect subset overlaps

### LINKED ISSUE #211: webhook: nodeSelectorsOverlap only detects identical selectors, missing partial overlaps
## Describe the bug

In `internal/webhook/nodereadinessgaterule_webhook.go`, the 
`nodeSelectorsOverlap` function uses string equality to detect overlapping 
node selectors:

```go
return sel1.String() == sel2.String()
```

This means two rules with **different but overlapping** selectors 
(e.g. `env=prod` and `env=prod,region=us`) are not detected as 
conflicting, even though both rules would match nodes labeled 
`env=prod,region=us` and apply the same taint — causing conflicts.

## Steps to reproduce

1. Create rule A with nodeSelector `env=prod`
2. Create rule B with nodeSelector `env=prod,region=us` and same taint key
3. Webhook allows both rules despite overlapping selectors
4. Both rules apply to nodes matching `env=prod,region=us`

## Expected behavior
Webhook should detect partial selector overlaps and reject conflicting rules.

## Actual behavior
Only identical selectors are detected. Partial overlaps are missed.

## File
`internal/webhook/nodereadinessgaterule_webhook.go` line 127

## Are you able to fix this issue?
Yes (I will propose a PR)

Shreya2005-2005:
## Description
Improves the nodeSelectorsOverlap function in the validation webhook to detect partial/subset selector overlaps using sel.Matches() instead of string equality comparison.

## Related Issue
Fixes #211

## Type of Change
/kind bug

## Testing
- All webhook tests pass: go test ./internal/webhook/...
- Verified build compiles cleanly: go build ./...

## Checklist
- [x] `make test` passes (webhook tests)
- [x] `make lint` passes

```release-note
Fixed webhook validation to correctly detect overlapping node selectors when one selector is a subset of another, preventing conflicting NodeReadinessRules from being created.
```

### DIFF
--- a/internal/webhook/nodereadinessgaterule_webhook.go
+++ b/internal/webhook/nodereadinessgaterule_webhook.go
@@ -22,6 +22,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -122,9 +123,12 @@ func (w *NodeReadinessRuleWebhook) nodeSelectorsOverlap(selector1, selector2 met
 		return true
 	}
 
-	// Simple heuristic: if selectors are identical, they definitely overlap
-	// For more complex overlap detection, we'd need to analyze the label requirements
-	return sel1.String() == sel2.String()
+	// Check overlap: if one selector's matchLabels is a subset of the other,
+	// a node could match both selectors, causing a conflict.
+	if sel1.Matches(labels.Set(selector2.MatchLabels)) {
+		return true
+	}
+	return sel2.Matches(labels.Set(selector1.MatchLabels))
 }
 
 // generateNoExecuteWarnings generates admission warnings for NoExecute taint usage.
--- a/internal/webhook/nodereadinessgaterule_webhook.go
+++ b/internal/webhook/nodereadinessgaterule_webhook.go
@@ -125,6 +125,8 @@ func (w *NodeReadinessRuleWebhook) nodeSelectorsOverlap(selector1, selector2 met
 
 	// Check overlap: if one selector's matchLabels is a subset of the other,
 	// a node could match both selectors, causing a conflict.
+	// Note: this only covers matchLabels-based overlap; selectors using
+	// matchExpressions may still overlap without being detected here.
 	if sel1.Matches(labels.Set(selector2.MatchLabels)) {
 		return true
 	}
--- a/internal/webhook/nodereadinessgaterule_webhook_test.go
+++ b/internal/webhook/nodereadinessgaterule_webhook_test.go
@@ -283,7 +283,7 @@ var _ = Describe("NodeReadinessRule Validation Webhook", func() {
 			Expect(overlaps).To(BeTrue()) // Both nil = both match all nodes
 		})
 
-		It("should not overlap when one selector is nil", func() {
+		It("should overlap when one selector is empty (matches all nodes)", func() {
 			selector := metav1.LabelSelector{
 				MatchLabels: map[string]string{
 					"node-role.kubernetes.io/worker": "",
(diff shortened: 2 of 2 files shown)


Shreya2005-2005:
@ajaysundark review this whenever you get a chance

internal/webhook/nodereadinessgaterule_webhook_test.go
```diff
@@ -283,18 +283,18 @@ var _ = Describe("NodeReadinessRule Validation Webhook", func() {
 			Expect(overlaps).To(BeTrue()) // Both nil = both match all nodes
 		})
 
-		It("should not overlap when one selector is nil", func() {
```
ajaysundark, line 286: we already have an empty selector validation.  Could you check if this test case is covered?

internal/webhook/nodereadinessgaterule_webhook_test.go
```diff
@@ -283,18 +283,18 @@ var _ = Describe("NodeReadinessRule Validation Webhook", func() {
 			Expect(overlaps).To(BeTrue()) // Both nil = both match all nodes
```
ajaysundark, line 283: Could we add a test-case for the subset scenario?

internal/webhook/nodereadinessgaterule_webhook.go
```diff
@@ -122,9 +123,12 @@ func (w *NodeReadinessRuleWebhook) nodeSelectorsOverlap(selector1, selector2 met
 		return true
 	}
 
-	// Simple heuristic: if selectors are identical, they definitely overlap
-	// For more complex overlap detection, we'd need to analyze the label requirements
-	return sel1.String() == sel2.String()
+	// Check overlap: if one selector's matchLabels is a subset of the other,
```
ajaysundark, line 126: Thanks, this is a nice improvement. But this is not a complete fix as it will not help the scenario when matchExpressions were used for the labels. Did you get a chance to check if there's any helper in api-machinery for conflict detection that'd cover this case?

ajaysundark: looks like there is no such primitive for label-selector overlap detection: https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/labels/selector.go#L123.

I think a `SelectorsOverlap(s1, s2 Selector` function there would benefit any controller that needs to validate non-conflicting labels.

cc @kubernetes-sigs/kubernetes-sig-api-machinery 

@Shreya2005-2005 for the scope of this PR, this change is acceptable for me. maybe we could add a comment capturing the limitation.

ajaysundark: cc @jpbetz / @siyuanfoundation does such a reusable primitive for conflict check belong upstream in k8s.io/apimachinery/pkg/labels?

ajaysundark (changes requested):
Thanks for your patch, could you please also add unit-tests for the scenario this PR aim to fix?

internal/webhook/nodereadinessgaterule_webhook.go
siyuanfoundation: I think so. Feel free to add a Overlap function in k8s.io/apimachinery/pkg/labels

sahitya-chandra: > I think so. Feel free to add a Overlap function in k8s.io/apimachinery/pkg/labels

can I go on and add an upstream overlap function?

Cc @siyuanfoundation @ajaysundark

internal/webhook/nodereadinessgaterule_webhook_test.go
Shreya2005-2005: Already covered — existing test "should overlap when one selector is empty (matches all nodes)" verifies this behavior.

internal/webhook/nodereadinessgaterule_webhook_test.go
Shreya2005-2005: Added test case "should detect subset selectors as overlapping" covering env=prod overlapping env=prod,region=us scenario.

internal/webhook/nodereadinessgaterule_webhook.go
Shreya2005-2005: Added comment noting that matchExpressions-based overlaps are not detected by this implementation.

Shreya2005-2005:
@ajaysundark whenever you get a change review the changes made

PR #214, opened 2026-05-05
fix: trigger reconcile on rule deletion-timestamp updates

### LINKED ISSUE #213: GenerationChangedPredicate prevents reconcileDelete from running on rules with finalizers
## What happened

`RuleReconciler` is set up with a bare `predicate.GenerationChangedPredicate{}` on its primary watch:

https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/nodereadinessrule_controller.go#L82-L88

```go
return ctrl.NewControllerManagedBy(mgr).
    Named("nodereadiness-controller").
    WithOptions(controller.Options{MaxConcurrentReconciles: 1}).
    For(&readinessv1alpha1.NodeReadinessRule{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
    Complete(r)
```

`NodeReadinessRule` has a finalizer (`readiness.node.x-k8s.io/cleanup-taints`). When a user deletes a rule, the API server only sets `metadata.deletionTimestamp`, it does **not** bump `metadata.generation`, because deletion is not a spec change. The watch fires an Update event with `old.Generation == new.Generation`, so `GenerationChangedPredicate` filters it out and `Reconcile` is never enqueued.

The result: `reconcileDelete` (which removes the managed taints from nodes and then removes the finalizer) never runs, and the rule sits in `Terminating` indefinitely. Cache resync also doesn't help because resync produces a synthetic Update with equal generations.

Deleting a `NodeReadinessRule` should trigger `reconcileDelete`, taints managed by the rule should be cleaned up from matching nodes, the finalizer should be removed, and the rule should be fully deleted.

## How to reproduce

1. Create a `NodeReadinessRule` and let the controller add the finalizer.
2. `kubectl delete nodereadinessrule <name>`.
3. Observe that the rule stays in `Terminating` and the managed taint is not removed from any node. No reconcile is logged for the deletion.

The existing unit test at `internal/controller/nodereadinessrule_controller_test.go` line 824 (\"should remove taints from nodes when rule is deleted\") doesn't catch this because it calls `Reconcile` directly, bypassing the predicate.

## Suggested fix

Combine `GenerationChangedPredicate` with a predicate that also passes Update events where the deletion timestamp transitions from unset to set, e.g. `predicate.Or(predicate.GenerationChangedPredicate{}, <deletion-timestamp predicate>)`. Or drop `GenerationChangedPredicate` for the primary resource since the predicate is incompatible with finalizer-based deletion on its own.

/kind bug

sahitya-chandra:
## Description

`RuleReconciler` was set up with a bare `predicate.GenerationChangedPredicate{}`. When a `NodeReadinessRule` with the cleanup-taints finalizer is deleted, the API server only sets `metadata.deletionTimestamp` - it does not bump `metadata.generation`. The predicate filtered out that Update event, so `Reconcile` was never enqueued, `reconcileDelete` never ran, the finalizer was never removed, and the managed taints were left orphaned on nodes

Replace the predicate with one that also passes Update events where `DeletionTimestamp` transitions from unset to set, while keeping the Generation-based filtering for normal updates so status writes still don't trigger reconciles

The existing deletion test calls `Reconcile` directly and bypasses the predicate, so it didn't catch this. Added unit coverage that exercises the predicate so a regression to a bare `GenerationChangedPredicate` fails the build

## Related Issue

Fixes #213

## Type of Change

/kind bug

## Testing

- `make test` passes locally (envtest, Kubernetes 1.34)
- `make lint` passes locally
- New `rulePredicate` unit tests cover: generation bump, deletion-timestamp transition, status-only update (filtered), update while already terminating (filtered)

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fix a bug where deleting a NodeReadinessRule with the cleanup-taints finalizer left the rule stuck in Terminating and the managed taint orphaned on nodes, because the controller's watch predicate filtered out deletion-timestamp updates
```

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 
 	readinessv1alpha1 "sigs.k8s.io/node-readiness-controller/api/v1alpha1"
@@ -78,12 +79,27 @@ func NewRuleReadinessController(mgr ctrl.Manager, clientset kubernetes.Interface
 	}
 }
 
+// Like GenerationChangedPredicate, but also fires when DeletionTimestamp is
+// set so finalizer deletion events aren't dropped.
+var rulePredicate = predicate.Funcs{
+	UpdateFunc: func(e event.UpdateEvent) bool {
+		if e.ObjectOld == nil || e.ObjectNew == nil {
+			return false
+		}
+		if e.ObjectNew.GetGeneration() != e.ObjectOld.GetGeneration() {
+			return true
+		}
+		return e.ObjectOld.GetDeletionTimestamp().IsZero() &&
+			!e.ObjectNew.GetDeletionTimestamp().IsZero()
+	},
+}
+
 // SetupWithManager sets up the controller with the Manager.
 func (r *RuleReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		Named("nodereadiness-controller").
 		WithOptions(controller.Options{MaxConcurrentReconciles: 1}).
-		For(&readinessv1alpha1.NodeReadinessRule{}, builder.WithPredicates(predicate.GenerationChangedPredicate{})).
+		For(&readinessv1alpha1.NodeReadinessRule{}, builder.WithPredicates(rulePredicate)).
 		Complete(r)
 }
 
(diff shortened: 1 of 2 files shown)


internal/controller/nodereadinessrule_controller.go
```diff
@@ -78,12 +79,27 @@ func NewRuleReadinessController(mgr ctrl.Manager, clientset kubernetes.Interface
 	}
 }
 
+// rulePredicate behaves like GenerationChangedPredicate, but also allows the
+// update that first sets DeletionTimestamp so finalizer cleanup is reconciled.
+var rulePredicate = predicate.Funcs{
```
sahitya-chandra, line 84: Self-review: this could also be expressed as `predicate.Or(predicate.GenerationChangedPredicate{}, deletionTimestampPredicate)`. I kept it as a single predicate because it's only ~10 lines and avoids a one-shot helper, but happy to switch to composition if maintainers prefer using the controller-runtime built-in

sahitya-chandra:
@ajaysundark sir please take a look at this whenever you get time, thanks

ajaysundark:
xref: https://github.com/kubernetes-sigs/node-readiness-controller/issues/213#issuecomment-4433160784 this is not required as API server manages the generations for deletion timestamp correctly-  https://github.com/kubernetes/kubernetes/issues/24935

PR #216, opened 2026-05-05
fix: only append to appliedNodes after successful node evaluation

### LINKED ISSUE #215: Nodes appear in both `appliedNodes` and `failedNodes` on evaluation failure
## Description:

Came across an inconsistency between the API docs and how `NodeReadinessRule` status is updated.

`status.appliedNodes` is described as the set of nodes where the taint was successfully managed. But, if evaluation fails, the node can still end up in `appliedNodes`, also being recorded in `failedNodes`.

So the same node ends up in both lists during the same reconcile.

**Expected:** If `evaluateRuleForNode` returns an error, the node should not be considered successfully managed, and should only appear in `failedNodes`.

**Actual:** The node can appear in both `appliedNodes` and `failedNodes`.

**Why this happens:**

- node matches `ruleAppliesTo` → added to `appliedNodes`
- then `evaluateRuleForNode` runs
- if it fails, `recordNodeFailure` adds it to `failedNodes`
- but the earlier append to `appliedNodes` is never removed

There’s also a related case where a node that previously failed but later succeeds can still have a stale entry in `failedNodes`, so it continues to appear in both lists even after recovery.

rawadhossain:
## Description
This change prevents nodes from appearing in both `status.appliedNodes` and `status.failedNodes` during rule reconciliation.

Previously, nodes were added to `appliedNodes` before `evaluateRuleForNode` ran. If evaluation fails, the same node ends up in `failedNodes` as well. so it shows up in both lists in a single reconcile.

There’s also a related case where a node that previously failed but later succeeds can still have a stale entry in `failedNodes`, so it continues to appear in both lists even after recovery.

To fix this:
- only add nodes to `appliedNodes` when `evaluateRuleForNode` succeeds
- remove any existing `failedNodes` entry for a node when it evaluates successfully
- add tests covering both failure and recovery scenarios

This keeps `appliedNodes` aligned with its intended meaning (nodes where the taint was successfully managed) and avoids overlap with `failedNodes`.

## Type of Change
/kind bug

## Testing
- Added tests to cover:
  - evaluation failure: node is not added to `appliedNodes` and appears in `failedNodes`
  - recovery case: stale `failedNodes` entry is removed and node is added to `appliedNodes`
- Ran:
  - `make test`
  - `make lint`

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -270,13 +270,20 @@ func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, ru
 	var appliedNodes []string
 	for _, node := range nodeList.Items {
 		if r.ruleAppliesTo(ctx, rule, &node) {
-			appliedNodes = append(appliedNodes, node.Name)
 			log.Info("Processing node for rule", "rule", rule.Name, "node", node.Name)
 			if err := r.evaluateRuleForNode(ctx, rule, &node); err != nil {
-				// Log error but continue with other nodes
 				log.Error(err, "Failed to evaluate node for rule", "rule", rule.Name, "node", node.Name)
 				r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
 				metrics.Failures.WithLabelValues(rule.Name, "EvaluationError").Inc()
+			} else {
+				appliedNodes = append(appliedNodes, node.Name)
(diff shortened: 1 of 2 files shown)


rawadhossain:
@ajaysundark could you please take a look?

ajaysundark:
/lgtm
/approve

Thanks for submitting the patch!

PR #133, opened 2026-02-20
:sparkles: Add nodereadinessrulereport CRD and controller

Karthik-K-N:
## Description

This PR creates  nodereadinessrulereport CRD and controller 
```

# kubectl get nrrp
NAME                                NODE                     MATCHED RULES   UNMATCHED RULES   APPLIED TAINTS   ERRORS   AGE
nrr-report-nrr-test-control-plane   nrr-test-control-plane   0               1                 0                0        82s
nrr-report-nrr-test-worker          nrr-test-worker          0               1                 1                0        82s
nrr-report-nrr-test-worker2         nrr-test-worker2         0               1                 1                0        82s


# kubectl apply -f examples/cni-readiness/network-readiness-rule.yaml
nodereadinessrule.readiness.node.x-k8s.io/network-readiness-rule unchanged


# kubectl get nrrp
NAME                                NODE                     MATCHED RULES   UNMATCHED RULES   APPLIED TAINTS   ERRORS   AGE
nrr-report-nrr-test-control-plane   nrr-test-control-plane   0               1                 0                0        3m33s
nrr-report-nrr-test-worker          nrr-test-worker          1               0                 0                0        3m33s
nrr-report-nrr-test-worker2         nrr-test-worker2         1               0                 0                0        3m33s


# kubectl get nrrp nrr-report-nrr-test-worker2 -o yaml
apiVersion: readiness.node.x-k8s.io/v1alpha1
kind: NodeReadinessRuleReport
metadata:
  creationTimestamp: "2026-02-20T13:05:33Z"
  generation: 1
  name: nrr-report-nrr-test-worker2
  ownerReferences:
  - apiVersion: v1
    blockOwnerDeletion: true
    controller: true
    kind: Node
    name: nrr-test-worker2
    uid: 8701cb81-06fa-4e92-9d82-5de9d6ef0c99
  resourceVersion: "1671"
  uid: db8a7abb-ecc0-4a6d-93f5-97ca3cbc6f53
spec:
  nodeName: nrr-test-worker2
status:
  readinessReports:
  - lastEvaluationTime: "2026-02-20T13:09:38Z"
    message: Node successfully matches all rule criteria.
    reason: CriteriaMet
    ruleName: network-readiness-rule
    ruleStatus: Matched
    taintStatus: Absent
  summary:
    appliedTaints: 0
    errors: 0
    matchedRules: 1
    unMatchedRules: 0

```

## Related Issue

[...]

PR #218, opened 2026-05-06
fix: skip NodeEvaluation upsert when evaluation did not run

### LINKED ISSUE #217: processNodeAgainstAllRules can write an empty NodeEvaluation that fails CRD validation and drops FailedNodes updates
## What happens

On a failed evaluation (`evaluateRuleForNode` returns an error, e.g. a non-retryable taint patch failure), `processNodeAgainstAllRules` still appends a zero-value `NodeEvaluation{}` to the rule-status patch:

https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/node_controller.go#L159-L206

Every required field on `NodeEvaluation` (`nodeName`, `conditionResults`, `taintStatus`, `lastEvaluationTime`) is `omitempty`, so the empty entry serializes to `{}` and the apiserver rejects the entire `Status().Patch` with a 422:

```
status.nodeEvaluations[0].conditionResults: Required value,
status.nodeEvaluations[0].lastEvaluationTime: Required value,
status.nodeEvaluations[0].nodeName: Required value,
status.nodeEvaluations[0].taintStatus: Required value
```

The patch is atomic, and the `FailedNodes` update (recorded via `recordNodeFailure`) is bundled into the same patch, so it is rejected too. Net effect: on a failed evaluation nothing is persisted to `rule.status`, and the only signal is a controller log line. Confirmed via envtest.

## Expected

Skip the empty `NodeEvaluation` on the failure path and still persist the `FailedNodes` entry, so the failure is visible via `kubectl get nodereadinessrule -o yaml`.

## Reproduce

Make a node's taint patch fail persistently with a non-conflict error (e.g. RBAC-deny the `nodes` patch) so `evaluateRuleForNode` returns an error: the controller logs the 422 above and `failedNodes` never appears for that node.

/kind bug

sahitya-chandra:
## Description

**The bug:** When adding or removing a node's taint fails with a non-retryable error (for example an RBAC denial or a persistent API error), `evaluateRuleForNode` returns early without recording an evaluation for that node. `processNodeAgainstAllRules` then still tried to save a `NodeEvaluation` for it, and with nothing recorded it saved an empty one. Every required field on that type uses `omitempty`, so an empty value serializes to `{}` and the API server rejects the entire status patch with a 422 (`Required value` for `nodeName`, `conditionResults`, `taintStatus`, and `lastEvaluationTime`). The empty {} NodeEvaluation validation failure occurs when the node has no cached NodeEvaluation. If an existing cached entry is present, the old code avoids the empty object but can still overwrite fresher persisted status with stale cached evaluation data. The `FailedNodes` update rides in that same patch, so it is dropped too. The failure then never appears in `kubectl get nodereadinessrule -o yaml`, and the only trace is a controller log line.

**The fix:** Save the `NodeEvaluation` only when the evaluation succeeded. On the failure path, leave any existing `NodeEvaluation` untouched. Since FailedNodes rides in the same `Status().Patch()` call, omitting the invalid empty NodeEvaluation prevents the 422 that was silently dropping it.

## Related Issue

Fixes #217

## Type of Change

/kind bug

## Testing

- `make test` passes locally
- `make lint` passes locally
- Regression test 1: a fake client fails `Patch` on the node, runs `NodeReconciler.Reconcile`, and asserts that the `FailedNodes` entry lands, no empty `NodeEvaluation` slipped in, and an unrelated pre-existing `NodeEvaluation` was preserved. Without the fix, the test fails on the empty-`NodeName` assertion
- Regression test 2 covers the stale-cache case: persisted status has `TaintStatus=Absent`, the cached rule snapshot has stale `TaintStatus=Present`, evaluation fails, and the persisted entry must stay `Absent`

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fix a bug where a transient taint patch failure on a node could drop the corresponding FailedNodes entry from a NodeReadinessRule's status, or overwrite a fresh persisted NodeEvaluation with a stale one from the controller's rule cache
```

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -147,12 +147,13 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			"rule", rule.Name,
 			"ruleResourceVersion", rule.ResourceVersion)
 
-		if err := r.evaluateRuleForNode(ctx, rule, node); err != nil {
-			log.Error(err, "Failed to evaluate rule for node",
+		evalErr := r.evaluateRuleForNode(ctx, rule, node)
+		if evalErr != nil {
+			log.Error(evalErr, "Failed to evaluate rule for node",
 				"node", node.Name, "rule", rule.Name)
 			// Continue with other rules even if one fails
-			r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
-			errs = append(errs, err)
+			r.recordNodeFailure(rule, node.Name, "EvaluationError", evalErr.Error())
+			errs = append(errs, evalErr)
 		}
 
 		// Persist the rule status
@@ -169,20 +170,18 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 
 			patch := client.MergeFrom(latestRule.DeepCopy())
 
-			// Upsert the node's evaluation only when evaluateRuleForNode produced
-			// one. On the failure path it returns before updateNodeEvaluationStatus
-			// runs, so writing the zero value here would either clobber a valid
-			// prior entry or append one with an empty NodeName, which the CRD
-			// rejects (MinLength=1) and would fail the whole status patch — taking
-			// the FailedNodes update below down with it.
-			var currEval *readinessv1alpha1.NodeEvaluation
-			for i := range rule.Status.NodeEvaluations {
-				if rule.Status.NodeEvaluations[i].NodeName == node.Name {
-					currEval = &rule.Status.NodeEvaluations[i]
-					break
+			// Upsert the node's evaluation only after a successful evaluation.
+			// On the failure path evaluateRuleForNode returns before recording a
+			// fresh NodeEvaluation, so this must leave any existing persisted
+			// evaluation untouched and only persist FailedNodes below.
+			if evalErr == nil {
+				var currEval *readinessv1alpha1.NodeEvaluation
+				for i := range rule.Status.NodeEvaluations {
+					if rule.Status.NodeEvaluations[i].NodeName == node.Name {
+						currEval = &rule.Status.NodeEvaluations[i]
+						break
+					}
 				}
-			}
-			if currEval != nil {
 				found := false
 				for i := range latestRule.Status.NodeEvaluations {
 					if latestRule.Status.NodeEvaluations[i].NodeName == node.Name {
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -169,28 +169,34 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 
 			patch := client.MergeFrom(latestRule.DeepCopy())
 
-			// update only this specific node evaluation status
-			currEval := readinessv1alpha1.NodeEvaluation{}
-			for _, eval := range rule.Status.NodeEvaluations {
-				if eval.NodeName == node.Name {
-					currEval = eval
+			// Upsert the node's evaluation only when evaluateRuleForNode produced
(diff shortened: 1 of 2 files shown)


sahitya-chandra:
/cc @AvineshTripathi 

could you please take a look at this PR whenever you get some free time? It has been open for quite a while now.

internal/controller/node_controller.go
```diff
@@ -169,28 +170,32 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
...
-			// update only this specific node evaluation status
-			currEval := readinessv1alpha1.NodeEvaluation{}
-			for _, eval := range rule.Status.NodeEvaluations {
-				if eval.NodeName == node.Name {
-					currEval = eval
-					break
+			// Upsert the node's evaluation only after a successful evaluation.
+			// On the failure path evaluateRuleForNode returns before recording a
+			// fresh NodeEvaluation, so this must leave any existing persisted
+			// evaluation untouched and only persist FailedNodes below.
```
ajaysundark, line 176: I view this more of a UX improvement than a bug fix --

1. Currently if `evaluateRuleForNode` fails for a node, it patches the rule status by overriding with NodeEvaluation{}, erasing previous (if existing) eval result. This change prevents from doing it, but the risk is possibly leaving a stale result for that rule.
2. We already capture the node failures at rule.status.failedNodes which should reflects recent evaluation failures for the node.

We have discussed moving the status object to a separate 'NodeReadinessEvaluation' CRD (in v1alpha2) to capture per rule results. This flow could look like --

```
processNodeAgainstAllRules {
   NodeReadinessEvaluationCR {
    // record result per rule
   }
}
```
instead of double looping and x-updates we are currently doing here.

Assuming we move to a 1--1 evaluation result present for each Node evaluated by NodeReadinessController, how would the rule-evaluation be structured? and how should we record per rule failures in it?

@Karthik-K-N  I think we should prioritize evaluating this to establish a cohesive observability for NRC ideally aligning it earlier with the mentorship timeline as there are some observability focus in it.

Karthik-K-N: If we have multiple rules per node and one among the rule evaluation fails we capture them in the ReadinessReport: https://github.com/kubernetes-sigs/node-readiness-controller/pull/133/changes#r2857725522

sahitya-chandra: Thanks @ajaysundark and @Karthik-K-N, I just went through #133, and the `NodeReadinessRuleReport` (nrrp) CRD makes sense to me. I agree a per-node report keyed by `ruleName` is cleaner for these results than the `NodeEvaluations[]` list inside the rule status, so I'm happy to align with that.

One small thing on the "UX or bug" point: I think it is a bit more than UX today. On the failure path, when the node has no prior cache `NodeEvaluation` (i.e. first evaluation), `currEval` is the zero value, so the patch carries `NodeEvaluation{}`. Its required fields are `omitempty`, so it serializes to `{}` and the apiserver rejects the whole `Status().Patch` with a 422. `failedNodes` is in that same patch, so it gets dropped too. So on that path the operator sees nothing, not even `failedNodes`, just a log line.

I checked this with envtest against the apiserver: on `main` the failed node ends with `failedNodes` empty, and with this change it persists. So I feel `failedNodes` only really captures the failure once we stop including the empty eval in the patch. They still go out in the same Status().Patch() call, but without the invalid {} entry the 422 no longer fires

On the stale result point, I agree. The idea here is just to keep the last good eval instead of erasing it, but I'm fine dropping the eval on failure instead if you prefer

Either way, I'm happy to keep this as a small v1alpha1 fix and leave the bigger change to the report work :)

ajaysundark:
/close

Superseded by #343

Thanks for the PR!

PR #222, opened 2026-05-06
fix: reconcile retry on rule processing errors

### LINKED ISSUE #220: NodeReconciler.Reconcile silently discards errors from processNodeAgainstAllRules, preventing workqueue requeue
## What happened

`NodeReconciler.Reconcile` calls `processNodeAgainstAllRules` but the function has no return value, so `Reconcile` always returns `ctrl.Result{}, nil` regardless of outcome:

https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/node_controller.go#L99-L102

Both `evaluateRuleForNode` errors (line 146) and `Status().Patch` errors (line 208) inside `processNodeAgainstAllRules` are logged but never surfaced. The controller-runtime workqueue treats every node reconciliation as successful and never requeues with backoff ; even when every rule evaluation failed.

This is inconsistent with `RuleReconciler`, which correctly returns errors from `processAllNodesForRule`.

When rule evaluations or status patches fail, `Reconcile` should return an error so the workqueue retries with exponential backoff instead of silently dropping the failure.

## How to reproduce

1. Create a `NodeReadinessRule` targeting a node.
2. Remove the `patch` verb on `nodes` from the controller's ClusterRole so `addTaintBySpec` fails persistently.
3. Trigger a node condition change.
4. Observe that `"Failed to evaluate rule for node"` is logged but no requeue occurs , the taint remains unapplied until the next unrelated node event.

## Suggested fix

Change `processNodeAgainstAllRules` to return an aggregate error and propagate it in `Reconcile`, similar to how `processAllNodesForRule` is handled in `RuleReconciler`.

/kind bug

dorodb-web22:
## Description

`processNodeAgainstAllRules` had no return value, so `Reconcile` always returned `ctrl.Result{}, nil` even when rule evaluation or status patch calls failed. This caused the workqueue to treat every reconciliation as successful, skipping backoff and requeue on transient errors. Node taint state could drift silently until an unrelated event triggered a new reconcile.

This change makes `processNodeAgainstAllRules` return an aggregated error via `errors.Join` and propagates it through `Reconcile`, consistent with how `RuleReconciler.processAllNodesForRule` already handles errors.

Adds a regression test that simulates a status patch failure using the interceptor pattern and verifies `Reconcile` returns a non-nil error.

## Related Issue

Fixes #220 

## Type of Change

/kind bug

## Testing

make test
make lint

## Checklist

- [x] make test passes
- [x] make lint passes

NONE

### DIFF
--- a/internal/controller/node_controller_test.go
+++ b/internal/controller/node_controller_test.go
@@ -940,7 +940,7 @@ var _ = Describe("Node Controller", func() {
 	})
 
 	Context("when rule status patch fails during node reconciliation", func() {
-		It("should return an error so the workqueue requeues", func() {
+		It("should return an error when rule status patch fails", func() {
 			ctx := context.Background()
 
 			testScheme := runtime.NewScheme()
--- a/internal/controller/node_controller_test.go
+++ b/internal/controller/node_controller_test.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"fmt"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/controller/node_controller_test.go
```diff
@@ -937,4 +938,73 @@ var _ = Describe("Node Controller", func() {
 				"Patch should not be called when taint removal is a no-op")
 		})
 	})
+
+	Context("when rule status patch fails during node reconciliation", func() {
+		It("should return an error so the workqueue requeues", func() {
```
Karthik-K-N: we are just checking if there is expected error and we are never checking workqueue length, its assumption that it requeue.
Do you think its better to rename test?

Karthik-K-N (commented):
Nice addition, Thank you

internal/controller/node_controller_test.go
dorodb-web22: Good point! i renamed it to "should return an error when rule status patch fails". since we're just checking the error returned rather than inspecting the workqueue directly, this is definitely more accurate.

ajaysundark:
Thank you for your contribution, @dorodb-web22!

/lgtm
/approve

PR #224, opened 2026-05-07
fix(controller): handle long rule names in bootstrap annotation keys

### LINKED ISSUE #223: [bug] Annotation key length limit exceeded for long NodeReadinessRule names
The `node-readiness-controller` uses a custom annotation on Node objects to track completion of bootstrap-only rules.

Currently the annotation key is generated like this:

```txt
readiness.k8s.io/bootstrap-completed-<rule-name>
```

However, Kubernetes annotation keys must follow the format:

```txt
[prefix/]name
```

The `name` part (after the last `/`) is limited to **63 characters**.

Since `bootstrap-completed-` already uses 20 characters, any `NodeReadinessRule` name longer than ~43 characters results in an invalid annotation key.

Because of this, the controller fails while patching the Node annotation and the bootstrap-only rule never gets marked as completed.

---

## Expected Behavior

The controller should support any valid `NodeReadinessRule` name without failing to patch Node annotations.

---

## Actual Behavior

The controller throws a validation error when trying to patch the Node annotation if the generated key exceeds Kubernetes annotation limits.

As a result:

- bootstrap-only rules never complete
- nodes can remain tainted indefinitely
- reconciliation loops may continue retrying

---

## Steps to Reproduce

1. Create a `NodeReadinessRule` with a long name (50+ characters)

Example:

```txt
my-very-important-network-readiness-rule-for-worker-nodes
```

2. Set:

```yaml
spec:
  enforcementMode: bootstrap-only
```

3. Wait for the rule conditions to become successful
4. Check controller logs

You will see annotation validation errors during the Node patch operation.

---

## Suggested Fix

The annotation key should be shortened when the rule name is too long.

Possible approaches:

- Truncation + hash
- Deterministic hash-based suffix
- Store bootstrap tracking somewhere other than Node annotations

Example idea:

```txt
readiness.k8s.io/bootstrap-<hash>
```

A hash-based approach would probably be the safest option to avoid collisions while staying within Kubernetes limits.

I would like to work on this issue

vishnukothakapu:
## Description
This PR fixes a bug where `NodeReadinessRule` resources with long names (longer than 43 characters) caused the controller to fail when patching Node annotations. Kubernetes strictly limits the name part of an annotation key to 63 characters. Since our key pattern was `readiness.k8s.io/bootstrap-completed-<rule-name>`, long rule names resulted in invalid annotation keys.

This PR implements a stable, UID-based key design:
- **Key format:** `readiness.k8s.io/bootstrap-completed-<ruleUID>`. Since `rule.GetUID()` is guaranteed by the API server to be ~36 characters (RFC 4122 UUID), this completely eliminates the 63-character limit issue without requiring custom hashing.
- **Value format:** `{"rule": "<name>"}`. The value is stored as a JSON payload containing the original rule name to preserve human readability for debugging.
- **Migration:** The controller seamlessly reads legacy `readiness.k8s.io/bootstrap-completed-<name>` keys and idempotently migrates them to the new UID-based format during node reconciliation.

This approach also cleanly addresses the stale annotation issue raised in #247: when a rule is deleted and recreated with the same name, it receives a fresh UID from the API server, ensuring it correctly tracks a new lifecycle state without being bypassed by a stale annotation key.

## Related Issue
Fixes #223, #247, #248

## Type of Change
/kind bug

## Testing
- Added `internal/controller/helper_unit_test.go`: Unit tests for `bootstrapAnnotationKey, bootstrapAnnotationValue`, and legacy migration helper functions.
- Updated `internal/controller/node_controller_reproduction_test.go`: Reproduction test that confirms the controller successfully uses UID-based annotations for rules with very long names, and explicitly tests legacy annotation migration idempotency.
- Verified with `go build ./...` and `go vet ./internal/controller/....`

## Checklist
- [X] `make test` passes
- [X] `make lint` passes

Yes. Existing `readiness.k8s.io/bootstrap-completed-<rule-name>` annotations will be automatically migrated to `readiness.k8s.io/bootstrap-completed-<ruleUID>` keys on the node, with a small JSON payload representing the rule name.

### DIFF
--- a/internal/controller/helper.go
+++ b/internal/controller/helper.go
@@ -17,8 +17,48 @@ limitations under the License.
 package controller
 
 import (
+	"encoding/json"
+
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
+
+const (
+	// bootstrapAnnotationPrefix is the common prefix for all bootstrap completion
+	// annotations on a Node. The suffix is the rule's metadata.uid (RFC 4122 UUID,
+	// ~36 chars), which is immutable for the object's lifetime and globally unique.
+	//
+	// Full key format: readiness.k8s.io/bootstrap-completed-<ruleUID>
+	// Value format:    {"rule-name":"<ruleName>"}   (for human readability)
+	bootstrapAnnotationPrefix = "readiness.k8s.io/bootstrap-completed-"
+)
+
+// bootstrapAnnotationKey returns the annotation key for a rule's bootstrap
+// completion state, using the rule's UID as the suffix.
+func bootstrapAnnotationKey(uid types.UID) string {
+	return bootstrapAnnotationPrefix + string(uid)
+}
+
+// bootstrapAnnotationValue returns the JSON-encoded value to store in the
+// bootstrap annotation. It includes the rule name for human readability.
+func bootstrapAnnotationValue(ruleName string) string {
+	v := struct {
+		RuleName string `json:"rule-name"`
+	}{RuleName: ruleName}
+	b, err := json.Marshal(v)
+	if err != nil {
+		return `{"rule-name":""}` // should never happen
+	}
+	return string(b)
+}
+
+
+
+// legacyBootstrapAnnotationKey returns the old-format annotation key used
+// before the UID migration: readiness.k8s.io/bootstrap-completed-<ruleName>.
+func legacyBootstrapAnnotationKey(ruleName string) string {
+	return bootstrapAnnotationPrefix + ruleName
+}
 
 // conditionsEqual checks if two condition slices are equal.
 func conditionsEqual(a, b []corev1.NodeCondition) bool {
--- a/internal/controller/helper_unit_test.go
+++ b/internal/controller/helper_unit_test.go
@@ -1,0 +1,56 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestBootstrapAnnotationKey(t *testing.T) {
+	g := NewWithT(t)
+
+	uid := types.UID("550e8400-e29b-41d4-a716-446655440000")
+	key := bootstrapAnnotationKey(uid)
+	g.Expect(key).To(Equal("readiness.k8s.io/bootstrap-completed-550e8400-e29b-41d4-a716-446655440000"))
+}
+
+func TestBootstrapAnnotationValue(t *testing.T) {
+	g := NewWithT(t)
+
+	t.Run("encodes rule name as JSON", func(t *testing.T) {
+		val := bootstrapAnnotationValue("my-rule")
+		g.Expect(val).To(Equal(`{"rule-name":"my-rule"}`))
+	})
+
+	t.Run("handles long rule names", func(t *testing.T) {
+		longName := "my-very-long-rule-name-that-exceeds-the-63-character-annotation-key-limit-strictly"
+		val := bootstrapAnnotationValue(longName)
+		g.Expect(val).To(ContainSubstring(longName))
+	})
+}
+
+func TestLegacyBootstrapAnnotationKey(t *testing.T) {
+	g := NewWithT(t)
+
+	key := legacyBootstrapAnnotationKey("my-rule")
+	g.Expect(key).To(Equal("readiness.k8s.io/bootstrap-completed-my-rule"))
+}
+
+
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -24,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
@@ -129,7 +130,7 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 		}
 
 		// Skip if bootstrap-only and already completed
-		if r.isBootstrapCompleted(ctx, node.Name, rule.Name) && rule.Spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
+		if r.isBootstrapCompleted(ctx, node.Name, rule.Name, rule.GetUID()) && rule.Spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
 			log.Info("Skipping bootstrap-only rule - already completed",
 				"node", node.Name, "rule", rule.Name)
 			continue
@@ -341,24 +342,20 @@ func (r *RuleReadinessController) removeTaintBySpec(ctx context.Context, node *c
 	})
 }
 
-// Bootstrap completion tracking.
-func (r *RuleReadinessController) isBootstrapCompleted(ctx context.Context, nodeName, ruleName string) bool {
-	// Check node annotation
+func (r *RuleReadinessController) isBootstrapCompleted(ctx context.Context, nodeName string, ruleName string, ruleUID types.UID) bool {
 	node := &corev1.Node{}
 	if err := r.Get(ctx, client.ObjectKey{Name: nodeName}, node); err != nil {
 		return false
 	}
-
-	annotationKey := fmt.Sprintf("readiness.k8s.io/bootstrap-completed-%s", ruleName)
-	_, exists := node.Annotations[annotationKey]
-	return exists
+	_, existsNew := node.Annotations[bootstrapAnnotationKey(ruleUID)]
+	_, existsLegacy := node.Annotations[legacyBootstrapAnnotationKey(ruleName)]
+	return existsNew || existsLegacy
 }
 
-func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, nodeName, ruleName string) {
+func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, nodeName, ruleName string, ruleUID types.UID) {
 	log := ctrl.LoggerFrom(ctx)
-
-	annotationKey := fmt.Sprintf("readiness.k8s.io/bootstrap-completed-%s", ruleName)
 	marked := false
+	annotationKey := bootstrapAnnotationKey(ruleUID)
 
 	// retry to handle conflict with concurrent node updates
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
@@ -367,21 +364,19 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
 			return err
 		}
 
-		// Check if already marked to avoid unnecessary updates
-		if node.Annotations != nil {
-			if _, exists := node.Annotations[annotationKey]; exists {
-				return nil
-			}
+		// Check if already marked to avoid unnecessary updates.
+		if _, exists := node.Annotations[annotationKey]; exists {
+			return nil
 		}
 
 		patch := client.MergeFrom(node.DeepCopy())
 
-		// Initialize annotations if nil
+		// Initialize annotations map if nil.
 		if node.Annotations == nil {
 			node.Annotations = make(map[string]string)
 		}
 
-		node.Annotations[annotationKey] = "true"
+		node.Annotations[annotationKey] = bootstrapAnnotationValue(ruleName)
 		if err := r.Patch(ctx, node, patch); err != nil {
 			return err
 		}
@@ -392,12 +387,12 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
 
 	switch {
 	case err != nil:
-		log.Error(err, "Failed to mark bootstrap completed", "node", nodeName, "rule", ruleName)
+		log.Error(err, "Failed to mark bootstrap completed", "node", nodeName, "rule", ruleName, "uid", ruleUID)
 	case marked:
-		log.Info("Marked bootstrap completed", "node", nodeName, "rule", ruleName)
+		log.Info("Marked bootstrap completed", "node", nodeName, "rule", ruleName, "uid", ruleUID)
 		metrics.BootstrapCompleted.WithLabelValues(ruleName).Inc()
 	default:
-		log.V(4).Info("Bootstrap already completed", "node", nodeName, "rule", ruleName)
+		log.V(4).Info("Bootstrap already completed", "node", nodeName, "rule", ruleName, "uid", ruleUID)
 	}
 }
 
--- a/internal/controller/nodereadinessrule_controller_test.go
+++ b/internal/controller/nodereadinessrule_controller_test.go
@@ -1147,9 +1147,10 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 		It("should handle bootstrap completion tracking", func() {
 			nodeName := "bootstrap-test-node"
 			ruleName := "bootstrap-test-rule"
+			ruleUID := types.UID("11111111-1111-1111-1111-111111111111")
 
 			// Initially not completed
-			completed := readinessController.isBootstrapCompleted(ctx, nodeName, ruleName)
+			completed := readinessController.isBootstrapCompleted(ctx, nodeName, ruleName, ruleUID)
 			Expect(completed).To(BeFalse())
 
 			// Create a node for testing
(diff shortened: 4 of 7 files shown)


internal/controller/helper.go
```diff
@@ -17,12 +17,32 @@ limitations under the License.
...
+
+// getBootstrapAnnotationKey generates a valid Kubernetes annotation key for a given rule name.
+// It handles rule names longer than 43 characters by using a hash to stay within the 63-character limit for the name part.
+func getBootstrapAnnotationKey(ruleName string) string {
+	namePart := ruleName
+	// Annotation name part (after prefix/) must be <= 63 characters.
+	// "bootstrap-completed-" is 20 characters.
+	if len(namePart) > 43 {
+		hash := md5.Sum([]byte(ruleName))
+		namePart = hex.EncodeToString(hash[:])
```
Karthik-K-N: Don't we need to restrict limit here?

internal/controller/helper.go
```diff
@@ -17,12 +17,32 @@ limitations under the License.
...
+)
+
+// getBootstrapAnnotationKey generates a valid Kubernetes annotation key for a given rule name.
+// It handles rule names longer than 43 characters by using a hash to stay within the 63-character limit for the name part.
+func getBootstrapAnnotationKey(ruleName string) string {
+	namePart := ruleName
+	// Annotation name part (after prefix/) must be <= 63 characters.
+	// "bootstrap-completed-" is 20 characters.
+	if len(namePart) > 43 {
+		hash := md5.Sum([]byte(ruleName))
```
Karthik-K-N: is md5 is recomended or is there any other better way? like sha256?

internal/controller/helper.go
vishnukothakapu: Since the hash is fixed to 32 characters, the annotation name length also stays fixed.
With a 20-character prefix + 32-character hash, the final annotation name becomes 52 characters total, no matter how long the original rule name is.

So this keeps it safely within Kubernetes’ 63-character limit.

internal/controller/helper.go
vishnukothakapu: I initially used MD5 because of its shorter fixed-length output, which fit easily within Kubernetes’ 63-character limit. I’ve now updated it to SHA256 and truncated it to 32 characters so it stays within the limit while providing stronger hashing.

ajaysundark:
Thanks for catching this. My only thoughts on this is that it takes away the human observability on this when a bootsrap-rule is done.  :/

internal/controller/helper.go
```diff
@@ -17,12 +17,34 @@ limitations under the License.
...
+	namePart := ruleName
+	// Annotation name part (after prefix/) must be <= 63 characters.
+	// "bootstrap-completed-" is 20 characters.
+	if len(namePart) > 43 {
+		hash := sha256.Sum256([]byte(ruleName))
+		// Use first 32 characters of SHA256 hash to stay within 63 char limit.
+		// 20 (prefix) + 32 (hash) = 52 characters.
+		namePart = hex.EncodeToString(hash[:16])
+	}
+	return fmt.Sprintf("%s%s", bootstrapAnnotationPrefix, namePart)
```
AvineshTripathi: should we go with a hybrid approach something like below 


```go

availableSpace := maxK8sAnnotationNameLen - len(bootstrapAnnotationPrefix)

const hashLen = 8
humanPartLen := availableSpace - hashLen - 1

hashBytes := sha256.Sum256([]byte(ruleName))
shortHash := hex.EncodeToString(hashBytes[:])[:hashLen]

namePart := fmt.Sprintf("%s-%s", ruleName[:humanPartLen], shortHash)


return fmt.Sprintf("%s%s", bootstrapAnnotationPrefix, namePart)
```

this way we keep the `ruleName`

ajaysundark:
The annotation restrictions are `<dns-prefix: 253>/<name: 63>`; `metadata.name` can also be 253 chars in length.

Hashing the rule name is one option, couple of alternatives to consider:
1. restricting the length of the rule-name to 63 (we can also move the "bootstrap-completed" to the value?)
2. alternatively, safer to use a standard key like "readiness.k8s.io/rule-status" and go with a json payload for value. Value doesnt seem to have a restriction. it'll allow us to capture the rule-name fully inside the payload.

We need to evaluate the pros/cons on the implementation. @vishnukothakapu Do you want to evaluate the alternatives and propose a plan here?

internal/controller/helper.go
ajaysundark: we have a lot of options to handle this. I added some different approach as well. We could discuss further over the meeting, feel free to join if you are available, @vishnukothakapu

vishnukothakapu:
> Thanks for catching this. My only thoughts on this is that it takes away the human observability on this when a bootsrap-rule is done. :/


Good point. I agree the full hash reduces readability during debugging. I’ll explore the hybrid approach with a readable prefix + short hash and compare it with the other alternatives discussed.

> The annotation restrictions are `<dns-prefix: 253>/<name: 63>`; `metadata.name` can also be 253 chars in length.
> 
> Hashing the rule name is one option, couple of alternatives to consider:
> 
> 1. restricting the length of the rule-name to 63 (we can also move the "bootstrap-completed" to the value?)
> 2. alternatively, safer to use a standard key like "readiness.k8s.io/rule-status" and go with a json payload for value. Value doesnt seem to have a restriction. it'll allow us to capture the rule-name fully inside the payload.
> 
> We need to evaluate the pros/cons on the implementation. @vishnukothakapu Do you want to evaluate the alternatives and propose a plan here?



Thanks @ajaysundark , these are good points. I’ll evaluate the tradeoffs between the current hashing approach, the hybrid readable-prefix approach, and the single annotation JSON payload design, then propose a direction based on readability, implementation complexity, and backward compatibility.

vishnukothakapu:
Hi @AvineshTripathi & @ajaysundark,
Thanks for the suggestion! I decided to go with the hybrid approach (truncated-name + short-hash) instead of the JSON payload approach because it preserves Kubernetes native selector support while still avoiding annotation length issues.
This keeps the annotations readable for debugging, ensures uniqueness with a deterministic hash, and stays compatible with the current architecture and existing workflows. I have updated the implementation and adjusted the related unit and integration tests accordingly.

ajaysundark:
> instead of the JSON payload approach because it preserves Kubernetes native selector support

Could you clarify your thoughts further on this? What are the downsides of using a json payload. This is also how kubectl saves last applied configurations in objects today - ref: https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#how-to-create-objects

@AvineshTripathi / @Karthik-K-N I think fixing this short-term with a hash based approach for length immunity doesnt feel right. A more reliable long term solution would be to maintain the rule-status inside a JSON payload to track individual rule evaluation data. It would also address concerns such as https://github.com/kubernetes-sigs/node-readiness-controller/issues/247

vishnukothakapu:
Hi @ajaysundark, thanks for the clarification, that makes sense. I can see how the structured JSON payload approach becomes more reliable long-term, especially with concerns like stale rule state and rule recreation handling from #247.

My initial preference for the hybrid approach was mainly to keep the fix minimal and avoid larger behavioral changes in this PR. But I agree the stable key + structured payload model feels more extensible and better suited for tracking rule lifecycle/state going forward.

vishnukothakapu:
> > instead of the JSON payload approach because it preserves Kubernetes native selector support
> 
> Could you clarify your thoughts further on this? What are the downsides of using a json payload. This is also how kubectl saves last applied configurations in objects today - ref: https://kubernetes.io/docs/tasks/manage-kubernetes-objects/declarative-config/#how-to-create-objects
> 
> @AvineshTripathi / @Karthik-K-N I think fixing this short-term with a hash based approach for length immunity doesnt feel right. A more reliable long term solution would be to maintain the rule-status inside a JSON payload to track individual rule evaluation data. It would also address concerns such as #247



@ajaysundark, the JSON payload approach is the right long-term design. It eliminates the key length issue at the root, preserves full observability (complete rule name in the value), and cleanly resolves #247, a deleted+recreated rule would no longer be blocked by a stale annotation. The main considerations are: 
1. handling read-modify-write races via RetryOnConflict (same pattern already used in addTaintBySpec), and 
2. documenting existing per-rule annotations as no-op leftovers in the changelog. Happy to update this PR to implement this if the direction is agreed.

ajaysundark:
We discussed this bug further in the meeting today.

A quick summary -

we agreed

```
readiness.k8s.io/bootstrap-completed-<ruleUID> = <>
```

or better, to overcome the 'opaque keys' argument:

```
readiness.k8s.io/bootstrap-completed-<ruleUID> = {
    "rule": "<name>", // for readability
}
```

maybe a preferred design for few reasons:

1. a single json-state with "read-modify-write" handled by multiple rule-workers may not work well in scale where parallel workers can result in cascading failures.
2. The value has 256K size / storage limitations. Though it isnt an immediate restriction for NRC as we dont expect # rules to be too many, this is a valid concern.

**Unique-key creation:**
1. Each NodeReadinessRule object already has a unique key assigned to it via `metadata.uid` by the k8s API server. So a new 'hashing' mechanism will be unnecessary. The api-machinery (RFC 4122 UUID) guarantees  ~36 chars, and immutable for the object's lifetime, and globally unique in the cluster). So `rule.GetUID()` is a stable, fixed-length suffix option here.
2. also addresses a case where a rule gets deleted and recreated with the same name, as each object creation gets a fresh UID in API server, which is the behavior we want (xref https://github.com/kubernetes-sigs/node-readiness-controller/issues/247)

**Migration path from "legacy" keys to UID based keys:**

~~1. On reconcile after the fix is implemented, the controller reads any legacy `readiness.k8s.io/bootstrap-completed-<name>` keys on the node, writes the equivalent UID-key record and deletes the legacy key. This is idempotent, once per node action.~~ Update: It is simpler to just controller handle either of the key format for a lazy migration.
3. Rules whose names were too long would never had succeeded writing the legacy key, as the patch would have always failed, so there is no migrate plan needed for them
4. New nodes simply will start setting the correct key under the new UID format.


cc @AvineshTripathi

ajaysundark:
@vishnukothakapu are you still available to take this fix?

vishnukothakapu:
Thanks @ajaysundark, I'm still available and happy to take this forward.

I agree with the UID-based annotation approach. Using `rule.GetUID()` avoids key length issues, eliminates the need for hashing, and correctly handles the delete/recreate scenario from #247.

I'll rework the implementation to use UID-based keys, add the legacy annotation migration logic, update the tests, and push an update to this PR soon.

vishnukothakapu:
Hi @ajaysundark, @AvineshTripathi, and @Karthik-K-N,

I've pushed the updated implementation based on the design discussed above.

- Moved to `readiness.k8s.io/bootstrap-completed-<ruleUID> keys with {"rule": "<name>"}` values.
- Implemented smooth legacy migration during node reconciliation.
- This natively resolves the stale annotation issue (#247) since recreated rules receive a fresh API server UID.

Tests are updated and passing. Ready for your review!

internal/controller/helper.go
```diff
@@ -17,9 +17,82 @@ limitations under the License.
...
+	if len(suffix) == 0 {
+		return false
+	}
+	// RFC 4122 UUIDs are exactly 36 chars with 4 hyphens (8-4-4-4-12).
+	// Legacy keys used the rule *name* which won't match this pattern.
+	return !isUUID(suffix)
+}
+
+// isUUID performs a lightweight check for RFC 4122 UUID format (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx).
+func isUUID(s string) bool {
```
ajaysundark: These functions are not used anywhere. can we remove them?

internal/controller/node_controller.go
```diff
@@ -337,24 +341,19 @@ func (r *RuleReadinessController) removeTaintBySpec(ctx context.Context, node *c
 	})
 }
 
-// Bootstrap completion tracking.
-func (r *RuleReadinessController) isBootstrapCompleted(ctx context.Context, nodeName, ruleName string) bool {
-	// Check node annotation
+func (r *RuleReadinessController) isBootstrapCompleted(ctx context.Context, nodeName string, ruleUID types.UID) bool {
```
ajaysundark: Can we just update  isBootstrapCompleted  to check for both the new UID-based key and the legacy key, and return true if either of them exists?

ajaysundark (changes requested):
Thanks for your PR. I think it may be simpler if we just check both keys in `isBootstrapCompleted`, and dont have to migrate/delete the old keys. Can you please iterate on this change?

internal/controller/node_controller.go
vishnukothakapu: Done! I've updated `isBootstrapCompleted` to check for both the UID-based key and the legacy key, and removed the migration code entirely since it's no longer needed.

ajaysundark (commented):
/lgtm

Thanks for the updates. One minor piece that Id like added is unit tests for us bootstrap completed for different annotation permutations (old vs new nodes). It's ready to merge otherwise.

vishnukothakapu:
> /lgtm
> 
> Thanks for the updates. One minor piece that Id like added is unit tests for us bootstrap completed for different annotation permutations (old vs new nodes). It's ready to merge otherwise.

Hi @ajaysundark , I have added the unit tests for `isBootstrapCompleted` to cover the different annotation permutations (old vs. new nodes, and both). It should be good to go now!

ajaysundark (approved):
Thanks for your contributions!

internal/controller/helper.go
```diff
@@ -17,9 +17,49 @@ limitations under the License.
...
+	"k8s.io/apimachinery/pkg/types"
+)
+
+const (
+	// bootstrapAnnotationPrefix is the common prefix for all bootstrap completion
+	// annotations on a Node. The suffix is the rule's metadata.uid (RFC 4122 UUID,
+	// ~36 chars), which is immutable for the object's lifetime and globally unique.
+	//
+	// Full key format: readiness.k8s.io/bootstrap-completed-<ruleUID>
+	// Value format:    {"rule":"<ruleName>"}   (for human readability)
```
ajaysundark:
```suggestion
	// Value format:    {"rule-name":"<ruleName>"}   (for human readability)
```

internal/controller/helper.go
```diff
@@ -17,9 +17,49 @@ limitations under the License.
...
+
+// bootstrapAnnotationValue returns the JSON-encoded value to store in the
+// bootstrap annotation. It includes the rule name for human readability.
+func bootstrapAnnotationValue(ruleName string) string {
+	v := struct {
+		Rule string `json:"rule"`
+	}{Rule: ruleName}
+	b, err := json.Marshal(v)
+	if err != nil {
+		return `{"rule":""}` // should never happen
```
ajaysundark:
```suggestion
		return `{"rule-name":""}` // should never happen
```

internal/controller/helper_unit_test.go
```diff
@@ -0,0 +1,56 @@
...
+	key := bootstrapAnnotationKey(uid)
+	g.Expect(key).To(Equal("readiness.k8s.io/bootstrap-completed-550e8400-e29b-41d4-a716-446655440000"))
+}
+
+func TestBootstrapAnnotationValue(t *testing.T) {
+	g := NewWithT(t)
+
+	t.Run("encodes rule name as JSON", func(t *testing.T) {
+		val := bootstrapAnnotationValue("my-rule")
+		g.Expect(val).To(Equal(`{"rule":"my-rule"}`))
```
ajaysundark:
```suggestion
		g.Expect(val).To(Equal(`{"rule-name":"my-rule"}`))
```

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -1162,42 +1163,44 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+			ruleUID := types.UID("22222222-2222-2222-2222-222222222222")
 
-			// Create a node with the bootstrap annotation already set
+			// Create a node with the UID-based bootstrap annotation already set
 			node := &corev1.Node{
 				ObjectMeta: metav1.ObjectMeta{
 					Name: nodeName,
 					Annotations: map[string]string{
-						"readiness.k8s.io/bootstrap-completed-" + ruleName: "true",
+						bootstrapAnnotationKey(ruleUID): `{"rule":"bootstrap-ctx-test-rule"}`,
```
ajaysundark:
```suggestion
						bootstrapAnnotationKey(ruleUID): `{"rule-name":"bootstrap-ctx-test-rule"}`,
```

internal/controller/helper.go
```diff
@@ -17,9 +17,49 @@ limitations under the License.
...
+func bootstrapAnnotationKey(uid types.UID) string {
+	return bootstrapAnnotationPrefix + string(uid)
+}
+
+// bootstrapAnnotationValue returns the JSON-encoded value to store in the
+// bootstrap annotation. It includes the rule name for human readability.
+func bootstrapAnnotationValue(ruleName string) string {
+	v := struct {
+		Rule string `json:"rule"`
+	}{Rule: ruleName}
```
ajaysundark:
```suggestion
		RuleName string `json:"rule-name"`
	}{RuleName: ruleName}
```

ajaysundark (commented):
minor suggestion for the field name.

internal/controller/helper.go
```diff
@@ -17,9 +17,49 @@ limitations under the License.
...
+)
+
+const (
+	// bootstrapAnnotationPrefix is the common prefix for all bootstrap completion
+	// annotations on a Node. The suffix is the rule's metadata.uid (RFC 4122 UUID,
+	// ~36 chars), which is immutable for the object's lifetime and globally unique.
+	//
+	// Full key format: readiness.k8s.io/bootstrap-completed-<ruleUID>
+	// Value format:    {"rule":"<ruleName>"}   (for human readability)
+	bootstrapAnnotationPrefix = "readiness.k8s.io/bootstrap-completed-"
```
AvineshTripathi, line 33: Not a blocker: Should we make this a generic rule annotation instead of being bootstrap specific? This would let us reuse it for other properties later, like showing a rule's dryRun status.

cc @ajaysundark

vishnukothakapu: Good point! Not opposed to making it generic. Wanted to flag one thing, if reconciles ever run concurrently, a single annotation could cause read-modify-write races. If we go generic, we'd need to be careful there. What do you think?

ajaysundark: Agree to not increase the scope for this change. We can handle this as a bug fix for now.

I think scoped annotation with readiness.k8s.io prefix will scale better than global/single metadata.. let us discuss the other use cases individually.

ajaysundark (approved):
/lgtm

PR #226, opened 2026-05-07
nrr: Refactor `r.List()` to call only once.

sujalshah-bit:
## Description
Previously, several child functions were calling
"r.List()" duplicating work. This commit address
the issue. Now we call it "r.List()" once
in "reconcile()" parent functions.

## Related Issue
None

## Type of Change

/kind cleanup

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
NONE
```
Doc #(issue)

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -120,16 +120,21 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{RequeueAfter: time.Second}, nil
 	}
 
+	nodeList := &corev1.NodeList{}
+	if err := r.List(ctx, nodeList); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	// Handle deletion reconciliation loop.
 	if !rule.DeletionTimestamp.IsZero() {
-		return r.reconcileDelete(ctx, rule)
+		return r.reconcileDelete(ctx, rule, nodeList)
 	}
 
 	// Detect nodeSelector changes and cleanup old nodes
 	cachedRule := r.Controller.getCachedRule(rule.Name)
 	if cachedRule != nil && nodeSelectorChanged(rule.Spec.NodeSelector, cachedRule.Spec.NodeSelector) {
 		log.Info("NodeSelector changed, cleaning up nodes from old selector", "rule", rule.Name)
-		if err := r.Controller.cleanupNodesAfterSelectorChange(ctx, cachedRule, rule); err != nil {
+		if err := r.Controller.cleanupNodesAfterSelectorChange(ctx, cachedRule, rule, nodeList); err != nil {
 			log.Error(err, "Failed to cleanup nodes after selector change", "rule", rule.Name)
 			return ctrl.Result{RequeueAfter: time.Minute}, err
 		}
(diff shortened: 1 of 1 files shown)


internal/controller/nodereadinessrule_controller.go
```diff
@@ -140,19 +145,13 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 
 	// Handle dry run
 	if rule.Spec.DryRun {
-		if err := r.Controller.processDryRun(ctx, rule); err != nil {
```
ajaysundark, line 143: Thanks for your PR.

I wonder whether a void return is the right long-term API. if future changes add error-returning logic to these functions, the caller won't see it. Can we keep the signature as-is and make changes targeted only to the List related changes?

internal/controller/nodereadinessrule_controller.go
```diff
@@ -140,19 +145,13 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
-			log.Error(err, "Failed to process dry run", "rule", rule.Name)
-			return ctrl.Result{RequeueAfter: time.Minute}, err
-		}
+		r.Controller.processDryRun(ctx, rule, nodeList)
 	} else {
 		// Clear previous dry run results
 		rule.Status.DryRunResults = readinessv1alpha1.DryRunResults{}
 
 		// Process all applicable nodes for this rule
-		if err := r.Controller.processAllNodesForRule(ctx, rule); err != nil {
```
ajaysundark, line 152: SAA

ajaysundark (changes requested):
mostly looks good. will need followup to retain error handling

internal/controller/nodereadinessrule_controller.go
AvineshTripathi: +1 here, returning an error keeps the flexibility of retrying in case of errors

internal/controller/nodereadinessrule_controller.go
```diff
@@ -516,12 +513,9 @@ func (r *RuleReadinessController) updateRuleStatus(ctx context.Context, rule *re
...
 
 // processDryRun processes dry run for a rule.
-func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule) error {
-	nodeList := &corev1.NodeList{}
-	if err := r.List(ctx, nodeList); err != nil {
-		return err
-	}
-
+//
+//nolint:unparam // Keep error return for future extensibility and API stability.
```
sujalshah-bit, line 517: Linter is complaining about:
```bash
internal/controller/nodereadinessrule_controller.go:510:140: (*RuleReadinessController).processDryRun - result 0 (error) is always nil (unparam)
func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule, nodeList *corev1.NodeList) error {
```
Therefore, I have now added this line.

internal/controller/nodereadinessrule_controller.go
sujalshah-bit: Initially, that was not my plan ( To remove err signatures), but the linter was complaining about [this](https://github.com/kubernetes-sigs/node-readiness-controller/pull/226/changes#r3227645830).

sujalshah-bit (commented):
@ajaysundark and @AvineshTripathi, I have addressed your feedback. Kindly review :)

Also, I was planning to benchmark  `internal/controller/nodereadinessrule_controller.go`.  So that we can optimize for memory and CPU, WDYT?

ajaysundark:
> Also, I was planning to benchmark internal/controller/nodereadinessrule_controller.go. So that we can optimize for memory and CPU, WDYT?

Out of scope for this PR. but you could create a separate issue with your plan / suggestions and follow up on this if you are interested.

PR #229, opened 2026-05-08
fix: return node evaluation errors to trigger reconcile retries

### LINKED ISSUE #220: NodeReconciler.Reconcile silently discards errors from processNodeAgainstAllRules, preventing workqueue requeue

Nesar976:
This change ensures that node evaluation failures are propagated back to the reconcile loop instead of being silently ignored.

Previously, transient failures during rule evaluation or taint patching were only logged internally, preventing controller-runtime from requeueing the reconcile request.

This PR:

* aggregates evaluation/status update errors
* returns reconcile errors properly for retry/backoff handling
* preserves existing behavior by continuing evaluation across rules
* adds test coverage for aggregated reconcile failures

Related Issue

Fixes #220

Type of Change

/kind bug

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
@@ -97,18 +98,22 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	}
 
 	// Process node against all applicable rules
-	r.Controller.processNodeAgainstAllRules(ctx, node)
+	if err := r.Controller.processNodeAgainstAllRules(ctx, node); err != nil {
+		return ctrl.Result{}, err
+	}
 
 	return ctrl.Result{}, nil
 }
 
 // processNodeAgainstAllRules processes a single node against all applicable rules.
-func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context, node *corev1.Node) {
+func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context, node *corev1.Node) error {
 	log := ctrl.LoggerFrom(ctx)
 
 	// Get all known (cached) applicable rules for this node
 	applicableRules := r.getApplicableRulesForNode(ctx, node)
 	log.Info("Processing node against rules", "node", node.Name, "ruleCount", len(applicableRules))
+
+	var errs []error
 
 	for _, rule := range applicableRules {
 		log.V(4).Info("Processing rule from cache",
@@ -148,6 +153,7 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 				"node", node.Name, "rule", rule.Name)
 			// Continue with other rules even if one fails
 			r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
+			errs = append(errs, fmt.Errorf("rule %s: %w", rule.Name, err))
 		}
 
 		// Persist the rule status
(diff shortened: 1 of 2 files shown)


internal/controller/node_controller.go
```diff
@@ -97,19 +98,23 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 // processNodeAgainstAllRules processes a single node against all applicable rules.
-func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context, node *corev1.Node) {
+func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context, node *corev1.Node) error {
 	log := ctrl.LoggerFrom(ctx)
 
 	// Get all known (cached) applicable rules for this node
 	applicableRules := r.getApplicableRulesForNode(ctx, node)
 	log.Info("Processing node against rules", "node", node.Name, "ruleCount", len(applicableRules))
 
+	var errs []error
```
sujalshah-bit:
```suggestion
	var retryableErrs []error
```

internal/controller/node_controller.go
```diff
@@ -211,13 +217,16 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 				"rule", rule.Name,
 				"resourceVersion", rule.ResourceVersion)
 			// continue with other rules
+			errs = append(errs, fmt.Errorf("rule %s status update: %w", rule.Name, err))
```
sujalshah-bit:
```suggestion
if isRetryable(err) {
        retryableErrs = append(retryableErrs, err)
    }
```

sujalshah-bit (commented):
Thanks for the PR @Nesar976 . 

I think returning aggregated errors here improves convergence semantics because previously transient failures (patch conflicts, API timeouts, status update failures, etc.) could get logged and then effectively lost without another reconcile trigger.

One concern though is that we're currently aggregating *all* errors:

```go
errs = append(errs, err)
...
return errors.Join(errs...)
```

Since `Reconcile()` returns that error directly, controller-runtime will requeue the node reconcile for any failure, including potentially non-retryable/business-logic errors.

For example:

* transient API/conflict errors → retry makes sense
* invalid rule configuration / unsupported evaluation state → retry likely won't help

With the current approach, a permanently failing rule could keep causing unnecessary reconcile retries for the node.

Would it make sense to distinguish retryable/system errors from expected rule evaluation failures and only return the retryable ones while still recording all failures in status/events/metrics?

Also, sign CLA :)

ajaysundark:
looks like this PR is duplicate to https://github.com/kubernetes-sigs/node-readiness-controller/pull/222/

PR #164, opened 2026-03-12
Make NodeReadinessRule spec fields immutable to prevent security issues

Priyankasaggu11929:
Currently, `NodeReadinessRule.spec` fields (especially `spec.taint`) are mutable after creation.
And so, when these fields are modified in-place, previously managed taints becomes orphaned on nodes selected by the NRR rule.

(please see `Example 1` in the testing section to see below case in action)

For example, let's say I chnaged the taint effect from `NoSchedule` to `NoExecute`, this will make the NRC controller add a new taint on the node, but without removing the original one ([clean up today only happens during NRR rule deletion](https://github.com/kubernetes-sigs/node-readiness-controller/blob/c2069218a220e482d5bff1d8291dd961282855cb/internal/controller/nodereadinessrule_controller.go#L173-L201), not in place updates).

So, if a malicious or misconfigured update then reverts the rule back to `NoSchedule`, it will leave the `NoExecute` taint on the node but the taint will no longer be managed by the NRC controller.

This will lead to significant workload disruption (unexpected pod evictions, permanent scheduling failures etc) until these taints are manually removed.

And if we don't make these fields immutable, then that will require NRC controller to track and reconcile previously managed taints  (in case of in-place updates), which significantly increase the controller complexity and could introduce potential race conditions.

### Proposed Fix

- The PR is making the following `NodeReadinessRule.spec` fields immutable using CRD validation (`x-kubernetes-validations`):
    - `spec.taint.key`
    - `spec.taint.value`
    - `spec.taint.effect`
    - `spec.nodeSelector`
    - `spec.conditions`
    - `spec.enforcementMode` (added in 5e3ac02b0759b511173e5f5a215a9ac82a2fd214)
   
    These fields collectively define the identity and scope of the taint managed by the NRC controller. 
    So, preventing these fields from being modified after creation will ensure that the NRC controller always maintains consistent ownership of the taints it manages.
    
    And If a user indeed needs to change these values, they must delete and recreate the `NodeReadinessRule`.

- Additional, this PR also fixes an issue with  [`hasTaintBySpec`](https://github.com/kubernetes-sigs/node-readiness-controller/blob/c2069218a220e482d5bff1d8291dd961282855cb/internal/controller/node_controller.go#L234-L242) function (see the example 2 in the testing section) where taint comparison currently ignores the `value` field, and thus, causing updates that only changed the taint `value` to be missed during NRC controller reconciliation.

[...]

PR #231, opened 2026-05-09
Fix: remove dead nil selector check

### LINKED ISSUE #230: cleanup: Remove dead code `nil` selector check in `cleanupNodesAfterSelectorChange`
### What happened:

In `RuleReadinessController.cleanupNodesAfterSelectorChange`, there is a conditional check for `oldSelector == nil`. This code is unreachable.

```go
// Build old selector
oldSelector, err := metav1.LabelSelectorAsSelector(&oldRule.Spec.NodeSelector)
if err != nil {
    return fmt.Errorf("failed to parse old node selector: %w", err)
}

// ... later in the loop ...
if oldSelector == nil {
    // nil selector matches all nodes
    matchedOld = true
} else {
    matchedOld = oldSelector.Matches(labels.Set(node.Labels))
}
```

Because `oldRule.Spec.NodeSelector` is a struct (not a pointer), its address is never `nil`. If `metav1.LabelSelectorAsSelector` parses an empty struct successfully (`err == nil`), it returns a `labels.Everything()` selector. It never returns `nil, nil`.

Therefore, `oldSelector` is guaranteed to be non-nil if no error was returned, making the `if oldSelector == nil` check dead code.

---

### What you expected to happen:

The dead code should be removed. Since `labels.Everything().Matches(...)` correctly returns `true` for all nodes, the fallback logic is unnecessary.

We can simplify the loop body to just:

```go
matchedOld := oldSelector.Matches(labels.Set(node.Labels))
```

---

### How to reproduce it (as minimally and precisely as possible):

N/A - Code inspection in `internal/controller/nodereadinessrule_controller.go`.

/kind cleanup

dorodb-web22:
## Description

`metav1.LabelSelectorAsSelector` is called with `&oldRule.Spec.NodeSelector`, where `NodeSelector` is a struct value, not a pointer. It never returns `nil, nil` ; an empty struct yields `labels.Everything()`, which correctly matches all nodes. The subsequent `if oldSelector == nil` branch was therefore dead code and could never be reached.

Remove the nil check and call `oldSelector.Matches(...)` directly.

## Related Issue

Fixes #230

## Type of Change

/kind cleanup

## Testing

make test
make lint

## Checklist

- [x] make test passes
- [x] make lint passes

NONE

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -645,13 +645,7 @@ func (r *RuleReadinessController) cleanupNodesAfterSelectorChange(ctx context.Co
 	var errors []string
 	for _, node := range nodeList.Items {
 		// Check if node matched old selector
-		matchedOld := false
-		if oldSelector == nil {
-			// nil selector matches all nodes
-			matchedOld = true
-		} else {
-			matchedOld = oldSelector.Matches(labels.Set(node.Labels))
-		}
+		matchedOld := oldSelector.Matches(labels.Set(node.Labels))
 
 		// Check if node matches new selector (use newRule for current evaluation)
 		matchesNew := r.ruleAppliesTo(ctx, newRule, &node)


internal/controller/node_controller.go
```diff
@@ -97,17 +98,20 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	}
 
 	// Process node against all applicable rules
-	r.Controller.processNodeAgainstAllRules(ctx, node)
+	if err := r.Controller.processNodeAgainstAllRules(ctx, node); err != nil {
```
ajaysundark: Thanks for your PR. Please note this is also addressed here: https://github.com/kubernetes-sigs/node-readiness-controller/pull/229. Can you coordinate with the other author?

ajaysundark: ah I see your original PR: https://github.com/kubernetes-sigs/node-readiness-controller/pull/222. This possibly is carrying your earlier changes as well. Can you scope the changes in this PR to only nil check?

dorodb-web22: @ajaysundark Good catch, I had accidentally branched from another PR instead of `main`. i have now force-pushed a cleaned-up version scoped only to the nil check in `nodereadinessrule_controller.go`. Thanks....

ajaysundark:
I commented at the issue - modifying the nodeselector is no longer valid. So this code path should not hit any longer. Could you confirm this at the code flow and create a new issue for this if needed?

If a larger tech-debt is needed, I'd consider fixing this is duplicate / low-value.

dorodb-web22:
> I commented at the issue - modifying the nodeselector is no longer valid. So this code path should not hit any longer. Could you confirm this at the code flow and create a new issue for this if needed?
> 
> If a larger tech-debt is needed, I'd consider fixing this is duplicate / low-value.

@ajaysundark  i Confirmed  that since `nodeSelector` was made immutable in #164, the `cleanupNodesAfterSelectorChange` path is no longer reachable during normal reconciliation, which makes this nil-check fix unnecessary .

i will close this PR and instead open a follow-up issue to remove `cleanupNodesAfterSelectorChange` entirely as dead code cleanup.

PR #237, opened 2026-05-10
 fix:return evaluation errors from processAllNodesForRule

### LINKED ISSUE #234: bug: processAllNodesForRule swallows node evaluation errors, preventing reconcile retries
## Describe the bug

`RuleReadinessController.processAllNodesForRule` catches errors from `evaluateRuleForNode` but always returns `nil`, even when evaluations fail:

```go
    if r.ruleAppliesTo(ctx, rule, &node) {
        if err := r.evaluateRuleForNode(ctx, rule, &node); err != nil {
            log.Error(err, "Failed to evaluate node for rule", ...)
            r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
            metrics.Failures.WithLabelValues(rule.Name, "EvaluationError").Inc()
            // error is dropped here
        }
    }
}
return nil // always nil
```

Because `processAllNodesForRule` always returns `nil`, the calling `RuleReconciler.Reconcile` at line 152 never sees the failure, so controller-runtime treats every reconciliation as successful and does not trigger a requeue/backoff.

This is the same class of bug fixed in #222 for the node reconciler path. This function is the rule reconciler path equivalent.

## Expected behavior

Transient failures during node evaluation (e.g. API conflicts, patch errors) should cause `processAllNodesForRule` to return an aggregated error so that controller-runtime can requeue with backoff.

## Suggested fix

Accumulate errors and return `errors.Join(errs...)`, same pattern as `processNodeAgainstAllRules` after #222. Existing behavior of continuing across all nodes is preserved.

/kind bug

dorodb-web22:
## Description

`processAllNodesForRule` was catching errors from `evaluateRuleForNode` but always returning `nil`. This prevented `RuleReconciler.Reconcile` from seeing any failures, so controller-runtime never triggered a requeue on transient errors like API conflicts or patch failures.

This fix accumulates errors across the node loop and returns them via `errors.Join`, using the same pattern applied to `processNodeAgainstAllRules` in #222. The existing behavior of continuing evaluation across all nodes is preserved.

## Related Issue

Fixes #234

## Type of Change

/kind bug

## Testing

make test
make lint

## Checklist

- [x] `make test` passes
- [x] `make lint` passes

```release-note
NONE

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
@@ -265,6 +266,7 @@ func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, ru
 	log.Info("Processing all nodes for rule", "rule", rule.Name, "totalNodes", len(nodeList.Items))
 
 	var appliedNodes []string
+	var errs []error
 	for _, node := range nodeList.Items {
 		if r.ruleAppliesTo(ctx, rule, &node) {
 			appliedNodes = append(appliedNodes, node.Name)
@@ -274,6 +276,7 @@ func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, ru
 				log.Error(err, "Failed to evaluate node for rule", "rule", rule.Name, "node", node.Name)
 				r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
 				metrics.Failures.WithLabelValues(rule.Name, "EvaluationError").Inc()
+				errs = append(errs, err)
 			}
 		}
 	}
@@ -287,7 +290,7 @@ func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, ru
 	}
 
 	log.Info("Completed processing nodes for rule", "rule", rule.Name, "processedCount", len(appliedNodes))
-	return nil
+	return errors.Join(errs...)
 }
 
 // evaluateRuleForNode evaluates a single rule against a single node.
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -145,6 +145,7 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	r.Controller.updateRuleCache(ctx, rule)
 
 	// Handle dry run
+	var processErr error
 	if rule.Spec.DryRun {
 		if err := r.Controller.processDryRun(ctx, rule, nodeList); err != nil {
 			log.Error(err, "Failed to process dry run", "rule", rule.Name)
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: internal/controller/nodereadinessrule_controller.go [hunk 1]
```diff
@@ -290,7 +293,7 @@ func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, ru
 	}
 
 	log.Info("Completed processing nodes for rule", "rule", rule.Name, "processedCount", len(appliedNodes))
-	return nil
+	return errors.Join(errs...)
```
AvineshTripathi, line 298: returning an error will keep the caller function(in this case [this](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/nodereadinessrule_controller.go#L152)) in an infinite loop. Other rules will not be evaluated too. It will also not update the status([here](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/nodereadinessrule_controller.go#L152)) till the error is fixed.

AvineshTripathi (commented):
cc @ajaysundark wdyt about this

internal/controller/nodereadinessrule_controller.go
dorodb-web22: the early return was skipping `updateRuleStatus` and `cleanupDeletedNodes`. i ffixed by saving the error, letting status update complete, then returning the error at the end to trigger requeue.

internal/controller/nodereadinessrule_controller.go
ajaysundark, line 279: Can you add a test case for this where `evaluateRuleForNode` returns an error?

internal/controller/nodereadinessrule_controller.go
```diff
@@ -287,7 +295,7 @@ func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, ru
...
```
ajaysundark, line 298: > requeue on transient errors like API conflicts or patch failures.

can you also add unit tests to capture these gaps that this PR aims to address?

internal/controller/nodereadinessrule_controller.go
ajaysundark, line 267: Can we remove this comment as it is no longer true?

ajaysundark:
@dorodb-web22 Thanks for your PR, Could you rebase and address the test failures?

AvineshTripathi:
/lgtm 

Please rebase

dorodb-web22:
Branch has been rebased onto the latest `upstream/main` . The conflict markers in `internal/controller/nodereadinessrule_controller.go` have been resolved , the branch should be conflict-free now. The `needs-rebase` label appears to be stale; waiting for Prow to re-sync.

Also addressed the lint failures from the previous CI run:
- Removed the unused `//nolint:unparam` directive on `processAllNodesForRule`
- Made `processAllNodesForRule` actually propagate errors via `errors.Join` (fixing the underlying issue the nolint was masking)
- Renamed `var errors []string` → `var errMsgs []string` in `cleanupTaintsForRule` and `cleanupNodesAfterSelectorChange` to resolve the `revive` import-shadowing warnings

PR #248, opened 2026-05-13
fix: remove bootstrap annotations from nodes when a bootstrap-only rule is deleted

### LINKED ISSUE #247: Bootstrap-only rule deletion leaves stale completion annotations, causing evaluation bypass on rule recreation
## What happened

When a NodeReadinessRule using enforcementMode: bootstrap-only is deleted, the controller removes the taints managed by the rule, but the bootstrap completion annotations written on nodes are left behind.

The annotation key is generated using the rule name:
```
go annotationKey := fmt.Sprintf("readiness.k8s.io/bootstrap-completed-%s", ruleName) 
```
isBootstrapCompleted() later checks this annotation to decide whether node evaluation should be skipped.

Because the annotation only depends on the rule name, recreating a rule with the same name causes previously-bootstrapped nodes to be treated as already completed, even if the new rule has different conditions.

As a result, node evaluation is skipped and the new taint may never be applied.

I traced this to the delete path in reconcileDelete(). It currently cleans up taints through cleanupTaintsForRule() but does not clean up bootstrap annotations before removing the finalizer.

Deleting a bootstrap-only rule should also remove the bootstrap completion state associated with that rule.

If a rule is recreated with the same name, nodes should be evaluated from scratch.

## How to reproduce
```
# Create bootstrap-only rule
kubectl apply -f - <<EOF
kind: NodeReadinessRule
metadata:
  name: network-readiness
spec:
  enforcementMode: bootstrap-only
  conditions:
  - type: NetworkUnavailable
    requiredStatus: "False"
  taint:
    key: readiness.k8s.io/network-not-ready
    effect: NoSchedule
  nodeSelector:
    matchLabels:
      kubernetes.io/os: linux
EOF

# Wait for bootstrap completion

# Delete the rule
kubectl delete nodereadinessrule network-readiness

# Verify annotations still exist
kubectl get nodes -o json | \
jq '.items[] | select(
  .metadata.annotations["readiness.k8s.io/bootstrap-completed-network-readiness"] != null
) | .metadata.name'

# Recreate rule with same name but different condition
kubectl apply -f - <<EOF
kind: NodeReadinessRule
metadata:
  name: network-readiness
spec:
  enforcementMode: bootstrap-only
  conditions:
  - type: DiskPressure
    requiredStatus: "False"
  taint:
    key: readiness.k8s.io/network-not-ready
    effect: NoSchedule
  nodeSelector:
    matchLabels:
      kubernetes.io/os: linux
EOF
```
At this point, nodes that previously completed bootstrap can skip evaluation because the old annotation is still present.

## Suggested fix

A few possible directions:

- Remove bootstrap annotations during finalizer cleanup
- Use rule UID instead of rule name in the annotation key
- Combine UID-based keys with best-effort cleanup during deletion

Happy to work on a fix if this approach makes sense.

/kind bug

Karman580:
## Description

`reconcileDelete` cleaned up managed taints but left bootstrap-completion
annotations (`readiness.k8s.io/bootstrap-completed-<rule-name>`) on node
objects indefinitely. Because the annotation key encodes only the rule
name, recreating a rule with the same name caused previously-bootstrapped
nodes to be skipped by `isBootstrapCompleted`, potentially leaving them
without the taint the new rule requires and allowing workloads to schedule
on nodes that should be gated.

This PR adds `cleanupBootstrapAnnotationsForRule`, called from
`reconcileDelete` after `cleanupTaintsForRule` for `bootstrap-only` rules.
It iterates all nodes matching the rule's selector and removes the
annotation via a `RetryOnConflict` `MergeFrom` patch, consistent with the
existing taint cleanup approach. The finalizer is not removed if any
annotation patch fails, so deletion is safely retried on the next cycle.

## Related Issue

Fixes #247

## Type of Change

/kind bug

## Testing

- Added integration test `should remove bootstrap annotations from nodes when a bootstrap-only rule is deleted` in `nodereadinessrule_controller_test.go`
- Test verifies: annotation is removed after rule deletion, unrelated annotations are preserved
- `make test` passes — all existing tests continue to pass

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fix: bootstrap-only rule deletion now removes completion annotations from
nodes so that recreating a rule with the same name evaluates all nodes
from scratch rather than skipping previously-bootstrapped nodes.
```

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


ajaysundark:
I added a comment here https://github.com/kubernetes-sigs/node-readiness-controller/issues/247#issuecomment-4463837480 - annotations are controller's metadata on the node, instead of cleaning up across the fleet for rule update, we could make the annotation reflect right state of the rule to decide whether this need another reconciliation.

ajaysundark:
Instead of solving piecewise, we could plan for a better design for the annotation to address this.

ajaysundark:
Thanks for raising this PR! This is no longer needed as https://github.com/kubernetes-sigs/node-readiness-controller/pull/224 covers this use-case.

PR #252, opened 2026-05-14
fix: webhook fails closed when rule listing errors

### LINKED ISSUE #251: bug: webhook fails open on rule listing error
## Summary

The `validateTaintConflicts` function in the admission webhook silently allows rule creation to proceed when the API server fails to list existing rules. This means two `NodeReadinessRule` objects with the **same taint key/effect and overlapping node selectors** can be created simultaneously during an API degradation window, causing unpredictable taint flip-flopping at runtime.

## Current Behavior

In `internal/webhook/nodereadinessgaterule_webhook.go`:

```go
func (w *NodeReadinessRuleWebhook) validateTaintConflicts(...) field.ErrorList {
    ruleList := &readinessv1alpha1.NodeReadinessRuleList{}
    if err := w.List(ctx, ruleList); err != nil {
        //  Silently allows the operation to proceed
        ctrl.Log.Error(err, "Failed to list rules for conflict validation")
        return allErrs
    }
    ...
}
```

When `w.List()` fails (e.g., degraded API server, cold cache, transient network error), the function returns an **empty error list**, allowing the admission to succeed without conflict validation.

## Impact

- Two rules with identical `spec.taint.key + spec.taint.effect` and overlapping `spec.nodeSelector` can be created
- Both `RuleReconciler` and `NodeReconciler` will independently apply/remove the same taint on the same nodes
- The taint will **flip-flop** with each reconcile cycle, making affected nodes permanently oscillate between schedulable and unschedulable states
- There is no recovery mechanism , the conflict persists until one rule is manually deleted

## Expected Behavior

Admission webhooks should **fail closed** for safety, especially for a security-sensitive controller that gates node scheduling. If conflict validation cannot be performed, the operation should be rejected with a clear error message so the user can retry.

## Proposed Fix

```go
if err := w.List(ctx, ruleList); err != nil {
    allErrs = append(allErrs, field.InternalError(
        field.NewPath("spec", "taint", "key"),
        fmt.Errorf("unable to validate taint conflicts, please retry: %w", err),
    ))
    return allErrs
}
```

This follows the Kubernetes security principle of **fail-closed admission** ,  if the webhook cannot make a safe decision, it should reject the request and let the client retry when the system is healthy.

## References

- [Kubernetes Admission Webhook Best Practices](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#best-practices-and-warnings)
- Related: `validateTaintConflicts` in `internal/webhook/nodereadinessgaterule_webhook.go`

dorodb-web22:
Fixes #251

When `validateTaintConflicts` cannot list existing rules, it previously logged the error and allowed the request through (fail-open). 
This means two rules with the same taint key/effect and overlapping selectors could be created during an API degradation window, causing taints to flip-flop indefinitely on affected nodes.

This change returns an `InternalError` instead, rejecting the request with a clear message so the client can retry when the system is healthy.
This aligns with the Kubernetes admission webhook best-practice offail-closed validation.

### DIFF
--- a/internal/webhook/nodereadinessgaterule_webhook.go
+++ b/internal/webhook/nodereadinessgaterule_webhook.go
@@ -80,9 +80,13 @@ func (w *NodeReadinessRuleWebhook) validateTaintConflicts(ctx context.Context, r
 	// List all existing rules
 	ruleList := &readinessv1alpha1.NodeReadinessRuleList{}
 	if err := w.List(ctx, ruleList); err != nil {
-		// If we can't list rules, allow the operation but log the issue
+		// Fail closed: if we can't list rules, we cannot safely validate
+		// for conflicts. Reject the request so the client can retry.
 		ctrl.Log.Error(err, "Failed to list rules for conflict validation")
-		return allErrs
+		return append(allErrs, field.InternalError(
+			field.NewPath("spec", "taint", "key"),
+			fmt.Errorf("failed to validate taint %q against existing rules: %w", rule.Spec.Taint.Key, err),
+		))
 	}
 
 	taintField := field.NewPath("spec", "taint", "key")


internal/webhook/nodereadinessgaterule_webhook.go
```diff
@@ -81,9 +80,12 @@ func (w *NodeReadinessRuleWebhook) validateTaintConflicts(ctx context.Context, r
 	// List all existing rules
 	ruleList := &readinessv1alpha1.NodeReadinessRuleList{}
 	if err := w.List(ctx, ruleList); err != nil {
-		// If we can't list rules, allow the operation but log the issue
-		ctrl.Log.Error(err, "Failed to list rules for conflict validation")
-		return allErrs
+		// Fail closed: if we can't list rules, we cannot safely validate
+		// for conflicts. Reject the request so the client can retry.
+		return append(allErrs, field.InternalError(
+			field.NewPath("spec", "taint", "key"),
```
AvineshTripathi, line 87: I'm not sure about using `field.NewPath("spec", "taint", "key")` here. A List failure is a systemic/infrastructure issue, not an error with the user's input in that specific field. Attributing it to the taint key could be misleading for users debugging their YAML.

dorodb-web22 (commented):
True, [@AvineshTripathi](https://github.com/AvineshTripathi)! using that path would mislead users into thinking their YAML is malformed when the real problem is on the controller side.

Fixed by switching to nil as the field path in field.InternalError, so the error correctly signals an internal issue rather than a field-level validation failure.

ajaysundark:
> Fixed by switching to nil as the field path in field.InternalError

@dorodb-web22 the PR doesnt seem to reflect this change? Could you PTAL.

ajaysundark:
/assign @dorodb-web22 

Please reassign the PR for approval once ready for final review. Thanks

dorodb-web22:
@ajaysundark Fixed, switched to nil path.

internal/webhook/nodereadinessgaterule_webhook.go
```diff
@@ -81,9 +80,12 @@ func (w *NodeReadinessRuleWebhook) validateTaintConflicts(ctx context.Context, r
...
 	if err := w.List(ctx, ruleList); err != nil {
-		// If we can't list rules, allow the operation but log the issue
-		ctrl.Log.Error(err, "Failed to list rules for conflict validation")
-		return allErrs
+		// Fail closed: if we can't list rules, we cannot safely validate
+		// for conflicts. Reject the request so the client can retry.
+		return append(allErrs, field.InternalError(
+			nil,
+			fmt.Errorf("unable to validate taint conflicts, please retry: %w", err),
+		))
```
Priyankasaggu11929, line 89: `nil` path is confusing in the actual admission error:

> Error from server (Forbidden): error when creating "test-nrr.yaml": admission webhook "vnodereadinessrule.kb.io" denied the request: validation failed: [\<nil\>: Internal error: unable to validate taint conflicts, please retry]


I suggest to add back the `spec.taint.key` path but change the error line to read coherently (because indeed the admission validation error happened at the taint key field in the nrr object schema) 


```suggestion
		// Fail closed: if we can't list rules, we cannot safely validate
		// for conflicts. Reject the request so the client can retry.
		return append(allErrs, field.InternalError(
			field.NewPath("spec", "taint", "key"),
			fmt.Errorf("failed to validate taint %q against existing rules: %w", rule.Spec.Taint.Key, err),
		))			
```

so, it outputs an error like:


> Error from server (Forbidden): error when creating "test-nrr.yaml": admission webhook "vnodereadinessrule.kb.io" denied the request: validation failed: [spec.taint.key: Internal error: failed to validate taint "readiness.k8s.io/example-taint" against existing rules: context deadline exceeded]

Priyankasaggu11929 (changes requested):

internal/webhook/nodereadinessgaterule_webhook.go
```diff
@@ -81,9 +80,12 @@ func (w *NodeReadinessRuleWebhook) validateTaintConflicts(ctx context.Context, r
 	// List all existing rules
 	ruleList := &readinessv1alpha1.NodeReadinessRuleList{}
 	if err := w.List(ctx, ruleList); err != nil {
-		// If we can't list rules, allow the operation but log the issue
-		ctrl.Log.Error(err, "Failed to list rules for conflict validation")
```
Priyankasaggu11929, line 84: Also, let's keep the log line as it is - `ctrl.Log.Error(err, "Failed to list rules for conflict validation")`

it will show up in the controller logs and would be useful for debugging.

Priyankasaggu11929 (changes requested):

dorodb-web22:
@Priyankasaggu11929 Done, restored the taint key path with a context-rich error message and added back the log line.

Priyankasaggu11929:
Thank you, @dorodb-web22.

Please squash the commit to one.

dorodb-web22:
@Priyankasaggu11929 Done, squashed.

Priyankasaggu11929:
@ajaysundark, for approve

/lgtm

ajaysundark:
/approve

Thanks!

PR #254, opened 2026-05-14
fix: pre-populate rule cache on first NodeReconciler pass to prevent bootstrap-only taint bypass

### LINKED ISSUE #253: bug: bootstrap-only taint enforcement silently skipped for nodes that join during the initial rule cache warm-up window
## Summary

`RuleReadinessController` maintains an in-memory `ruleCache` that is populated
lazily as `RuleReconciler` processes each `NodeReadinessRule`. `NodeReconciler`
derives all applicable rules exclusively from this cache via
`getApplicableRulesForNode`. Because both controllers start draining their work
queues concurrently after informer sync, there is a race window on startup and
after leader re-election where node events are processed against an empty or
partially-warm cache.

For `continuous` enforcement this is operationally benign the controller
re-evaluates the node on the next condition, taint, or label change and
converges. For `bootstrap-only` rules it is a permanent correctness gap: if a
node is evaluated before its applicable rule reaches the cache, the taint is
never applied, the bootstrap completion annotation is never written, and the
intended admission gate is silently bypassed with no error surfaced.

## Current Behavior

Startup sequence:

1. `mgr.Start()` starts informers and waits for cache synchronization all
   `NodeReadinessRule` objects are now readable from the local informer cache.
2. `RuleReconciler` and `NodeReconciler` begin draining their respective work
   queues **concurrently**.
3. `RuleReconciler` reconciles existing rules one by one
   (`MaxConcurrentReconciles: 1`), calling `updateRuleCache` for each.
4. `NodeReconciler` immediately processes queued Node events including events
   for nodes that joined during or just before startup against whatever rules
   happen to be in the cache at that moment.

With a non-trivial rule count the `ruleCache` reaches full coverage only after
all rules have been individually reconciled. Any node event processed before a
given rule is cached is invisible to that rule.

## Why Bootstrap-Only Mode Is Specifically Vulnerable

The bootstrap completion path in `evaluateRuleForNode` is:

conditions not satisfied → taint added
conditions satisfied     → taint removed → markBootstrapCompleted → annotation written
subsequent reconciles    → isBootstrapCompleted returns true → skip

`markBootstrapCompleted` is reached only inside the
`shouldRemoveTaint && currentlyHasTaint` branch. If a node is first evaluated
while its applicable rule is absent from the cache:

- `getApplicableRulesForNode` returns an empty slice  no taint is applied.
- The node's conditions are satisfied (CNI ready, security agent healthy, etc.).
- When the rule eventually enters the cache and the node is re-evaluated,
  `evaluateRuleForNode` sees `shouldRemoveTaint=true, currentlyHasTaint=false`
  and falls through to the `default` branch: no action, no call to
  `markBootstrapCompleted`.
- The bootstrap annotation is never written. The rule re-evaluates the node on
  every subsequent event indefinitely but the intended gate was never enforced.

## Steps to Reproduce / Concrete Scenario

1. Cluster undergoes a rolling restart of the controller (upgrade, crash, or
   leader re-election).
2. Within the startup window a replacement node joins. Its CNI plugin reports
   `NetworkReady=True` within ~8 seconds (typical for a pre-warmed agent).
3. `NodeReconciler` picks up the node CREATE event; the `network-readiness`
   bootstrap-only rule has not yet been reconciled by `RuleReconciler` (queue
   still has earlier items ahead of it) and is absent from `ruleCache`.
4. `getApplicableRulesForNode` returns empty  no taint is applied.
5. The rule is cached ~30 seconds later. Re-evaluation: `NetworkReady=True`,
   no taint present → `default` branch → no action, no annotation.
6. The node accepts workload scheduling without the readiness gate ever having
   been enforced.

This reproduces on any controller restart in a cluster actively scaling out and
on every leader re-election when the replacement leader starts with a cold cache.

## Expected Behavior

Nodes should never be evaluated by `NodeReconciler` against an empty or
incomplete cache when `NodeReadinessRules` already exist in the cluster.
Bootstrap-only rules must apply their taint reliably regardless of the ordering
in which the two controllers drain their queues on startup.

## Root Cause

The `ruleCache` is populated entirely as a side effect of
`RuleReconciler.Reconcile()` via `updateRuleCache`. There is no mechanism to
seed it from the already-synced informer cache before `NodeReconciler` begins
processing events. `mgr.Start()` guarantees informers are synced before
controllers run, so all `NodeReadinessRules` are available locally the moment
both controllers start the gap is that `ruleCache` is not initialized from
this data.

## Possible Implementation Direction

A `WarmCache` method on `RuleReadinessController` that lists existing rules from
the already-synced informer cache (no live API call) and calls `updateRuleCache`
for each would close the gap:

```go
func (r *RuleReadinessController) WarmCache(ctx context.Context) error {
    ruleList := &readinessv1alpha1.NodeReadinessRuleList{}
    if err := r.List(ctx, ruleList); err != nil {
        return fmt.Errorf("failed to warm rule cache: %w", err)
    }
    for i := range ruleList.Items {
        r.updateRuleCache(ctx, &ruleList.Items[i])
    }
    return nil
}
```
This can be invoked via a manager.Runnable registered with mgr.Add() that
executes after cache sync and completes before the manager signals readiness.
Because the list is served from the local informer cache, cost is bounded by
rule count rather than node count and adds negligible startup latency. The
warm-up should be re-run on each leadership term so that leader re-election does
not reintroduce the window.

An alternative is a lazy fallback inside NodeReconciler.Reconcile: on the
first pass detect that len(ruleCache) == 0 while rules exist in the informer
cache and perform an inline warm-up before evaluating the node.

## Acceptance Criteria
 A test (integration or e2e) covers nodes joining during the rule cache warm-up window and verifies that bootstrap-only taints are correctly applied.
 No node event is processed against an empty cache when NodeReadinessRules exist in the cluster.
 Cache warm-up cost is bounded by rule count, not node count (served from local informer cache, not live API).
 Leader re-election does not reintroduce the race window  the replacement leader warms its cache before declaring ready.
 No regression in continuous enforcement mode behavior.

## Additional Context
The ruleCache also drives the DeletionTimestamp guard in
processNodeAgainstAllRules that prevents taint operations on rules being
deleted. A fully-warm cache on startup is equally important for that invariant.
The recent hardening of bootstrap annotation cleanup on rule deletion
([ea74209](https://github.com/kubernetes-sigs/node-readiness-controller/commit/ea74209))
addresses the deletion side of the bootstrap lifecycle; this issue addresses the
complementary startup-time gap in the same semantic guarantee.

/kind bug

Karman580:
## Description

`RuleReadinessController` maintains an in-memory `ruleCache` populated lazily as `RuleReconciler` processes each `NodeReadinessRule`. `NodeReconciler` derives applicable rules exclusively from this cache. Both controllers start consuming their work queues concurrently after informer sync, creating a race where node events are processed against an empty or partially-warm cache.

For `continuous` enforcement this window is self-healing — the next condition, taint, or label change re-triggers evaluation and the controller converges. For `bootstrap-only` rules it is a **permanent correctness gap**: if a node is evaluated before its applicable rule reaches the cache, the taint is never applied, `markBootstrapCompleted` is never reached, the bootstrap completion annotation is never written, and the intended admission gate is silently bypassed with no error surfaced.

### Root Cause

`ensureCacheWarmed` did not exist. The `ruleCache` was built entirely as a side effect of `RuleReconciler.Reconcile()`. Although the controller-runtime manager guarantees informers are synced before controllers start, there was no mechanism to seed `ruleCache` from the already-available informer data before `NodeReconciler` began processing events.

### Fix

Add `ensureCacheWarmed` to `RuleReadinessController`. On the **first call** it lists all `NodeReadinessRules` from the already-synced informer cache (no live API round-trip) and adds any rule not yet present in `ruleCache`. Subsequent calls are no-ops guarded by an `atomic.Bool`. Existing cache entries — including rules already marked for deletion by `RuleReconciler` — are intentionally preserved so that in-flight state is not overwritten.

`ensureCacheWarmed` is called at the top of `NodeReconciler.Reconcile` before any rule evaluation takes place.

## Related Issue

Fixes #253

## Type of Change

/kind bug

## Testing

- Added `cold-start rule cache warm-up` integration test context in `node_controller_test.go` with two cases:
  - Verifies that a bootstrap-only taint **is applied** even when `ruleCache` is empty at reconcile time (the rule is discovered via warm-up).
  - Verifies that the warm-up is **not repeated** on subsequent reconciles (`cacheWarmed` remains true).
- All 49 existing tests continue to pass.
- Coverage: `internal/controller` 73.5% → 73.9%.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fix bootstrap-only taint enforcement correctness gap: NodeReconciler now pre-populates the rule cache from the informer cache on its first reconcile, preventing nodes from bypassing the admission gate during the controller startup race window.
```

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -91,6 +91,16 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	log := ctrl.LoggerFrom(ctx)
 	log.Info("Reconciling node", "node", req.Name)
 
+	// Pre-populate the rule cache from the informer cache on the first reconcile.
+	// This prevents the startup race where this controller processes node events
+	// before RuleReconciler has reconciled existing rules into the cache. For
+	// bootstrap-only rules the race is non-recoverable: if the node is evaluated
+	// with an empty cache the taint is never applied and the bootstrap completion
+	// annotation is never written.
+	if err := r.Controller.ensureCacheWarmed(ctx); err != nil {
+		return ctrl.Result{}, err
+	}
+
 	// Fetch the node
 	node := &corev1.Node{}
 	if err := r.Get(ctx, req.NamespacedName, node); err != nil {
--- a/internal/controller/node_controller_test.go
+++ b/internal/controller/node_controller_test.go
@@ -1007,4 +1007,148 @@ var _ = Describe("Node Controller", func() {
 			Expect(err.Error()).To(ContainSubstring("status patch failed"))
 		})
 	})
+
+	// These tests verify that NodeReconciler pre-populates the rule cache from the
+	// informer cache on its first reconcile, closing the startup race where node
+	// events arrive before RuleReconciler has warmed the cache. The scenario is
+	// particularly critical for bootstrap-only rules: if the cache is cold when a
+	// node is first evaluated, the taint is never applied and the bootstrap
+	// completion annotation is never written, permanently bypassing the admission gate.
+	Context("cold-start rule cache warm-up", func() {
+		var (
+			ctx                 context.Context
+			readinessController *RuleReadinessController
+			nodeReconciler      *NodeReconciler
+			coldStartNode       *corev1.Node
+			coldStartRule       *nodereadinessiov1alpha1.NodeReadinessRule
+			coldStartNodeName   types.NamespacedName
+		)
+
+		const (
+			coldTaintKey      = "readiness.k8s.io/cold-start-taint"
+			coldConditionType = "ColdStartCondition"
+		)
+
+		BeforeEach(func() {
+			ctx = context.Background()
+
+			// Intentionally do NOT call updateRuleCache — this simulates the
+			// NodeReconciler processing a node event before the RuleReconciler
+			// has reconciled the rule and populated the cache.
+			readinessController = &RuleReadinessController{
+				Client:        k8sClient,
+				Scheme:        k8sClient.Scheme(),
+				clientset:     fake.NewSimpleClientset(),
+				ruleCache:     make(map[string]*nodereadinessiov1alpha1.NodeReadinessRule),
+				EventRecorder: record.NewFakeRecorder(10),
+			}
+
+			nodeReconciler = &NodeReconciler{
+				Client:     k8sClient,
+				Scheme:     k8sClient.Scheme(),
+				Controller: readinessController,
+			}
+
+			coldStartNode = &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:   "cold-start-node",
+					Labels: map[string]string{"cold-start-group": "test"},
+				},
+				Status: corev1.NodeStatus{
+					Conditions: []corev1.NodeCondition{
+						// Condition not yet satisfied — taint should be applied.
+						{Type: coldConditionType, Status: corev1.ConditionFalse},
+					},
+				},
+			}
+
+			coldStartRule = &nodereadinessiov1alpha1.NodeReadinessRule{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:       "cold-start-rule",
+					Finalizers: []string{finalizerName},
+				},
+				Spec: nodereadinessiov1alpha1.NodeReadinessRuleSpec{
+					Conditions: []nodereadinessiov1alpha1.ConditionRequirement{
+						{Type: coldConditionType, RequiredStatus: corev1.ConditionTrue},
+					},
+					Taint: corev1.Taint{
+						Key:    coldTaintKey,
+						Effect: corev1.TaintEffectNoSchedule,
+					},
+					NodeSelector: metav1.LabelSelector{
+						MatchLabels: map[string]string{"cold-start-group": "test"},
+					},
+					EnforcementMode: nodereadinessiov1alpha1.EnforcementModeBootstrapOnly,
+				},
+			}
+
+			coldStartNodeName = types.NamespacedName{Name: coldStartNode.Name}
+
+			Expect(k8sClient.Create(ctx, coldStartNode)).To(Succeed())
+			Expect(k8sClient.Create(ctx, coldStartRule)).To(Succeed())
+		})
+
+		AfterEach(func() {
+			_ = k8sClient.Delete(ctx, coldStartNode)
+
+			updatedRule := &nodereadinessiov1alpha1.NodeReadinessRule{}
+			if err := k8sClient.Get(ctx, types.NamespacedName{Name: coldStartRule.Name}, updatedRule); err == nil {
+				updatedRule.Finalizers = nil
+				_ = k8sClient.Update(ctx, updatedRule)
+				_ = k8sClient.Delete(ctx, updatedRule)
+			}
+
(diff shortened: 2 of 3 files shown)


Karman580:
recheck

Priyankasaggu11929:
Thanks again for writing the gist, @Karman580. I tried to reproduce it locally.

You' are using `Ready=True` node status condition in your example NRR rule.
In a regular kind cluster setup, the node is  (almost always) healthy (i.e, `Ready=True` is already satisfied) before the controller even starts, and so the taint is never applied in the first place and the cold-cache timing window is never hit.

what you are actually seeing with your repro steps is a different scenario (the [default case](https://github.com/kubernetes-sigs/node-readiness-controller/blob/b4a52fa25628b70b5dde535b5da3f181bf187547/internal/controller/nodereadinessrule_controller.go#L372-L375) (case 4) in the switch block) - `node is already healthy on first evaluation` 

IMO, a better reproduction setup would be to use a custom node status condition that kind does not satisfy by default, so, we can actually observe the stated race condition.

I did try  do this ^ locally.

While the PR change address the one scenario you described (ie. on a controller restart, (Node Controller) [Reconciler](https://github.com/kubernetes-sigs/node-readiness-controller/pull/254/changes#diff-a6b2f7655e9ad10430883e9c49e3a1a24f3bbb47be140ce8d67f96d9477b2320R94-R103) will first seed the in-memory rule cache from the informer, before it proceeds with further evaluation on the node, but we still need to address the following gap (which remains regardless of the cache is warm or not)

bootstrap annotation is only added in [case 1](https://github.com/kubernetes-sigs/node-readiness-controller/blob/b4a52fa25628b70b5dde535b5da3f181bf187547/internal/controller/nodereadinessrule_controller.go#L341-L353) of the switch block

```
  1. shouldRemoveTaint=true  + currentlyHasTaint=true  -> remove taint and add annotation
  2. shouldRemoveTaint=false + currentlyHasTaint=false -> add taint
  3. shouldRemoveTaint=false + currentlyHasTaint=true  -> adopt taint
  4. shouldRemoveTaint=true  + currentlyHasTaint=false -> "No taint action needed" # no annotation
  ```

In case of a node which is already healthy on first evaluation (the case that your gist is actually testing with `Ready=True` node condition), it will always go to the default case, so the annotation is never added.

And what happens then is if that node later has a transient condition failure (e.g. a kubelet restart or network interruption that flips the condition to unknown/False), the controller has no record of its bootstrap stage being completed and it will re-apply the taint and block any workload.

I think we should discuss it in the upcoming project call about whether we should design the default case (`shouldRemoveTaint=true + currentlyHasTaint=false`) to also write the annotation in bootstrap-only mode.

cc: @ajaysundark

Karman580:
Thanks for the detailed review and for reproducing it locally.

You're right  the current repro using Ready=True overlaps with the default healthy-node path rather than isolating the cache timing race itself. I also see the larger bootstrap annotation gap in the shouldRemoveTaint=true + currentlyHasTaint=false case that you pointed out.

I’ll revisit the reproduction using a custom condition to isolate the cache behavior more clearly, and I agree the bootstrap annotation semantics are probably better discussed separately before I iterate further on the PR.

ajaysundark:
/close

Superseded by #204

PR #256, opened 2026-05-15
fix: pass NodeList to processAllNodesForRule in tests

sahitya-chandra:
## Description

`main` is currently broken. Lint, Tests and E2E all fail at `go vet`:

```
vet: internal/controller/nodereadinessrule_controller_test.go:1881:58: not enough arguments in call to failController.processAllNodesForRule
	have (context.Context, *NodeReadinessRule)
	want (context.Context, *NodeReadinessRule, *corev1.NodeList)
```

#226 changed `processAllNodesForRule` to take a `*corev1.NodeList`, but the two new tests added in #216 still call it with the old two-argument signature. Both PRs were green on their own branches; the test file only stopped compiling once #216 landed on top of #226 on `main`.

Fix wraps the test node in a `*corev1.NodeList` and passes it through, same as `reconcileNormal` does. Filtering still happens inside the function, so behaviour is unchanged.

## Related Issue

None

## Type of Change

/kind bug
/kind failing-test

## Testing

- `make lint` clean
- `make test` passes

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
NONE
```

### DIFF
--- a/internal/controller/nodereadinessrule_controller_test.go
+++ b/internal/controller/nodereadinessrule_controller_test.go
@@ -1878,7 +1878,8 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 				EventRecorder: record.NewFakeRecorder(10),
 			}
 
-			Expect(failController.processAllNodesForRule(ctx, rule)).To(Succeed())
+			nodeList := &corev1.NodeList{Items: []corev1.Node{*failNode}}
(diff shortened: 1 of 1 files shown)


ajaysundark:
Thanks @sahitya-chandra !

/lgtm
/approve

PR #257, opened 2026-05-15
CI: Output test coverprofile to Artifacts tab in Prow

vitorfloriano:
When run in Prow, `make test` will output the `coverprofile` file to `$ARTIFACTS`, which is a directory already present and exported by Prow when using `decorate: true` in the job description, and will show up in the Artifacts tab in spyglass. 

When run locally, `make test` will output the file to the root of the repo, as it is currently.

See: https://docs.prow.k8s.io/docs/components/pod-utilities/#what-the-test-container-can-expect

Follow-up to https://github.com/kubernetes/test-infra/pull/37035

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,7 @@ TEST_DIR := test
 TOOLS_DIR := hack/tools
 TOOLS_BIN_DIR := $(abspath $(TOOLS_DIR)/$(BIN_DIR))
 GO_INSTALL := ./scripts/go_install.sh
+ARTIFACTS ?= .
 
 export PATH := $(abspath $(TOOLS_BIN_DIR)):$(PATH)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -27,6 +27,7 @@ TEST_DIR := test
 TOOLS_DIR := hack/tools
 TOOLS_BIN_DIR := $(abspath $(TOOLS_DIR)/$(BIN_DIR))
 GO_INSTALL := ./scripts/go_install.sh
+ARTIFACTS ?= .
```
AvineshTripathi, line 30: i would put it under either hack/tools/artifacts or something similar

vitorfloriano: What about `./artifacts/` only?

Priyankasaggu11929: whatever we choose, let's make sure it is self-explanatory and added to .gitignore.

Within Kubernetes projects, I have seen a pattern of  `_output/*` dir for collecting test artifacts, logs or binaries etc.

Priyankasaggu11929:
I'm good with the changes as it is so far.

But if you want to improve the default ARTIFACTS dir, adding a hold. (feel free to unhold)
/hold

/lgtm

(@ajaysundark, for approve label. Thanks!)

ajaysundark:
/lgtm
/approve

Thanks!

PR #263, opened 2026-05-20
feat (reporter): reduce API-server load by skipping unchanged node conditions

LightCreator1007:
## Description
Adds an idempotency gate to `updateNodeCondition` in the `readiness-condition-reporter`. 
`UpdateStatus` is now bypassed if the `Status`, `Reason`, and `Message` are unchanged, preventing unnecessary API server flooding and `etcd` write amplification on every tick.

## Related Issue
NONE

## Type of Change
/kind bug

## Testing
- Modified `main_test.go` locally to assert against the `fake.Clientset`'s tracked actions, confirming that the `UpdateStatus` call is definitively bypassed when the condition state is unchanged.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
NONE
```

### DIFF
--- a/cmd/readiness-condition-reporter/main.go
+++ b/cmd/readiness-condition-reporter/main.go
@@ -230,8 +230,20 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
 
 		// If the semantic state is completely unchanged, bypass the API write
 		// to prevent etcd write amplification and control plane flooding.
+		needsUpdate := true
 		if existingCondition != nil && existingCondition.Status == status && existingCondition.Reason == health.Reason && existingCondition.Message == health.Message {
-			// since state did not change return early to avoid unnecessary API call
+			needsUpdate = false
+			/*
+				NOTE: Skipping the write stops refreshing the LastHeartbeatTime on every tick.
+				To mitigate this, force an update every 5 minutes even if the state is unchanged.
+			*/
+			if time.Since(existingCondition.LastHeartbeatTime.Time) >= 5*time.Minute {
+				needsUpdate = true
+			}
+		}
+
+		if !needsUpdate {
+			//state has not changed for 5 mins, skip the write
 			return nil
 		}
 
--- a/cmd/readiness-condition-reporter/main.go
+++ b/cmd/readiness-condition-reporter/main.go
@@ -243,7 +243,7 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
 		}
 
 		if !needsUpdate {
-			//state has not changed for 5 mins, skip the write
+			// state has not changed for 5 mins, skip the write
 			return nil
 		}
 
--- a/cmd/readiness-condition-reporter/main.go
+++ b/cmd/readiness-condition-reporter/main.go
@@ -215,13 +215,24 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
 
 		// Find existing condition to preserve transition time if status hasn't changed
 		var transitionTime metav1.Time
+		var existingCondition *corev1.NodeCondition
+
 		for _, condition := range node.Status.Conditions {
 			if string(condition.Type) == conditionType {
+				condCopy := condition
+				existingCondition = &condCopy
 				if condition.Status == status {
 					transitionTime = condition.LastTransitionTime
 				}
 				break
 			}
+		}
+
+		// If the semantic state is completely unchanged, bypass the API write
+		// to prevent etcd write amplification and control plane flooding.
+		if existingCondition != nil && existingCondition.Status == status && existingCondition.Reason == health.Reason && existingCondition.Message == health.Message {
+			// since state did not change return early to avoid unnecessary API call
+			return nil
 		}
 
 		if transitionTime.IsZero() {
--- a/cmd/readiness-condition-reporter/main.go
+++ b/cmd/readiness-condition-reporter/main.go
@@ -35,13 +35,15 @@ import (
 )
 
 const (
-	envNodeName          = "NODE_NAME"
-	envConditionType     = "CONDITION_TYPE"
-	envCheckEndpoint     = "CHECK_ENDPOINT"
-	envCheckInterval     = "CHECK_INTERVAL"
-	envImpersonateNode   = "IMPERSONATE_NODE"
-	defaultCheckInterval = 30 * time.Second
-	defaultHTTPTimeout   = 10 * time.Second
+	envNodeName            = "NODE_NAME"
+	envConditionType       = "CONDITION_TYPE"
+	envCheckEndpoint       = "CHECK_ENDPOINT"
+	envCheckInterval       = "CHECK_INTERVAL"
+	envImpersonateNode     = "IMPERSONATE_NODE"
+	envHeartbeatPeriod     = "HEARTBEAT_PERIOD"
+	defaultCheckInterval   = 30 * time.Second
+	defaultHTTPTimeout     = 10 * time.Second
+	defaultHeartbeatPeriod = 5 * time.Minute
 )
 
 // HealthResponse represents the health check response structure.
@@ -86,6 +88,19 @@ func main() {
 		}
 	}
 
+	heartbeatPeriodStr := os.Getenv(envHeartbeatPeriod)
+	heartbeatPeriod := defaultHeartbeatPeriod
+	if heartbeatPeriodStr != "" {
+		parsedPeriod, err := time.ParseDuration(heartbeatPeriodStr)
+		if err == nil {
+			heartbeatPeriod = parsedPeriod
+		} else {
+			klog.ErrorS(err, "Failed parse heartbeat period, using default",
+				"input", heartbeatPeriodStr,
+				"default", defaultHeartbeatPeriod)
+		}
+	}
+
 	// Create Kubernetes client
 	config, err := rest.InClusterConfig()
 	if err != nil {
@@ -121,20 +136,20 @@ func main() {
 	defer ticker.Stop()
 
 	// Run immediately on startup, then on each tick
-	runCheck(ctx, httpClient, clientset, checkEndpoint, nodeName, conditionType)
+	runCheck(ctx, httpClient, clientset, checkEndpoint, nodeName, conditionType, heartbeatPeriod)
 	for {
 		select {
 		case <-ctx.Done():
 			klog.InfoS("Shutting down readiness condition reporter", "reason", ctx.Err())
 			return
 		case <-ticker.C:
-			runCheck(ctx, httpClient, clientset, checkEndpoint, nodeName, conditionType)
+			runCheck(ctx, httpClient, clientset, checkEndpoint, nodeName, conditionType, heartbeatPeriod)
 		}
 	}
 }
 
 // runCheck performs a single health check and updates the node condition.
-func runCheck(ctx context.Context, httpClient *http.Client, clientset kubernetes.Interface, checkEndpoint, nodeName, conditionType string) {
+func runCheck(ctx context.Context, httpClient *http.Client, clientset kubernetes.Interface, checkEndpoint, nodeName, conditionType string, heartbeatPeriod time.Duration) {
 	health, err := checkHealth(ctx, httpClient, checkEndpoint)
 	if err != nil {
 		klog.ErrorS(err, "Health check failed", "endpoint", checkEndpoint)
@@ -145,7 +160,7 @@ func runCheck(ctx context.Context, httpClient *http.Client, clientset kubernetes
 		}
 	}
 
-	if err := updateNodeCondition(ctx, clientset, nodeName, conditionType, health); err != nil {
+	if err := updateNodeCondition(ctx, clientset, nodeName, conditionType, health, heartbeatPeriod); err != nil {
 		klog.ErrorS(err, "Failed to update node condition", "node", nodeName, "condition", conditionType)
 	}
 }
@@ -198,7 +213,7 @@ func checkHealth(ctx context.Context, client *http.Client, endpoint string) (*He
 }
 
 // updateNodeCondition updates the node condition based on health check.
-func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeName, conditionType string, health *HealthResponse) error {
+func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeName, conditionType string, health *HealthResponse, heartbeatPeriod time.Duration) error {
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		// Get the node
 		node, err := client.CoreV1().Nodes().Get(ctx, nodeName, metav1.GetOptions{})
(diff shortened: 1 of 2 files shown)


sahitya-chandra:
Hey @LightCreator1007, PR looks good but the diff only touches `main.go`, so the skip path isn't covered. Could you add a test where the node already has the matching condition and assert no update is sent?

Also, a comment noting that skipping the write stops refreshing the heartbeat (so it no longer signals the reporter is alive), would be good

cmd/readiness-condition-reporter/main.go
```diff
@@ -215,15 +215,26 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
...
 				}
 				break
 			}
 		}
 
+		// If the semantic state is completely unchanged, bypass the API write
+		// to prevent etcd write amplification and control plane flooding.
+		if existingCondition != nil && existingCondition.Status == status && existingCondition.Reason == health.Reason && existingCondition.Message == health.Message {
+			// since state did not change return early to avoid unnecessary API call
+			return nil
```
AvineshTripathi, line 263: with the change we will miss the `lastHeartbeatTime` update. While this is an optional field. It is an important field to show indicating that the health check is running.

how about we add 
```
if time.Since(existingCondition.LastHeartbeatTime.Time) >= 5*time.Minute {
  needsUpdate = true // State is unchanged, but we need to refresh the heartbeat
}
```

so after 5 min even if there is no update to status, we still update once for the LastHeartbeatTime?

sahitya-chandra: Yeah, that works well. The 5m floor keeps the heartbeat alive so it still signals the reporter is running, while skipping the redundant writes the rest of the time. +1 from me.

LightCreator1007: yes i added that in the latest commit.
Thanks for the input!

LightCreator1007:
> skip path isn't covered. Could you add a test where the node already has the matching condition and assert no update is sent

Yes!
I have added the same in the newer commits to cover both the idempotency and the 5 min refresh cases :)

Thanks for the help!!

Priyankasaggu11929:
I ran the reporter locally with a 10s check interval, to track how many writes actually went to the node.

from my testing, it's one write, per tick (10s interval), per node.

For a component that exists to periodically report node health, that's exactly what you would expect. 

So, in my opinion, it is not a problem to optimise away with 1 write/tick. Is there a specific scale scenario motivating this?

LightCreator1007:
> I ran the reporter locally with a 10s check interval, to track how many writes actually went to the node.
> 
> from my testing, it's one write, per tick (10s interval), per node.
> 
> For a component that exists to periodically report node health, that's exactly what you would expect.
> 
> So, in my opinion, it is not a problem to optimise away with 1 write/tick. Is there a specific scale scenario motivating this?

Hi @Priyankasaggu11929 , thanks for testing this!

While 1 update per 10 seconds is fine locally, the project targets a scale of say 5,000 nodes.

At 5,000 nodes, a 10-second interval would create 500 requests per second hitting the server just for health reporting. Processing this constant stream of "no change" updates may quickly overwhelm the API server.

To prevent this, the reporter should cache its state locally and only send an update to the API server when its health status actually changes.

Priyankasaggu11929:
@LightCreator1007, thanks for the response.

To be clear, I'm trying to make sure I understand the actual problem.

> At 5,000 nodes, a 10-second interval would create 500 requests per second hitting the server just for health reporting. Processing this constant stream of "no change" updates may quickly overwhelm the API server.

500 req/s at 5000 nodes with 10s interval - this rate is well within normal baseline and not something I'd frame as API server flooding.

> To prevent this, the reporter should cache its state locally and only send an update to the API server when its health status actually changes.

There's no local caching. Even with this PR, we are hitting apiserver with same amount of [get requests](https://github.com/kubernetes-sigs/node-readiness-controller/blob/8b11ed1180bb122db24d40462d5aeac6e14dc422/cmd/readiness-condition-reporter/main.go#L204) every tick.
The state comparision is against the API server response, not a local cache.

The number of requests/second is not the problem actually.
It is what each write is actually adding to etcd storage. etcd stores the full node object on every write, so at scale, even updating a single condition field writes the entire node status blob.
The changes in the PR will help with this bit (although we should check what that number actually looks like in our scalability testing cc: @Karthik-K-N)

For this reason ^, defaulting to a 5-minute forced heartbeat is fine,  it aligns with other controllers as well (like [Node Problem Detector](https://github.com/kubernetes/node-problem-detector/blob/87bbc58e7ef78eb0da0805d32362f897d9f60b9e/cmd/options/options.go#L126))

One request though - make the heartbeat period configurable.
The 5-minute default is ok, but people may still have valid reasons to sync earlier, and hardcoding it removes that option.
(For example - `LastHeartbeatTime` is the only signal that the reporter is still alive and so having an option to allow people to tune how long a dead reporter goes undetected, is good.)

cc: @ajaysundark

cmd/readiness-condition-reporter/main.go
```diff
@@ -215,15 +215,38 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
...
+				To mitigate this, force an update every 5 minutes even if the state is unchanged.
+			*/
+			if time.Since(existingCondition.LastHeartbeatTime.Time) >= 5*time.Minute {
+				needsUpdate = true
+			}
+		}
+
+		if !needsUpdate {
+			// state has not changed for 5 mins, skip the write
+			return nil
```
Priyankasaggu11929, line 263: let's add a ~long~ log line here rather than returning silently?

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -188,3 +188,101 @@ func TestUpdateNodeCondition(t *testing.T) {
...
+	health := &HealthResponse{Healthy: true, Reason: "EndpointOK", Message: "All good"}
+
+	countUpdates := func(c *fake.Clientset) int {
+		n := 0
+		for _, a := range c.Actions() {
+			if a.GetVerb() == "update" && a.GetSubresource() == "status" && a.GetResource().Resource == "nodes" {
+				n++
+			}
+		}
+		return n
```
Priyankasaggu11929: let's extend the existing test [`TestUpdateNodeCondition`](https://github.com/kubernetes-sigs/node-readiness-controller/blob/8b11ed1180bb122db24d40462d5aeac6e14dc422/cmd/readiness-condition-reporter/main_test.go#L109) - add new rows to the table for these new test cases (easy to read/maintain in future)

Also, add a case for state change path to confirm that the logic still triggers and immediate write.

Priyankasaggu11929 (changes requested):

LightCreator1007:
hello @Priyankasaggu11929 ,
Thank you for taking out the time for going through my PR!

>There's no local caching. Even with this PR, we are hitting apiserver with same amount of  [get requests](https://github.com/kubernetes-sigs/node-readiness-controller/blob/8b11ed1180bb122db24d40462d5aeac6e14d>c422/cmd/readiness-condition-reporter/main.go#L204) every tick. The state comparision is against the API server response, not a local cache.

Thank you for correcting me there, I overlooked the fact that it still does send a GET request.

>The number of requests/second is not the problem actually.
>It is what each write is actually adding to etcd storage. etcd stores the full node object on every write, so at scale, even updating a single condition field writes the entire node status blob.

I was too fixated on the API request rate to realize that the actual problem would be the etcd storage, which is the bottleneck this PR will help out with. I was completely unaware of this, so thanks a lot for pointing that out :)

I have made the requested changes and have made the heartbeat time configurable with a default fallback to 5min as suggested.

cmd/readiness-condition-reporter/main.go
```diff
@@ -35,13 +35,15 @@ import (
...
-	envImpersonateNode   = "IMPERSONATE_NODE"
-	defaultCheckInterval = 30 * time.Second
-	defaultHTTPTimeout   = 10 * time.Second
+	envNodeName            = "NODE_NAME"
+	envConditionType       = "CONDITION_TYPE"
+	envCheckEndpoint       = "CHECK_ENDPOINT"
+	envCheckInterval       = "CHECK_INTERVAL"
+	envImpersonateNode     = "IMPERSONATE_NODE"
+	envHeartbeatPeriod     = "HEARTBEAT_PERIOD"
+	defaultCheckInterval   = 30 * time.Second
```
ajaysundark, line 44: if the intent is to reduce the frequency of readiness updates, why it cannot be set with a different check interval?

Could we first clarify the concept of 'heart-beat'? Would there be a case when the component health check be checked often but not update the API server know if it degraded?

LightCreator1007: >if the intent is to reduce the frequency of readiness updates, why it cannot be set with a different check interval?

A different check interval (in this case a larger one) would mean that if the node status changes we won't be able to detect it immediately. We need to check the node status frequently for fast detection, but if the state stays stable/healthy for a long time, we may want to skip writing that identical state to the API server on every tick. Skipping those redundant writes prevents etcd write amplification.

>Could we first clarify the concept of 'heart-beat'?

The heartbeat here would be a liveness proof, It ensures that if the component stays perfectly healthy for a long time, we still write an update every 5 minutes just to bump the `LastHeartbeatTime`, proving to the API server that the reporter hasn't crashed.

>Would there be a case when the component health check be checked often but not update the API server know if it degraded?

I do not think there is a case where a degraded state would be missed or delayed. If the health check degrades (Status, Reason, or Message changes), the idempotency gate would instantly open and the API server is updated on that exact tick.

ajaysundark:
I agree this is useful in a large scale cluster. Reducing frequency of no-op updates from nodes could help the API load.

It would also be similar to NRD behavior, as it watches the pod-readiness events directly from Kubelet - only event updates are posted to API, not the heartbeat style regular health updates.

cmd/readiness-condition-reporter/main.go
```diff
@@ -215,15 +230,39 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
...
 					transitionTime = condition.LastTransitionTime
 				}
 				break
 			}
 		}
 
+		// If the semantic state is completely unchanged, bypass the API write
+		// to prevent etcd write amplification and control plane flooding.
+		needsUpdate := true
+		if existingCondition != nil && existingCondition.Status == status && existingCondition.Reason == health.Reason && existingCondition.Message == health.Message {
```
AnuragThePathak, line 249: From purely cognitive point of view, I felt it belongs to the part where we set the transitionTime during the linear search of the condition. Segregation feels unnecessary and rather inconvenient to understand the flow. Also probably can replace transitionTime separate variable with existing condition variable only.

AnuragThePathak:
I just completed the review. The functionality added looks good to me. I have minor suggestions for refactor which I will drop in review in a while.

@ajaysundark, also saw your input on EXIT_ON_READY, I'm not covering that in my review as not implemented.

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -127,11 +130,13 @@ func TestUpdateNodeCondition(t *testing.T) {
...
-			wantStatus: corev1.ConditionTrue,
-			wantReason: "EndpointOK",
+			heartbeatPeriod:  5 * time.Minute,
+			wantStatus:       corev1.ConditionTrue,
+			wantReason:       "EndpointOK",
+			wantUpdateCalled: true,
 		},
 		{
-			name: "Update Condition to Unhealthy",
+			name: "State change triggers immediate write",
```
AnuragThePathak, line 139: I think it may be a better idea to have a name such as "update condition on state change", because this test is primarily for the path updating a condition when state changes. Immediate write is a desired outcome of state change, not the cause. We can leverage a comment to include details like immediate write.

cmd/readiness-condition-reporter/main.go
```diff
@@ -215,15 +230,39 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
...
+			needsUpdate = false
+			/*
+				NOTE: Skipping the write stops refreshing the LastHeartbeatTime on every tick.
```
AnuragThePathak, line 252: The style of the comment is an anti-pattern in Go

cmd/readiness-condition-reporter/main.go
```diff
@@ -215,15 +230,39 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
...
+				To mitigate this, force an update every 5 minutes even if the state is unchanged.
```
AnuragThePathak, line 253: I think the mention of 5 minute was a miss in comments maybe? Because the code clearly use `heartbeatPeriod`

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -148,20 +153,94 @@ func TestUpdateNodeCondition(t *testing.T) {
...
+			wantUpdateCalled: true,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			client := fake.NewSimpleClientset(tt.existingNode)
 
-			err := updateNodeCondition(context.Background(), client, nodeName, conditionType, tt.health)
+			countUpdates := func() int {
```
AnuragThePathak, line 221: In future we may or may not benefit by from returning the integer value of countUpdate but for a now we simply need a boolean variable that tells if the update was actually called or not. Use of an integer can be confusing to newly onboarded developers.

That change will also drastically simply the checks below.

AnuragThePathak (commented):
No blocker

ajaysundark:
> I just completed the review. The functionality added looks good to me. I have minor suggestions for refactor which I will drop in review in a while.
> 
> @ajaysundark, also saw your input on EXIT_ON_READY, I'm not covering that in my review as not implemented.

Thanks for your review. I removed my earlier comment on exit-on-ready to avoid confusion. It can still be achieved by configuring a very large heartbeat period. it is not one-shot (container will not terminate), but a side-car with a larger periodic check is arguably better. if needed in the future we can plan as a separate improvement.

cmd/readiness-condition-reporter/main.go
ajaysundark: I think possibly referring the `defaultHeartbeatPeriod` interval, which is set as 5min.

ajaysundark:
@LightCreator1007 are you available to address the open comments? This looks mostly ready to merge, and if possible can be picked up for upcoming release.

cmd/readiness-condition-reporter/main.go
AnuragThePathak: That yes. Just felt better to refer heartbeatPeriod rather than saying 5min in that comment, as if user overrides the default 5 minute, that logic will evaluate to that overriden value only.

LightCreator1007:
Yep I will be adressing those in a few.

ajaysundark:
Since the pending changes are only source cleanup, I'm okay to merge this in and @LightCreator1007 could follow up as a separate PR.

Thanks for your contributions!

/lgtm
/approve

ajaysundark:
cc @arnab-logs for how we can plan to include this feature in our docs.

PR #266, opened 2026-05-26
cleanup: remove duplicate KIND_CLUSTER definition in Makefile

sahitya-chandra:
## Description
`KIND_CLUSTER` was defined twice with `?=`. The first definition (`nrr-test`)
always wins, so the second (`nrr-test-e2e`, above the e2e targets) never took
effect. This removes the dead line. No behavior change.

## Related Issue
None

## Type of Change
/kind cleanup

## Testing
`make -p -n | grep '^KIND_CLUSTER ='` reports `nrr-test` before and after.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -394,7 +394,6 @@ test: manifests generate fmt vet setup-envtest ## Run tests.
 # The default setup assumes Kind is pre-installed and builds/loads the Manager Docker image locally.
 # CertManager is installed by default; skip with:
 # - CERT_MANAGER_INSTALL_SKIP=true
-KIND_CLUSTER ?= nrr-test-e2e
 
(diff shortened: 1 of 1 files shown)


sahitya-chandra:
I just removed the line since `KIND_CLUSTER` already resolves to `nrr-test`, so e2e isn't really on a separate cluster anyway.

I did think about giving e2e its own `KIND_CLUSTER_E2E ?= nrr-test-e2e` to keep it isolated, but that changes behavior, so I left it out of this cleanup. can update the current changes if we want that.

ajaysundark (approved):
/lgtm

Thanks

PR #270, opened 2026-06-02
Add test-e2e-kind target with hack script and artifact collection

Priyankasaggu11929:
Following the k/jobset repo's pattern - https://github.com/kubernetes/test-infra/blob/98798bcae0a846ab0c1938f81044844a7df148f3/config/jobs/kubernetes-sigs/jobset/jobset-periodics-main.yaml

PR is adding a new `hack/e2e-test.sh` and a `test-e2e-kind` makefile target, that handles Kind cluster creation, artifact collection, and log export in one place.

This is prep work for the PR https://github.com/kubernetes/test-infra/pull/37165, to give us periodic e2e test signals for NRC across all latest supported Kubernetes releases.

---

Note:

With this, the existing `test-e2e` target is redundant now, but it is still in use in the GH Action workflow (we can clean it up later once all our GH actions CI jobs are migrated to Prow).

/assign @ajaysundark

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -1,0 +1,80 @@
+#!/usr/bin/env bash
+
+# Copyright The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+export KIND=${KIND:-kind}
+export KUBECTL=${KUBECTL:-kubectl}
+
+KIND_CLUSTER="${KIND_CLUSTER:-nrr-test}"
+USE_EXISTING_CLUSTER="${USE_EXISTING_CLUSTER:-false}"
+ARTIFACTS="${ARTIFACTS:-.}"
+E2E_KIND_VERSION="${E2E_KIND_VERSION:-v1.36}"
+
+if [[ "$E2E_KIND_VERSION" =~ ^v[0-9]+\.[0-9]+$ ]]; then
+    K8S_VERSION=$(curl -sf --retry 3 --retry-delay 5 --max-time 30 \
+        "https://api.github.com/repos/kubernetes/kubernetes/git/matching-refs/tags/${E2E_KIND_VERSION}." \
+        | grep -oP 'v[0-9]+\.[0-9]+\.[0-9]+(?=")' \
+        | sort -V | tail -1)
+    if [ -z "$K8S_VERSION" ]; then
+        echo "ERROR: could not resolve latest patch for ${E2E_KIND_VERSION}" >&2
+        exit 1
+    fi
+else
+    K8S_VERSION="$E2E_KIND_VERSION"
+fi
+
+# Use a temporary KUBECONFIG so that the script does not mess up the current user's kubeconfig.
+KUBECONFIG=""
+
+function cleanup {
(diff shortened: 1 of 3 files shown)


ajaysundark:
Id also consider this to be Reliability improvement on, so cc @vitorfloriano when you find time to get to this.

hack/e2e-test.sh
```diff
@@ -0,0 +1,80 @@
...
+set -o nounset
+set -o pipefail
+
+export KIND=${KIND:-kind}
+export KUBECTL=${KUBECTL:-kubectl}
+
+KIND_CLUSTER="${KIND_CLUSTER:-nrr-test}"
+USE_EXISTING_CLUSTER="${USE_EXISTING_CLUSTER:-false}"
+ARTIFACTS="${ARTIFACTS:-.}"
+E2E_KIND_VERSION="${E2E_KIND_VERSION:-v1.36}"
```
vitorfloriano, line 27: I'm wondering if there could be a way to not have v1.36 hardcoded here.

test/e2e/e2e_test.go
```diff
@@ -84,6 +84,21 @@ var _ = Describe("Manager", Ordered, func() {
 		cmd := exec.Command("kubectl", "delete", "pod", "curl-metrics", "-n", namespace)
 		_, _ = utils.Run(cmd)
 
+		// Collect controller logs and pod descriptions to ARTIFACTS before teardown.
+		if artifactsDir := os.Getenv("ARTIFACTS"); artifactsDir != "" {
+			By("collecting controller-manager logs to ARTIFACTS")
+			if logsOut, err := utils.Run(exec.Command("kubectl", "logs", "-n", namespace,
+				"deployment/nrr-controller-manager", "--all-containers")); err == nil {
+				_ = os.WriteFile(filepath.Join(artifactsDir, "nrr-controller-manager.log"),
```
vitorfloriano: Couldn't we handle the error here instead of _? Or at least log it if the write failed.

hack/e2e-test.sh
```diff
@@ -0,0 +1,80 @@
...
+
+if [[ "$E2E_KIND_VERSION" =~ ^v[0-9]+\.[0-9]+$ ]]; then
+    K8S_VERSION=$(curl -sf \
+        "https://api.github.com/repos/kubernetes/kubernetes/git/matching-refs/tags/${E2E_KIND_VERSION}." \
+        | grep -oP 'v[0-9]+\.[0-9]+\.[0-9]+(?=")' \
+        | sort -V | tail -1)
```
vitorfloriano: Shouldn't we be a bit more defensive about curl-ing the GitHub API, maybe adding a timeout or a retry? Or would that be overkill in this case?

vitorfloriano (commented):
Overall, looks good! Just a few inquiries.

/lgtm

test/e2e/e2e_test.go
Priyankasaggu11929: addressed.

hack/e2e-test.sh
Priyankasaggu11929: added a retry and max-time to retry.

hack/e2e-test.sh
Priyankasaggu11929: this is a fall back value (to latest k8s version)

i think we can keep it as it is (unless you mean something else?)

Priyankasaggu11929:
Thanks for the review @vitorfloriano. Addressed all comments and left explanation for one.

hack/e2e-test.sh
vitorfloriano: I mean, the latest version is always a moving target and having any value hardcoded in here is something more we need to remember to update when we change to the next version.

But since it's a fallback value, that's fine. We can try to automate the version bump later with the release automation, check every places where k8s version is hardcoded and update.

vitorfloriano (approved):
Looks good! Thanks! 

/lgtm

ajaysundark:
@Priyankasaggu11929 Thanks! when you find time can you also get to our TEST/README files if they need an update for this?

/approve

Priyankasaggu11929:
> @Priyankasaggu11929 Thanks! when you find time can you also get to our TEST/README files if they need an update for this?
> 
> /approve

kept a TODO for myself. (will address it in july 1st week)

PR #185, opened 2026-04-15
✨ Add metrics and scale test

Karthik-K-N:
## Description

This PR adds metrics and scale tests using KWOK

## Type of Change
/kind feature

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

## Does this PR introduce a user-facing change?
Adds 

```release-note
Add metrics and scale test

```
Doc #(issue)

Screenshot 2026-04-15 at 11 51 46
Screenshot 2026-04-15 at 11 51 52
Screenshot 2026-04-15 at 11 51 58

PR #271, opened 2026-06-02
✨ Add new metrics and testing documentation

Karthik-K-N:
## Description

This PR adds few metrics into the controller.

Contains only metrics part from PR: https://github.com/kubernetes-sigs/node-readiness-controller/pull/185

## Type of Change

/kind feature

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Add new metrics and testing documentation
```
Doc #(issue)

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -161,6 +161,8 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			"rule", rule.Name,
 			"resourceVersion", rule.ResourceVersion)
 
+		var successfullyPatchedRule *readinessv1alpha1.NodeReadinessRule
+
 		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			latestRule := &readinessv1alpha1.NodeReadinessRule{}
 			if err := r.Get(ctx, client.ObjectKey{Name: rule.Name}, latestRule); err != nil {
@@ -207,7 +209,12 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			}
 			latestRule.Status.FailedNodes = updatedFailedNodes
 
-			return r.Status().Patch(ctx, latestRule, patch)
+			if err := r.Status().Patch(ctx, latestRule, patch); err != nil {
+				return err
+			}
+
+			successfullyPatchedRule = latestRule
+			return nil
 		})
 
 		if err != nil {
@@ -222,6 +229,10 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 				"node", node.Name,
 				"rule", rule.Name,
 				"newResourceVersion", rule.ResourceVersion)
+
+			if successfullyPatchedRule != nil {
+				r.SyncNodeStateMetrics(ctx, successfullyPatchedRule)
+			}
 		}
 	}
 
@@ -406,4 +417,29 @@ func (r *RuleReadinessController) recordNodeFailure(
 	})
 
 	rule.Status.FailedNodes = failedNodes
+}
+
+// SyncNodeStateMetrics synchronizes the NodesByState Prometheus metrics with the current rule status.
+func (r *RuleReadinessController) SyncNodeStateMetrics(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule) {
+	var ready, notReady, bootstrapping float64
+
+	for _, eval := range rule.Status.NodeEvaluations {
+		if eval.TaintStatus == readinessv1alpha1.TaintStatusAbsent {
+			ready++
+		} else {
+			if rule.Spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
+				if !r.isBootstrapCompleted(ctx, eval.NodeName, rule.Name) {
+					bootstrapping++
+				} else {
+					notReady++
+				}
+			} else {
+				notReady++
+			}
+		}
+	}
+
+	metrics.NodesByState.WithLabelValues(rule.Name, "ready").Set(ready)
+	metrics.NodesByState.WithLabelValues(rule.Name, "not_ready").Set(notReady)
+	metrics.NodesByState.WithLabelValues(rule.Name, "bootstrapping").Set(bootstrapping)
 }
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -230,8 +230,10 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 				"rule", rule.Name,
 				"newResourceVersion", rule.ResourceVersion)
 
-			if successfullyPatchedRule != nil {
-				r.SyncNodeStateMetrics(ctx, successfullyPatchedRule)
+			if r.EnableNodeStateMetrics {
+				if successfullyPatchedRule != nil {
+					r.SyncNodeStateMetrics(ctx, successfullyPatchedRule)
+				}
 			}
 		}
 	}
@@ -427,12 +429,10 @@ func (r *RuleReadinessController) SyncNodeStateMetrics(ctx context.Context, rule
 		if eval.TaintStatus == readinessv1alpha1.TaintStatusAbsent {
 			ready++
 		} else {
+			// The taint is still present.
 			if rule.Spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
-				if !r.isBootstrapCompleted(ctx, eval.NodeName, rule.Name) {
-					bootstrapping++
-				} else {
-					notReady++
-				}
+				// In BootstrapOnly mode, if the taint is present, it is still bootstrapping.
+				bootstrapping++
 			} else {
 				notReady++
 			}
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -162,6 +162,10 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{RequeueAfter: time.Minute}, err
 	}
 
+	// Update top-level rule metrics.
+	metrics.RuleLastReconciliationTime.WithLabelValues(rule.Name).Set(float64(time.Now().Unix()))
+	r.Controller.SyncNodeStateMetrics(ctx, rule)
+
 	return ctrl.Result{}, nil
 }
 
@@ -192,6 +196,13 @@ func (r *RuleReconciler) reconcileDelete(ctx context.Context, rule *readinessv1a
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+
+	// Clean up metrics for deleted rule to prevent memory leaks.
+	metrics.RuleLastReconciliationTime.DeleteLabelValues(rule.Name)
+	metrics.NodesByState.DeleteLabelValues(rule.Name, "ready")
+	metrics.NodesByState.DeleteLabelValues(rule.Name, "not_ready")
+	metrics.NodesByState.DeleteLabelValues(rule.Name, "bootstrapping")
+
 	return ctrl.Result{}, nil
 }
 
@@ -303,6 +314,7 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule 
 
 		if !satisfied {
 			allConditionsSatisfied = false
+			metrics.ConditionEvaluationFailures.WithLabelValues(rule.Name, condReq.Type).Inc()
 		}
 
 		conditionResults = append(conditionResults, readinessv1alpha1.ConditionEvaluationResult{
@@ -325,6 +337,25 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule 
 
 	isFirstEvaluation := r.getPreviousNodeEvaluation(rule, node.Name) == nil
 
+	recordLatency := func(operation string) {
+		// We intentionally isolate the most recent transition time among all required conditions.
+		// Since the controller must wait for the combined state of all conditions to change
+		// before taking action, the condition that changed most recently is the "trigger" event.
+		// Older transition times are ignored because the controller was intentionally waiting.
+		var latestTransition metav1.Time
+		for _, req := range rule.Spec.Conditions {
+			for _, cond := range node.Status.Conditions {
+				if string(cond.Type) == req.Type && cond.LastTransitionTime.After(latestTransition.Time) {
+					latestTransition = cond.LastTransitionTime
+				}
+			}
+		}
+		if !latestTransition.IsZero() {
+			latency := time.Since(latestTransition.Time).Seconds()
+			metrics.ReconciliationLatency.WithLabelValues(rule.Name, operation).Observe(latency)
+		}
+	}
+
 	var err error
 
 	switch {
@@ -335,11 +366,25 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule 
 			metrics.Failures.WithLabelValues(rule.Name, "RemoveTaintError").Inc()
 			return fmt.Errorf("failed to remove taint: %w", err)
 		}
+
+		// Record taint removal latency and taint operation counter.
 		metrics.TaintOperations.WithLabelValues(rule.Name, "remove").Inc()
+		recordLatency("remove_taint")
 
 		// Mark bootstrap completed if bootstrap-only mode
 		if rule.Spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
 			r.markBootstrapCompleted(ctx, node.Name, rule.Name)
+
+			// Only record the bootstrap duration if the node was created AFTER the rule.
+			// This prevents legacy nodes from poisoning the histogram with massive outliers.
+			if !node.CreationTimestamp.Time.Before(rule.CreationTimestamp.Time) {
+				duration := time.Since(node.CreationTimestamp.Time).Seconds()
+				metrics.BootstrapDuration.WithLabelValues(rule.Name).Observe(duration)
+			} else {
+				log.V(4).Info("Skipping bootstrap duration metric for legacy node",
+					"node", node.Name,
+					"rule", rule.Name)
+			}
 		}
 
 	case !shouldRemoveTaint && !currentlyHasTaint:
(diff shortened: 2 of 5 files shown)


internal/metrics/metrics.go
```diff
@@ -65,6 +65,58 @@ var (
...
+		prometheus.GaugeOpts{
+			Name: "node_readiness_nodes_by_state",
+			Help: "Number of nodes in each readiness state per rule",
+		},
+		[]string{"rule", "state"}, // state: ready, not_ready, bootstrapping
+	)
+
+	// ConditionEvaluationFailures tracks which specific checks within a rule are failing.
+	// This allows operators to pinpoint the exact infrastructure component blocking node readiness.
+	ConditionEvaluationFailures = prometheus.NewCounterVec(
```
AvineshTripathi, line 103: We are only incrementing this counter. Let's say the rule condition failed and later succeeded. In this case we are not dec the counter. Is that expected?

internal/metrics/metrics.go
```diff
@@ -65,6 +65,58 @@ var (
...
+			Name:    "node_readiness_bootstrap_duration_seconds",
+			Help:    "Time from node creation to bootstrap completion (taint removal) for bootstrap-only rules",
+			Buckets: []float64{1, 5, 10, 30, 60, 120, 300, 600, 1200}, // 1s to 20min
+		},
+		[]string{"rule"},
+	)
+
+	// ReconciliationLatency tracks end-to-end latency from condition change to taint operation.
+	// This measures how quickly the controller responds to node condition changes.
+	ReconciliationLatency = prometheus.NewHistogramVec(
```
AvineshTripathi, line 82: looking at this, I was  wondering if we should club this and   https://github.com/kubernetes-sigs/node-readiness-controller/pull/271/changes#diff-6e56d7c057f6a96124b78f2b669dd7dface2718c304aece205017d114b3a7088R43?? What do you think? 

I feel both share the same purpose

internal/controller/nodereadinessrule_controller.go
```diff
@@ -335,11 +362,18 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
...
+		// Record taint removal latency and taint operation counter.
 		metrics.TaintOperations.WithLabelValues(rule.Name, "remove").Inc()
+		recordLatency("remove_taint")
 
 		// Mark bootstrap completed if bootstrap-only mode
 		if rule.Spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
 			r.markBootstrapCompleted(ctx, node.Name, rule.Name)
+
+			// Record end-to-end Bootstrap Duration.
+			duration := time.Since(node.CreationTimestamp.Time).Seconds()
```
AvineshTripathi: thinking sillly maybe: there can be instances where the node was created a long time back and the controller was added later. Since we do not store that bootstrapping is started, this would potentially show a time > the age of the controller.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -325,6 +337,21 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
 
 	isFirstEvaluation := r.getPreviousNodeEvaluation(rule, node.Name) == nil
 
+	recordLatency := func(operation string) {
+		var latestTransition metav1.Time
```
AvineshTripathi: While this is fine because we want to calculate a single latency for the entire rule, the current nested loop structure is a bit misleading. It iterates through every condition type but `latestTransition` keeps getting overwritten. As a result, it only tracks whichever condition changed most recently, and the older condition's latencies are completely ignored.

AvineshTripathi (commented):
Thanks for the PR @Karthik-K-N. I have put some comments! PTAL

internal/controller/nodereadinessrule_controller.go
Karthik-K-N: This is not silly at all, Really valid case, thanks for bringing it up, I just updated the PR to record the metric only if the node is created post the rule creation. we will ignore the legacy nodes.
What do you think about this approach.

internal/metrics/metrics.go
Karthik-K-N: Yes, this is completely expected. Because `ConditionEvaluationFailures` is a Prometheus Counter, it must be monotonically increasing by design. If we were to decrement it, it would break PromQL functions like rate() or increase() on our dashboards. To track the actual current live state (going up and down), we use the `NodesByState` gauge instead.

internal/controller/nodereadinessrule_controller.go
Karthik-K-N: Yes, You are right that it ignores the older condition's latencies, but that is actually intentional. Since the controller acts based on the combined state of all required conditions, it is blocked from acting until the final required condition transitions. Say for example we want to have StorageReady and NetworkReady condition to be met for the Node to be marked as ready.

The condition that transitioned most recently is the actual 'trigger' event that unblocks the controller. We measure our reaction latency against this final trigger, because the older conditions were essentially just sitting in a 'waiting' state. I have added a comment above the loop to make this intent explicit so it doesn't look like a bug for the readers.

internal/metrics/metrics.go
Karthik-K-N: which one are you referring to? I am not able to trace.

AvineshTripathi: apologies, here is the link https://github.com/Karthik-K-N/node-readiness-controller/blob/398a91c694977524361d55de0d9997f1e1c5275b/internal/metrics/metrics.go#L43

internal/controller/nodereadinessrule_controller.go
AvineshTripathi: I dont see anything wrong in the approach. Thanks for fixing it.

internal/controller/nodereadinessrule_controller.go
AvineshTripathi: thanks for explaining!

internal/metrics/metrics.go
Karthik-K-N: No Issues thanks.

`EvaluationDuration` is an internal profiling metric, it tells us how fast our Go code executes a single rule evaluation (usually in milliseconds). `ReconciliationLatency` is our external Service Level Indicator (SLI). It measures the total end-to-end reaction time from the moment the node changed state in the real world to the moment we successfully patched the taint (which includes API server lag, rate limiting, and queue wait times, usually in seconds). 
Keeping them separate allows us to instantly know if a slowdown is caused by inefficient code vs. Kubernetes API queueing.

ajaysundark:
cc @rawadhossain

rawadhossain:
Thanks for looping me into this PR (somehow missed it earlier😅). I went through the changes, and this aligns quite well with the Headlamp project.

The separation between `EvaluationDuration` and `ReconciliationLatency` makes sense, and the additional rule-level metrics would be helpful for the Grafana dashboards and Headlamp views we'll be building.

Also nice to see `nodes_by_state` being updated from both reconciler paths so the state remains accurate regardless of where the reconciliation is triggered.

Looking forward to building on top of this once it lands.

cc @ajaysundark @Karthik-K-N

ajaysundark:
In a high level, metric shapes look good to me. But I'm not sure if updating state / metric per event is scalable for the controller. We may need to take a second look at the compute cost for these

Karthik-K-N:
May be can we get this in to unblock or not to overlap with other work and based on the scale test results we can revisit and update the metrics?

ajaysundark:
> May be can we get this in to unblock or not to overlap with other work and based on the scale test results we can revisit and update the metrics?

Could we create separate feature branches for the LFX projects? it'd help with the velocity of the projects as well.

internal/controller/node_controller.go
```diff
@@ -407,3 +418,28 @@ func (r *RuleReadinessController) recordNodeFailure(
 
 	rule.Status.FailedNodes = failedNodes
 }
+
+// SyncNodeStateMetrics synchronizes the NodesByState Prometheus metrics with the current rule status.
+func (r *RuleReadinessController) SyncNodeStateMetrics(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule) {
```
ajaysundark, line 425: this loops every node patch x every rule reconciliation. This O(NxR) could quickly add to perf cost as we scale to 1000s of nodes.  I'm seeing the value of this metric, but IMHO, it feels very heavy for the 'rule healthiness histogram'. 

Can we keep the per-node-metrics to behind a flag for now? We dont have feature-gate mechanism in NRC, which would be ideal for these changes.

ajaysundark:
> > May be can we get this in to unblock or not to overlap with other work and based on the scale test results we can revisit and update the metrics?
> 
> Could we create separate feature branches for the LFX projects? it'd help with the velocity of the projects as well.

Thinking more about it, feature-branch is not the best option as it brings 'merge cost' to the contributors for every feature. Feature-gates like k/k is light weight. It maybe portable here. but we could just use even a recognizable flag that we can deprecate later.

internal/controller/node_controller.go
```diff
@@ -407,3 +418,28 @@ func (r *RuleReadinessController) recordNodeFailure(
...
+	var ready, notReady, bootstrapping float64
+
+	for _, eval := range rule.Status.NodeEvaluations {
+		if eval.TaintStatus == readinessv1alpha1.TaintStatusAbsent {
+			ready++
+		} else {
+			if rule.Spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
+				if !r.isBootstrapCompleted(ctx, eval.NodeName, rule.Name) {
```
ajaysundark: This actually performs a `Get` every call.

https://github.com/kubernetes-sigs/node-readiness-controller/blob/f6adb3ab70775137d64a0f7e7976d088144354d4/internal/controller/node_controller.go#L331

can we use `TaintStatus == TaintStatusPresent ` to determine `bootstrapping` directly?

internal/controller/nodereadinessrule_controller.go
```diff
@@ -325,6 +337,25 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
...
+		// Older transition times are ignored because the controller was intentionally waiting.
+		var latestTransition metav1.Time
+		for _, req := range rule.Spec.Conditions {
+			for _, cond := range node.Status.Conditions {
+				if string(cond.Type) == req.Type && cond.LastTransitionTime.After(latestTransition.Time) {
+					latestTransition = cond.LastTransitionTime
+				}
+			}
+		}
+		if !latestTransition.IsZero() {
```
ajaysundark, line 368: Should we check for latency >= 0 instead?

internal/controller/nodereadinessrule_controller.go
```diff
@@ -335,11 +366,25 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
...
+		recordLatency("remove_taint")
 
 		// Mark bootstrap completed if bootstrap-only mode
 		if rule.Spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
 			r.markBootstrapCompleted(ctx, node.Name, rule.Name)
+
+			// Only record the bootstrap duration if the node was created AFTER the rule.
+			// This prevents legacy nodes from poisoning the histogram with massive outliers.
+			if !node.CreationTimestamp.Time.Before(rule.CreationTimestamp.Time) {
+				duration := time.Since(node.CreationTimestamp.Time).Seconds()
```
ajaysundark: I think this would not be ideal for bootstrap_duration, as it includes everything between API to controller. Controller metrics should preferably be reported only on its purview of events.

I'm not an expert in this area, but would publishing timestamps vs durations, just like similar to 'CreationTimestamps' in general be preferred?

ajaysundark: I think it also will have clock skew between two nodes, so duration between two machines may not even be accurate.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -192,6 +196,13 @@ func (r *RuleReconciler) reconcileDelete(ctx context.Context, rule *readinessv1a
 	if err != nil {
 		return ctrl.Result{}, err
 	}
+
+	// Clean up metrics for deleted rule to prevent memory leaks.
```
ajaysundark: curious, why does this clean up only two metrics, is the concern not present for other metrics as well?

Karthik-K-N: I originally only cleaned up the Gauges because failing to clean those up leaves stale data, Now I have updated it.

internal/controller/nodereadinessrule_controller.go
Karthik-K-N: Your are right that using time.Since() mixes the controller's local clock with the API server's clock, which guarantees skew.

Regarding exporting timestamps vs. durations: publishing timestamps would force us to add a node_name label to the metric so we can query them individually. At thousands of nodes, that causes a massive Prometheus cardinality explosion. By keeping it as an aggregate Histogram, the metric footprint stays totally flat.
To solve your clock skew concern I have updated the code.

internal/controller/node_controller.go
Karthik-K-N: Nice, Catch updated it thanks.

internal/controller/node_controller.go
Karthik-K-N: Added this behind a flag.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -192,6 +201,22 @@ func (r *RuleReconciler) reconcileDelete(ctx context.Context, rule *readinessv1a
...
+	metrics.RuleLastReconciliationTime.DeleteLabelValues(rule.Name)
+	metrics.BootstrapCompleted.DeleteLabelValues(rule.Name)
+
+	// For multi-label metrics, use DeletePartialMatch to wipe all combinations
+	metrics.NodesByState.DeletePartialMatch(ruleLabel)
+	metrics.Failures.DeletePartialMatch(ruleLabel)
+	metrics.ConditionEvaluationFailures.DeletePartialMatch(ruleLabel)
+	metrics.TaintOperations.DeletePartialMatch(ruleLabel)
+	metrics.ReconciliationLatency.DeletePartialMatch(ruleLabel)
+	metrics.BootstrapDuration.DeletePartialMatch(ruleLabel)
```
ajaysundark: You are defining the `BootstrapDuration` as single-label metric here: internal/metrics/metrics.go. Does this need to go up instead?

internal/controller/nodereadinessrule_controller.go
```diff
@@ -335,11 +388,40 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
...
 
 		// Mark bootstrap completed if bootstrap-only mode
 		if rule.Spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
 			r.markBootstrapCompleted(ctx, node.Name, rule.Name)
+
+			// Only record the bootstrap duration if the node was created AFTER the rule.
+			// This prevents legacy nodes from poisoning the histogram with massive outliers.
+			if !node.CreationTimestamp.Time.Before(rule.CreationTimestamp.Time) {
+				duration := time.Since(node.CreationTimestamp.Time).Seconds()
+				metrics.BootstrapDuration.WithLabelValues(rule.Name).Observe(duration)
```
ajaysundark, line 404: is this a missing cleanup?

internal/controller/nodereadinessrule_controller.go
Karthik-K-N: Yeah, I have updated it.

config/prometheus/manager_prometheus_metrics.yaml
```diff
@@ -2,3 +2,6 @@
 - op: add
   path: /spec/template/spec/containers/0/args/-
   value: --metrics-bind-address=:8080
+- op: add
+  path: /spec/template/spec/containers/0/args/-
+  value: --enable-node-state-metrics=true
```
ajaysundark: Should this also be disabled by default here as well until we evaluate this for scale?

internal/controller/node_controller.go
```diff
@@ -407,3 +420,26 @@ func (r *RuleReadinessController) recordNodeFailure(
 
 	rule.Status.FailedNodes = failedNodes
 }
+
+// SyncNodeStateMetrics synchronizes the NodesByState Prometheus metrics with the current rule status.
+func (r *RuleReadinessController) SyncNodeStateMetrics(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule) {
+	var ready, notReady, bootstrapping float64
+
+	for _, eval := range rule.Status.NodeEvaluations {
```
ajaysundark, line 428: This is behind a flag so this is okay to merge this as-is. But wanted to note this for discussion, I expect this would also add to the heapsize for in-memory processing with a large array. In the long term we are reevaluating to drop `Status.NodeEvaluations`, so this will be of lesser concern in the future.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -335,11 +388,40 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
...
 		metrics.TaintOperations.WithLabelValues(rule.Name, "remove").Inc()
+		recordLatency("remove_taint")
 
 		// Mark bootstrap completed if bootstrap-only mode
 		if rule.Spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
 			r.markBootstrapCompleted(ctx, node.Name, rule.Name)
+
+			// Only record the bootstrap duration if the node was created AFTER the rule.
+			// This prevents legacy nodes from poisoning the histogram with massive outliers.
+			if !node.CreationTimestamp.Time.Before(rule.CreationTimestamp.Time) {
```
ajaysundark, line 402: I'm still not convinced whether this should be emitted metric or can be computed on the client side.

As we discussed in the slack, this could give a false impression that the controller took this time to unblock the node, but in reality there could be a lot of things in between like other node-controllers or components etc. also the time skew between multiple nodes would make this an inaccurate measurement.

But since this is still alpha, we have room to reconsider this metric shape in the future releases.

ajaysundark (commented):
Thanks @Karthik-K-N . My only blocking concern on this is that the node-state metrics could be disabled also from the yaml / charts. Plmk wdyt.

/lgtm

config/prometheus/manager_prometheus_metrics.yaml
Karthik-K-N: Yes, Missed it , Removed this block now. Thanks

ajaysundark (approved):
/lgtm

Thanks @Karthik-K-N !

PR #283, opened 2026-06-26
feat: add optional DefaultStatus field to ConditionRequirement for missing node conditions

### LINKED ISSUE #282: [FEATURE] support 'defaultStatus' for ConditionRequirement

This is a new feature ask to enable NRC to have cleaner integration with NPD / problem states.

This feature ask is to add an optional per-condition field `defaultStatus` in `ConditionRequirement` that controls how an absent node condition is evaluated.

This enables a readiness-rule to work as a problem gate (default-allow) during a node-bootstrap. This enables NRC to work for maintenance/readiness situations with continuous-mode rules that taint a node only when a condition explicitly signals a problem (eg: NPD detection, or hold scheduling for a configuration update on the node).

Requires an additional step that Node always starts with a `MaintenanceRequired=False` condition, and flipped to `MaintenanceRequired=True` by external controller / NPD when required.

Above adds an inherent (race and) delay at the bootstrap time to support such a scenario, requiring a maintenance controller to monitor all create events and patching a no-maintenance event as "missing" condition is evaluated as 'unsatisfied' by the rule.

Adding an optional `defaultStatus` status with `Unknown` as default value will help this use-case while also maintaining existing behavior

AnuragThePathak:
## Description

Issue #282 already describes why need the DefaultStatus field to act as a problem-gate during node bootstrap. This PR makes the following changes to implement that -

- Introduce a new optional `DefaultStatus` field in `ConditionRequirement`. Type `Corev1.ConditionStatus`.
- Introduce a `GetDefaultStatus()` accessor for safe access of that field. A kubebuilder schema default is not used due to project policy, so defaulting to `Unknown` happens at read-time via this method.
- Update the evaluator to use `defaultStatus` when matching a condition's effective status against `requiredStatus` for absent conditions.
- Report the actual observed condition status (`Unknown`) in `ConditionEvaluationResult.currentStatus` rather than the effective (default-applied) value, preserving the "observed" semantic of the field.
- Populate `ConditionEvaluationResult.defaultStatus` to reflect what default was configured in the spec, giving full visibility into evaluation decisions.
- Aligned the `DryRunResults.riskyOperations` logic to strictly count physically missing conditions (adhering to the API schema definition), rather than conflating them with explicitly reported Unknown statuses.
- Regenerate CRD manifests with `make manifests` to include the new `defaultStatus` property.

## Related Issue

Fixes #282 

## Type of Change

/kind feature
/kind api-change

## Testing

**Unit/Integration Tests**:
- Added test case verifying `getConditionStatus` resolves to the configured default status when a condition is missing from a node.
- Added integration test verifying that a node with an absent condition satisfies a rule when the condition's `defaultStatus` matches `requiredStatus` (problem-gate scenario).
- Added dry-run integration test verifying that missing conditions on a node trigger `riskyOperations` increment, even if the condition is satisfied via `defaultStatus`, and correctly flags `taintsToRemove` while keeping `taintsToAdd` at `0`.

**Validation**:
- Verified code compiles, lints, and all tests pass with `make test` and `make lint`.
- Ran E2E suite (`make test-e2e`) locally; all 14 tests passed successfully, ensuring no regressions.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Added an optional `defaultStatus` field to `ConditionRequirement` (`spec.conditions[]`) in the `NodeReadinessRule` API. This field controls how an absent node condition is evaluated. When omitted, the effective default is `Unknown`, preserving existing behaviour. This enables NRC to function as a problem gate (default-allow) during node bootstrap when integrating with NPD or similar external condition controllers.
```

## Documentation
The following documentation files were updated:
- **`docs/book/src/introduction.md`**: Added `defaultStatus` as a Key Feature.
- **`docs/book/src/user-guide/getting-started.md`**: Added `defaultStatus` as a third field under "Define Readiness Conditions" with a link to Concepts for implications.
- **`docs/book/src/user-guide/concepts.md`**: Added a new `### Default Condition Status (defaultStatus)` subsection covering:
  - What the field does and what it defaults to when omitted
  - Problem-gate (default-allow) use case for NPD integration
  - `[!WARNING]` block explicitly documenting that a satisfying `defaultStatus` must **not** be used with `bootstrap-only` rules, as it permanently bypasses the gate
- **`docs/book/src/reference/api-spec.md`**: Added `defaultStatus` rows to both the `ConditionRequirement` and `ConditionEvaluationResult` tables, with the `Default: Unknown` column populated for `ConditionRequirement` (spec field), and left blank for `ConditionEvaluationResult` (status field).

### DIFF
--- a/api/v1alpha1/nodereadinessrule_types.go
+++ b/api/v1alpha1/nodereadinessrule_types.go
@@ -128,13 +128,13 @@ type ConditionRequirement struct {
 	// defaultStatus is the status a condition is evaluated to if the condition
 	// is not found in a node.
 	//
-	// Accepted values True, False, Unknown. It is optional.
-	// If no value is provided, then Unknown becomes the default value.
+	// Accepted values are True, False, Unknown. It is optional.
+	// When omitted, the effective default is Unknown, applied transparently by
+	// the controller at evaluation time.
 	//
 	// +optional
-	// +kubebuilder:default="Unknown"
 	// +kubebuilder:validation:Enum=True;False;Unknown
-	DefaultStatus corev1.ConditionStatus `json:"defaultStatus,omitempty"`
+	DefaultStatus corev1.ConditionStatus `json:"defaultStatus,omitempty"` // Use GetDefaultStatus() for safe access; field may be empty even when a default applies.
 }
 
 // NodeReadinessRuleStatus defines the observed state of NodeReadinessRule.
@@ -349,6 +349,21 @@ type NodeReadinessRuleList struct {
 	metav1.ListMeta `json:"metadata,omitempty"`
 	// items is the list of NodeReadinessRule.
 	Items []NodeReadinessRule `json:"items"`
+}
+
+// GetDefaultStatus returns the effective default status for a condition that is
+// not found on a node. If the field is unset (empty string), it falls back to
+// corev1.ConditionUnknown.
+//
+// Always use this method instead of reading DefaultStatus directly. The field
+// is intentionally left without an OpenAPI schema default (kubebuilder:default
+// is forbidden by project policy) and the Spec is immutable, so defaulting
+// must happen at read time via this accessor.
+func (c *ConditionRequirement) GetDefaultStatus() corev1.ConditionStatus {
+	if c.DefaultStatus == "" {
+		return corev1.ConditionUnknown
+	}
+	return c.DefaultStatus
 }
 
 func init() {
--- a/api/v1alpha1/nodereadinessrule_types.go
+++ b/api/v1alpha1/nodereadinessrule_types.go
@@ -107,7 +107,8 @@ type NodeReadinessRuleSpec struct {
 }
 
 // ConditionRequirement defines a specific Node condition and the status value
-// required to trigger the controller's action.
+// required to trigger the controller's action. It also contains an optional
+// default status value.
 type ConditionRequirement struct {
 	// type of Node condition
 	//
@@ -123,6 +124,17 @@ type ConditionRequirement struct {
 	// +required
 	// +kubebuilder:validation:Enum=True;False;Unknown
 	RequiredStatus corev1.ConditionStatus `json:"requiredStatus,omitempty"`
+
+	// defaultStatus is the status a condition is evaluated to if the condition
+	// is not found in a node.
+	//
+	// Accepted values True, False, Unknown. It is optional.
+	// If no value is provided, then Unknown becomes the default value.
+	//
+	// +optional
+	// +kubebuilder:default="Unknown"
+	// +kubebuilder:validation:Enum=True;False;Unknown
+	DefaultStatus corev1.ConditionStatus `json:"defaultStatus,omitempty"`
 }
 
 // NodeReadinessRuleStatus defines the observed state of NodeReadinessRule.
--- a/api/v1alpha1/nodereadinessrule_types.go
+++ b/api/v1alpha1/nodereadinessrule_types.go
@@ -267,6 +267,14 @@ type ConditionEvaluationResult struct {
 	// +required
 	// +kubebuilder:validation:Enum=True;False;Unknown
 	RequiredStatus corev1.ConditionStatus `json:"requiredStatus,omitempty"`
+
+	// defaultStatus is the status a condition is evaluated to if the condition
+	// is not found in a node. Reflects the defaultStatus configured in the rule
+	// spec.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum=True;False;Unknown
+	DefaultStatus corev1.ConditionStatus `json:"defaultStatus,omitempty"`
 }
 
 // DryRunResults provides a summary of the actions the controller would perform if DryRun mode is enabled.
--- a/config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml
+++ b/config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml
@@ -70,8 +70,22 @@ spec:
                 items:
                   description: |-
                     ConditionRequirement defines a specific Node condition and the status value
-                    required to trigger the controller's action.
+                    required to trigger the controller's action. It also contains an optional
+                    default status value.
                   properties:
+                    defaultStatus:
+                      description: |-
+                        defaultStatus is the status a condition is evaluated to if the condition
+                        is not found in a node.
+
+                        Accepted values are True, False, Unknown. It is optional.
+                        When omitted, the effective default is Unknown, applied transparently by
+                        the controller at evaluation time.
+                      enum:
+                      - "True"
+                      - "False"
+                      - Unknown
+                      type: string
                     requiredStatus:
                       description: requiredStatus is status of the condition, one
                         of True, False, Unknown.
--- a/config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml
+++ b/config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml
@@ -377,6 +377,16 @@ spec:
                             - "False"
                             - Unknown
                             type: string
+                          defaultStatus:
+                            description: |-
+                              defaultStatus is the status a condition is evaluated to if the condition
+                              is not found in a node. Reflects the defaultStatus configured in the rule
+                              spec.
+                            enum:
+                            - "True"
+                            - "False"
+                            - Unknown
+                            type: string
                           requiredStatus:
                             description: requiredStatus is the status value defined
                               in the rule that must be matched, one of True, False,
--- a/docs/book/src/user-guide/concepts.md
+++ b/docs/book/src/user-guide/concepts.md
@@ -24,6 +24,29 @@ Typical taint keys look like:
 
 The segment after `readiness.k8s.io/` should describe the dependency or subsystem whose readiness is being guarded (for example, a CNI plugin, storage backend, or security agent). Treat this domain as reserved for the controller and closely related components, and avoid reusing it for unrelated taints.
 
+### Default Condition Status (`defaultStatus`)
+
+When defining readiness conditions, you can configure an optional `defaultStatus` field (one of `True`, `False`, or `Unknown`) under `spec.conditions[]`. This field determines how a condition is evaluated when it is completely absent from the Node's status.
+
+If `defaultStatus` is omitted, the controller assumes a fallback status of `Unknown`.
+
+#### Use Case: Problem Gating (Default-Allow)
+By default, the controller operates in a "default-deny" fashion. If a guarded condition is missing from a new node, it resolves to `Unknown`, which typically does not satisfy `requiredStatus: True` or `requiredStatus: False`, causing the node to be tainted.
+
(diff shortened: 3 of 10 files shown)


ajaysundark:
@AnuragThePathak Thanks for the PR. Is this ready for review?

AnuragThePathak:
Yes @ajaysundark you can review the code changes. There are no expected changes unless suggested during review.

In the meantime, I'll keep working on updating the tests which should affect the implementation review.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -846,15 +846,15 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 			}
 
 			// Test condition exists and matches
-			status := readinessController.getConditionStatus(node, "Ready")
+			status, _ := readinessController.getConditionStatus(node, "Ready", "Unknown")
```
ajaysundark: prefer typed constants like `corev1.ConditionUnknown` instead of string literals.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -846,15 +846,15 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 			}
 
 			// Test condition exists and matches
```
ajaysundark, line 1039: Could you also add a quick test case to demonstrate a custom default is actually respected when the condition is missing?

ajaysundark:
The direction of the PR looks good to me, publish the PR so the CI kicks in.

We should also document the implications of this feature. For example, defaultStatus doesnt fit for bootstrap-only enforcement mode, where 'unknown' status fit as a better default.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -298,21 +298,33 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
...
 			Type:           condReq.Type,
-			CurrentStatus:  currentStatus,
+			CurrentStatus:  observedStatus,
 			RequiredStatus: condReq.RequiredStatus,
+			DefaultStatus:  condReq.GetDefaultStatus(),
 		})
 
 		log.V(1).Info("Condition evaluation", "node", node.Name, "rule", rule.Name,
-			"conditionType", condReq.Type, "current", currentStatus, "required", condReq.RequiredStatus,
+			"conditionType", condReq.Type, "current", observedStatus, "required", condReq.RequiredStatus,
```
ajaysundark:
```suggestion
			"conditionType", condReq.Type, "observed", observedStatus, "effective", effectiveStatus, "required", condReq.RequiredStatus,
```

config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml
```diff
@@ -70,8 +70,22 @@ spec:
...
+                        defaultStatus is the status a condition is evaluated to if the condition
+                        is not found in a node.
+
+                        Accepted values are True, False, Unknown. It is optional.
+                        When omitted, the effective default is Unknown, applied transparently by
+                        the controller at evaluation time.
+                      enum:
+                      - "True"
+                      - "False"
+                      - Unknown
```
ajaysundark, line 87: nit: why only Unknown is left unquoted?

AnuragThePathak:
@ajaysundark It is good from my side now. By the way you may benefit reviewing by commit, I have created different commit for separate concerns/categories.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -339,6 +359,21 @@ type NodeReadinessRuleList struct {
 	Items []NodeReadinessRule `json:"items"`
 }
 
+// GetDefaultStatus returns the effective default status for a condition that is
+// not found on a node. If the field is unset (empty string), it falls back to
+// corev1.ConditionUnknown.
+//
+// Always use this method instead of reading DefaultStatus directly. The field
```
ajaysundark, line 366: This could also go as a field comment

docs/book/src/user-guide/concepts.md
```diff
@@ -24,6 +24,29 @@ Typical taint keys look like:
...
+
+#### Important Implication: Enforcement Modes
+The choice of `defaultStatus` has critical interaction with the rule's `enforcementMode`:
+
+> [!WARNING]
+> **Do not use a satisfying `defaultStatus` with `bootstrap-only` rules.**
+> 
+> If a rule is `bootstrap-only` and the absent condition evaluates to a status that matches `requiredStatus` (e.g. `requiredStatus: False` and `defaultStatus: False`), the controller will immediately mark the bootstrap as completed (`bootstrap-completed-<ruleName>=true`) and remove the taint. 
+> 
+> Since `bootstrap-only` rules are never re-evaluated once marked complete, the gate is permanently bypassed, even if a reporter eventually updates the condition to an unsatisfied status. For `bootstrap-only` rules, leaving the default status as `Unknown` is almost always the correct approach to guarantee the controller waits for the actual condition to be reported.
```
ajaysundark, line 48: Thanks for documenting this! We could check whether it is a good idea to validate this explicitly in webhook. I was considering it, but couldnt decide if there's a use-case when a defaultStatus would be needed in bootstrap-only mode. so we could leave that to a future PR.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -572,6 +567,103 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 			}, time.Second*5).Should(ContainSubstring("missing conditions"))
 		})
 
+		It("should count riskyOps and taintsToRemove when absent condition is satisfied via defaultStatus", func() {
+			// Absent condition + defaultStatus:False + requiredStatus:False:
+			// conditionFound=false → missingConditions++ → riskyOps++ (condition absent = risky)
```
ajaysundark, line 572: // no op in this PR

something that has been on my list is to refactor RiskyOperations. It doesnt give clarity to the user, more pronounced now with a default-status.

This field is suppose to be an indicator that when the rule was evaluated, there were X nodes with incomplete evaluation / condition states. Missing conditions were the obvious ones and we werent sure if we need to plan for anything else. But it feels like clubbing together everything into one bucket has less value than individual fields for the admin.

@Karthik-K-N Another one for alpha2, could we possibly deprecate this field and consider a MissingConditions explicitly instead?

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -766,6 +855,108 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 				}, BeNumerically(">=", int32(1))),
 			))
 		})
+
+		It(
+			"should satisfy a rule when an absent condition matches via defaultStatus (problem-gate scenario)",
+			func() {
+				// Scenario: NPD-style gate — taint is added when a problem condition fires.
```
ajaysundark, line 862: We should also consider documenting this use-case example for the site. Could you create a followup doc issue?

ajaysundark (commented):
/lgtm

api/v1alpha1/nodereadinessrule_types.go
AnuragThePathak: @ajaysundark I actually did that initially just to find that if I do it as a typical go-doc style comment it overlaps with the Kubernetes API schema. I felt it will be a terrible choice to display that information in the schema, hence shifted that information to the inline review of DefaultStatus field declaration.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -123,6 +124,17 @@ type ConditionRequirement struct {
...
+	// defaultStatus is the status a condition is evaluated to if the condition
+	// is not found in a node.
+	//
+	// Accepted values are True, False, Unknown. It is optional.
+	// When omitted, the effective default is Unknown, applied transparently by
+	// the controller at evaluation time.
+	//
+	// +optional
+	// +kubebuilder:validation:Enum=True;False;Unknown
+	DefaultStatus corev1.ConditionStatus `json:"defaultStatus,omitempty"` // Use GetDefaultStatus() for safe access; field may be empty even when a default applies.
```
AnuragThePathak, line 137: @ajaysundark here I have mentioned this way.

internal/controller/nodereadinessrule_controller_test.go
AnuragThePathak: https://github.com/kubernetes-sigs/node-readiness-controller/issues/289

api/v1alpha1/nodereadinessrule_types.go
AnuragThePathak: This is how I had to adjust it https://github.com/kubernetes-sigs/node-readiness-controller/pull/283#discussion_r3518862116

docs/book/src/user-guide/concepts.md
AnuragThePathak: I have listed it down under the defaultStatus issue only https://github.com/kubernetes-sigs/node-readiness-controller/issues/282#issuecomment-4877786594

Would you like to have a issue created or something else or this is good for now?

ajaysundark: yes, could you open a issue for this? If you are available to take it up, please feel free to raise a separate PR for the webhook changes, or we could leave this as a good-first issue.

I am inclined that it would be a good idea to start with a validation and later we can remove this when a requirement arises later.

api/v1alpha1/nodereadinessrule_types.go
ajaysundark: sgtm.

docs/book/src/user-guide/concepts.md
AnuragThePathak: @ajaysundark I was working on creating the issue and I created it as well now https://github.com/kubernetes-sigs/node-readiness-controller/issues/290

But while doing so I realized that what I mentioned here is slightly off the track the consequence especially. In current code, in order to complete a bootstrap it has to be first tainted by unsatisfying a Condition somehow, and when it satisfies then only bootstrap is marked as completed.

Definitely the other issue will fix this problem, and I'll update this part of doc too along with that but I just wondered if it's fundamentally intentional to mark bootsrap as completed only in case we removed a taint i.e. at least one condition was unsatisfied for a moment at least.

ajaysundark: > to complete a bootstrap it has to be first tainted by unsatisfying a Condition somehow

This is expected; The natural flow for nodes to satisfy bootstrap-only rules are by leveraging Kubelet's `--register-with-taints` flag to explicitly taint the node at the beginning and the components to satisfy the requirements when they are up.

AnuragThePathak: I remember seeing this in docs. Will revisit the section and capture the exact details while updating the docs.

config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml
AnuragThePathak: This is auto generated by `controller-gen` when we run `make manifests`. As far I can see, the reason quotes were applied around True and False by the tool is because without them they will be considered as a boolean value not string. That is not the case with Unknown hence no quote.

PR #284, opened 2026-06-27
Add Scale Test

vitorfloriano:
## Description

This PR adds a scale test harness for Node Readiness Controller. The test harness runs on a single rule on a 1000 nodes by default, and offers different tuning parameters for QPS, Burst, concurrent workers, among others.

### Benchmarking plan:

In this first iteration, we test how NRC performs when reconciling a single NodeReadinessRule in continuous mode to 1000 nodes (default node count). 
We switch conditions to true and false to observe the controller adding and removing taints from nodes.

Next iterations can include more test cases, like:
- Different node counts
- Different rule counts
- Bootstrap-only mode
- Different runtimes (kind, podman)

We also selected the metrics we deemed most relevant for the scalability report.

Here's a list of all metrics exposed/scraped during the test run:

<details><summary>available metrics at the controller endpoint</summary>
<p>

- Custom Controller Metrics (Node Readiness Specific)

  • node_readiness_condition_failures_total: Total number of failed condition evaluations by rule and condition name.
  • node_readiness_evaluation_duration_seconds_bucket / _count / _sum: Histogram of the duration (in seconds) taken
  to evaluate rules.
  • node_readiness_reconciliation_latency_seconds_bucket / _count / _sum: Histogram of total node readiness
  reconciliation latency.
  • node_readiness_rule_last_reconciliation_timestamp_seconds: Timestamp (Unix epoch) when a rule was last reconciled.
  • node_readiness_rules_total: Total number of active node readiness rules.
  • node_readiness_taint_operations_total: Total number of taint operations (add/remove) executed.

- controller-runtime Core Metrics

  • controller_runtime_active_workers: Number of active reconciler workers.
  • controller_runtime_max_concurrent_reconciles: Configured maximum concurrent reconciles for a controller.
  • controller_runtime_reconcile_total: Total number of reconciliations.
  • controller_runtime_reconcile_errors_total: Total number of reconciliation errors.
  • controller_runtime_reconcile_panics_total: Total number of reconciler panics.
  • controller_runtime_reconcile_time_seconds_bucket / _count / _sum: Histogram of time spent on reconciliations.
  • controller_runtime_conversion_webhook_panics_total / controller_runtime_webhook_panics_total: Total panics inside
  conversion or admission webhooks.
  • controller_runtime_terminal_reconcile_errors_total: Total terminal reconciliation errors.

- client-go Workqueue Metrics

  • workqueue_adds_total: Total number of tasks added to the workqueue.
  • workqueue_depth: Current queue depth.
  • workqueue_longest_running_processor_seconds: Processing time of the longest running task.
  • workqueue_queue_duration_seconds_bucket / _count / _sum: Histogram of how long items sit in the workqueue before
  processing.
  • workqueue_work_duration_seconds_bucket / _count / _sum: Histogram of how long processing an item takes.
  • workqueue_retries_total: Total number of task retries.
  • workqueue_unfinished_work_seconds: Seconds spent on currently running unfinished tasks.

- client-go REST Client Metrics

  • rest_client_requests_total: Total HTTP requests made by the Kubernetes API client.

- OS Process Metrics (Host Resources)

  • process_cpu_seconds_total: Total CPU user and system time spent in seconds.
  • process_resident_memory_bytes: Resident memory size (RSS) in bytes.
  • process_virtual_memory_bytes / _max_bytes: Virtual memory size.
  • process_open_fds / _max_fds: Number of open file descriptors.
  • process_network_receive_bytes_total / _transmit_bytes_total: Network traffic bytes.
  • process_start_time_seconds: Start time of the process.

- Go Runtime Internal Metrics

  • go_goroutines: Number of active goroutines.
  • go_threads: Number of OS threads.
  • go_info: Information about the Go runtime version.
  • go_memstats_alloc_bytes / go_memstats_sys_bytes: Go heap memory stats.
  • go_gc_duration_seconds / _count / _sum: Garbage collector cycles durations.
  • go_sched_latencies_seconds_bucket: Go scheduler queue latency histogram.

- Prometheus Scraper Metrics

  • scrape_duration_seconds: Time taken to run the scraper query.
  • scrape_samples_scraped: Number of metric samples retrieved.
  • up: Health status of the scraped controller instance (1 = healthy, 0 = down).

</p>
</details> 

### Key design choices:

- Use `kwokctl` to create the cluster and `kwok` binary to simulate k8s core components to keep it lightweight (no `docker`/ `kind` overhead).
- Run the NRC as a separate binary instead of deploying to the cluster, since that would not be possible with the fake cluster created by Kwok.
- Simulate changes in conditions using Kwok's `Stage` CRD. See: https://kwok.sigs.k8s.io/docs/user/stages-configuration/
- Use Kwok's native Prometheus integration to keep the setup simple. See: https://kwok.sigs.k8s.io/docs/examples/prometheus/

### Dependencies

The test only depends on `kwokctl` and `kubectl` being available at PATH. We ensure `kwokctl` is available and download the binary to `hack/tools/bin` right at the start of the test, if missing. 

The dependency on Prometheus and core components of the control-plane is completely managed by `kwokctl`.

## Testing

I've run the scale test multiple times, with different configurations.

## Release Notes

Since this is mostly infra, a release note may not be necessary.

> [!NOTE]
> LLM Usage: I used Gemini 3.5 Flash Medium with Antigravity CLI for planning and writing some parts with tricky logic. The code was reviewed by me and manually test multiple times.

Supersedes #185


hack/scale-test-setup.sh
```diff
@@ -0,0 +1,88 @@
...
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+CLUSTER_NAME="nrr-scale-test"
+KWOK_VERSION="${KWOK_VERSION:-v0.8.0}"
+PROJECT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+BIN_DIR="${PROJECT_DIR}/hack/tools/bin"
```
vitorfloriano: We're downloading the binary to the existing hack/tools/bin dir. Let me know if there would be a better destination.

hack/scale-test-setup.sh
```diff
@@ -0,0 +1,88 @@
...
+ensure_kwok_tools
+
+# Check if cluster already exists
+if ! "${KWOKCTL}" get clusters | grep -q "${CLUSTER_NAME}"; then
+    echo "Creating kwokctl cluster '${CLUSTER_NAME}'..."
+    "${KWOKCTL}" create cluster \
+        --name "${CLUSTER_NAME}" \
+        --runtime binary \
+        --node-lease-duration-seconds 400 \
+        --disable-qps-limits \
```
vitorfloriano: I thought of adding these two just so the apiserver/etcd does not become the bottleneck.

hack/scale-test-setup.sh
```diff
@@ -0,0 +1,88 @@
...
+
+# Main Setup Execution
+ensure_kwok_tools
+
+# Check if cluster already exists
+if ! "${KWOKCTL}" get clusters | grep -q "${CLUSTER_NAME}"; then
+    echo "Creating kwokctl cluster '${CLUSTER_NAME}'..."
+    "${KWOKCTL}" create cluster \
+        --name "${CLUSTER_NAME}" \
+        --runtime binary \
```
vitorfloriano: We're using kwok's binary as runtime to keep it lightweight (skip docker/kind overhead).

vitorfloriano (commented):
After much research, this is the initial work. I got a LGTM on the approach from Kwok's maintainer, so I believe I'm headed in the right direction. PTAL.

vitorfloriano:
@ajaysundark @Priyankasaggu11929 @Karthik-K-N 

Following up to our conversation, I'm going to refine the markdown report, perhaps vet some of the metrics there to make it cleaner and easier to draw insights from.

vitorfloriano:
@ajaysundark @Priyankasaggu11929 @Karthik-K-N 

I believe this is ready for review now. 

I have simplified the test by a lot to make review easier. I plan to add more test cases in next iterations.

PTAL. Thanks!

test/scale/scale_suite_test.go
```diff
@@ -0,0 +1,273 @@
...
+	crdConfigPath := filepath.Join(projectRootDir, "config", "crd")
+	crdCmd := exec.Command("kubectl", "apply", "-k", crdConfigPath) // #nosec G204
+	crdOutput, err := utils.Run(crdCmd)
+	Expect(err).NotTo(HaveOccurred(), "Failed to apply NodeReadinessRule CRD via Kustomize:\n%s", crdOutput)
+
+	By("Compiling node-readiness-controller manager binary")
+	controllerBinName := "node-readiness-controller"
+	if runtime.GOOS == "windows" {
+		controllerBinName += ".exe"
+	}
```
Priyankasaggu11929: we don't support windows right now

https://github.com/kubernetes-sigs/node-readiness-controller/blob/c1021031fb688c3ae61af1b95fbce1e126c58e04/Makefile#L218

test/scale/scale_suite_test.go
```diff
@@ -0,0 +1,273 @@
...
+	if ruleConc := os.Getenv("RULE_CONCURRENT_RECONCILES"); ruleConc != "" {
+		args = append(args, "--rule-concurrent-reconciles="+ruleConc)
+	}
+
+	if runtime.GOOS != "windows" {
+		controllerCmd = exec.Command("setsid", append([]string{controllerBinPath}, args...)...) // #nosec G204
+	} else {
+		controllerCmd = exec.Command(controllerBinPath, args...) // #nosec G204
+	}
```
Priyankasaggu11929: same as https://github.com/kubernetes-sigs/node-readiness-controller/pull/284/changes#r3621819393

no support for windows today

test/scale/testdata/security-agent-stage-false.yaml
```diff
@@ -0,0 +1,27 @@
...
+        status: {{ .status | quote }}
+        reason: {{ .reason | quote }}
+        message: {{ .message | quote }}
+      {{- end }}
+      {{- end }}
+      - type: security.node.x-k8s.io/SecurityAgentReady
+        status: "False"
+        reason: "SecurityAgentUnhealthy"
+        message: "Security agent not running"
+        lastTransitionTime: "{{ Now }}"
```
Priyankasaggu11929, line 34:
```suggestion
  selector:
    matchLabels:
      type: kwok
    matchExpressions:
      - key: ".status.conditions[] | select(.type==\"security.node.x-k8s.io/SecurityAgentReady\") | .status"
        operator: NotIn
        values:
          - "False"
  next:
    statusTemplate: |
      conditions:
      {{- range .status.conditions }}
      {{- if ne .type "security.node.x-k8s.io/SecurityAgentReady" }}
      - type: {{ .type }}
        status: {{ .status | quote }}
        reason: {{ .reason | quote }}
        message: {{ .message | quote }}
        lastHeartbeatTime: {{ .lastHeartbeatTime | quote }}
        lastTransitionTime: {{ .lastTransitionTime | quote }}
      {{- end }}
      {{- end }}
      - type: security.node.x-k8s.io/SecurityAgentReady
        status: "False"
        reason: "SecurityAgentUnhealthy"
        message: "Security agent not running"
        lastTransitionTime: "{{ Now }}"
```

selecting on just `matchLabels.type: kwok` is not enough because it matches every kwok node regardless of whether it already has a `security.node.x-k8s.io/SecurityAgentReady:False` set.

so right now kwok re-fires the stage on every NRC taint patch as well, even when the condition is already set, which becomes one extra unnecessary status write per taint operation and add noise in the metrics we are trying to measure.

---

Also, every time the stage fires, `lastHeartbeatTime` and `lastTransitionTime` are wiped from other node status conditions (like `Ready`, `MemoryPressure`, `PIDPressure` cetc.)   
verified locally locally, both fields go `null` after the stage fires.

IMO, the stage should only be adding `SecurityAgentReady` condition, not touching other node conditions at all.

test/scale/testdata/security-agent-stage-true.yaml
```diff
@@ -0,0 +1,27 @@
...
+        status: {{ .status | quote }}
+        reason: {{ .reason | quote }}
+        message: {{ .message | quote }}
+      {{- end }}
+      {{- end }}
+      - type: security.node.x-k8s.io/SecurityAgentReady
+        status: "True"
+        reason: "SecurityAgentHealthy"
+        message: "Security agent initialized"
+        lastTransitionTime: "{{ Now }}"
```
Priyankasaggu11929, line 34:
```suggestion
  selector:
    matchLabels:
      type: kwok
    matchExpressions:
      - key: ".status.conditions[] | select(.type==\"security.node.x-k8s.io/SecurityAgentReady\") | .status"
        operator: NotIn
        values:
          - "True"
  next:
    statusTemplate: |
      conditions:
      {{- range .status.conditions }}
      {{- if ne .type "security.node.x-k8s.io/SecurityAgentReady" }}
      - type: {{ .type }}
        status: {{ .status | quote }}
        reason: {{ .reason | quote }}
        message: {{ .message | quote }}
        lastHeartbeatTime: {{ .lastHeartbeatTime | quote }}
        lastTransitionTime: {{ .lastTransitionTime | quote }}
      {{- end }}
      {{- end }}
      - type: security.node.x-k8s.io/SecurityAgentReady
        status: "True"
        reason: "SecurityAgentHealthy"
        message: "Security agent initialized"
        lastTransitionTime: "{{ Now }}"
```

same reasoning as https://github.com/kubernetes-sigs/node-readiness-controller/pull/284/changes#r3622248531

test/scale/helpers.go
```diff
@@ -0,0 +1,281 @@
...
+		LabelSelector: "type=kwok",
+	})
+	if err != nil {
+		return 0, err
+	}
+
+	count := 0
+	for _, node := range nodes.Items {
+		for _, taint := range node.Spec.Taints {
+			if taint.Key == "readiness.k8s.io/SecurityAgentNotReady" && taint.Value == "pending" {
```
Priyankasaggu11929: in both functions (`countKwokNodes` and `countTaintedNodes`), you're hardcoding the label selector `"type=kwok"` and the taint key `"readiness.k8s.io/SecurityAgentNotReady:pending"` values

If we need to add a different taint later, the helper will need to change.

let's update both functions to take the label selector/taint key as parameters instead.

test/scale/helpers.go
```diff
@@ -0,0 +1,281 @@
...
+
+	valStr, ok := promResp.Data.Result[0].Value[1].(string)
+	if !ok {
+		return "", fmt.Errorf("invalid value format")
+	}
+	return valStr, nil
+}
+
+//nolint:unparam
+func collectAndReportMetricsForWindow(ctx context.Context, phaseTitle string, phaseStart time.Time, phaseEnd time.Time) (queryResult, error) {
```
Priyankasaggu11929: this is confusing.

 `collectAndReportMetricsForWindow` function always results in nil error (it will never change to anything else) and you're also suppressing the linter with `//nolint:unparam` 

let's drop the "error" return entirely (if that is the intention) or do proper error handling.

---

also, the function name confused me a bit ( i thought it was for Windows OS) - can we call it something else please :)

test/scale/scale_suite_test.go
```diff
@@ -0,0 +1,273 @@
...
+	kwokctlBinaryPath = ensureKwokctl(kwokctlVersion, toolsBinDir)
+
+	By("Cleaning up any existing simulated cluster and stale controller processes")
+	_ = exec.Command(kwokctlBinaryPath, "delete", "cluster").Run()
+
+	if runtime.GOOS == "windows" {
+		_ = exec.Command("taskkill", "/IM", "node-readiness-controller.exe", "/F").Run()
+	} else {
+		_ = exec.Command("pkill", "-f", "node-readiness-controller").Run()
+	}
```
Priyankasaggu11929: NRC only support linux right now.

same as https://github.com/kubernetes-sigs/node-readiness-controller/pull/284/changes#r3621819393

test/scale/helpers.go
```diff
@@ -0,0 +1,281 @@
...
+			}
+		}
+	}
+	return count, nil
+}
+
+func queryPrometheusInstant(ctx context.Context, query string, ts float64) (string, error) {
+	// Construct the Prometheus Instant Query HTTP endpoint.
+	// Query parameters are URL-escaped, and the evaluation timestamp float is formatted to 3 decimal places.
+	urlStr := fmt.Sprintf("http://127.0.0.1:%s/api/v1/query?query=%s&time=%.3f", prometheusPort, url.QueryEscape(query), ts)
```
Priyankasaggu11929:
```suggestion
	urlStr := fmt.Sprintf("http://%s:%s/api/v1/query?query=%s&time=%.3f", prometheusHost, prometheusPort, url.QueryEscape(query), ts)
```

prometheusHost is also configurable right?

test/scale/scale_suite_test.go
```diff
@@ -0,0 +1,273 @@
...
+//go:embed testdata/security-agent-stage-false.yaml
+var securityAgentStageFalseManifest string
+
+//go:embed testdata/security-agent-stage-true.yaml
+var securityAgentStageTrueManifest string
+
+//go:embed testdata/prometheus-nrc-job.yaml
+var prometheusJobTemplate string
+
+var _ = BeforeSuite(func() {
```
Priyankasaggu11929, line 83: This block is 100+ lines and very hard to read (and maintain over time).

I feel like each By() block is right now doing a single operation and most (and maybe all) can turn into a helper function in helper.go 

That would then look something like:

```
var _ = BeforeSuite(func() {
    readEnvConfig()

    By("Ensuring kwokctl binary is present")
    kwokctlBinaryPath = ensureKwokctl(...)

    By("Cleaning up any existing simulated cluster and stale controller processes")
    cleanupStaleResources(...)

    By("Creating the simulated KWOK cluster")
    createKwokCluster(...)

    By("Applying NodeReadinessRule CRD manifests")
    applyCRDs(...)

    By("Compiling node-readiness-controller manager binary")
    controllerBinPath = buildController(...)

    By("Configuring and reloading Prometheus")
    patchPrometheusConfig(...)

    By("Scaling nodes up")
    scaleNodes(...)

    By("Starting the controller")
    controllerCmd = startController(...)
})
```
WDYT?

test/scale/scale_suite_test.go
```diff
@@ -0,0 +1,273 @@
...
+
+	prometheusConfigBytes, err := os.ReadFile(prometheusConfigPath) // #nosec G304
+	Expect(err).NotTo(HaveOccurred(), "Failed to read Prometheus configuration")
+
+	newConfig := string(prometheusConfigBytes) + fmt.Sprintf(prometheusJobTemplate, controllerMetricsPort)
+	err = os.WriteFile(prometheusConfigPath, []byte(newConfig), 0600)
+	Expect(err).NotTo(HaveOccurred(), "Failed to update Prometheus configuration with new job")
+
+	By("Restarting Prometheus to load updated configuration")
+	err = exec.Command("pkill", "-SIGHUP", "prometheus").Run()
```
Priyankasaggu11929: Prometheus provides and API endpoint for reloading with the updated config file - https://prometheus.io/docs/prometheus/latest/management_api/#reload 

I think we should use that instead of directly sending pkill signals

Although right now, using the API endpoint requires starting prometheus with `--web.enable-lifecycle` flag, but kwokctl doesn't allow setting a flag with empty value today.

Created PR https://github.com/kubernetes-sigs/kwok/pull/172 

when ^ PR or some solution along the lines merges, we can use the prometheus API endpoint.


Till then let's use `-HUP` instead of `-SIGHUP`

test/scale/testdata/prometheus-nrc-job.yaml
```diff
@@ -0,0 +1,7 @@
+- job_name: node-readiness-controller
+  scrape_interval: 1s
+  metrics_path: /metrics
+  scheme: http
+  static_configs:
+  - targets:
+    - 127.0.0.1:%s
```
Priyankasaggu11929: since you're already using go templates in other parts, let's turn this file also into a go template, 

because right now there's only one `%s` so it works, but if we end up adding more placeholders for replacement later, doing that with simple `fmt.Sprintf` would become complex or confusing even.

Priyankasaggu11929 (changes requested):
Thanks for your work, @vitorfloriano.

Did a first round of review. added a bunch of inline comments.

test/scale/promqueries.go
```diff
@@ -0,0 +1,177 @@
...
+	{
+		Key:       "workqueue_queue_duration_p99",
+		QueryTmpl: "histogram_quantile(0.99, sum(rate(workqueue_queue_duration_seconds_bucket{name=\"node\",job=\"node-readiness-controller\"}[%ds])) by (le))",
+		Unit:      "s",
+	},
+	{
+		Key:       "workqueue_work_duration_p50",
+		QueryTmpl: "histogram_quantile(0.50, sum(rate(workqueue_work_duration_seconds_bucket{name=\"node\",job=\"node-readiness-controller\"}[%ds])) by (le))",
+		Unit:      "s",
+	},
```
Priyankasaggu11929, line 82: all three (`reconcile_time*`, `reconciliation_latency*`, `workqueue_work_duration*`) are measuring the same reconcile latency

let's keep `reconciliation_latency*` since it is specific to NRC and tagged per rule?

test/scale/promqueries.go
```diff
@@ -0,0 +1,177 @@
...
+	{
+		Key:       "rule_evaluation_duration_p90",
+		QueryTmpl: "histogram_quantile(0.90, sum(rate(node_readiness_evaluation_duration_seconds_bucket{job=\"node-readiness-controller\"}[%ds])) by (le))",
+		Unit:      "s",
+	},
+	{
+		Key:       "rule_evaluation_duration_p99",
+		QueryTmpl: "histogram_quantile(0.99, sum(rate(node_readiness_evaluation_duration_seconds_bucket{job=\"node-readiness-controller\"}[%ds])) by (le))",
+		Unit:      "s",
+	},
```
Priyankasaggu11929, line 107: we can drop these?

`reconciliation_latency*`  is the total time for the whole reconcile loop, and  `rule_evaluation_duration*`  is just one step inside it, which is already accounted for.

test/scale/promqueries.go
```diff
@@ -0,0 +1,177 @@
...
+	{
+		Key:       "resident_memory_peak",
+		QueryTmpl: "max(max_over_time(process_resident_memory_bytes{job=\"node-readiness-controller\"}[%ds]))",
+		Unit:      "bytes",
+	},
+	{
+		Key:       "resident_memory_avg",
+		QueryTmpl: "avg(avg_over_time(process_resident_memory_bytes{job=\"node-readiness-controller\"}[%ds]))",
+		Unit:      "bytes",
+	},
```
Priyankasaggu11929, line 152: i think under load, `cpu_cores_peak` and `resident_memory_peak` are useful, the other two we can drop.

test/scale/promqueries.go
```diff
@@ -0,0 +1,177 @@
...
+		QueryTmpl: "sum(workqueue_retries_total{name=\"node\",job=\"node-readiness-controller\"}) - (sum(workqueue_retries_total{name=\"node\",job=\"node-readiness-controller\"} @ %.3f) or vector(0))",
+		Unit:      "retries",
+		IsCounter: true,
+	},
+	{
+		Key:       "workqueue_retries_rules",
+		QueryTmpl: "sum(workqueue_retries_total{name=\"nodereadiness-controller\",job=\"node-readiness-controller\"}) - (sum(workqueue_retries_total{name=\"nodereadiness-controller\",job=\"node-readiness-controller\"} @ %.3f) or vector(0))",
+		Unit:      "retries",
+		IsCounter: true,
+	},
```
Priyankasaggu11929, line 164: for now, we have only one rule and x no of nodes (meaning no load on rule reconciller)

if there's plan to scale test with multiple no of rules also, then let's keep it, otherwise remove?

test/scale/promqueries.go
```diff
@@ -0,0 +1,177 @@
...
+		Unit:      "adds",
+		IsCounter: true,
+	},
+	{
+		Key:       "workqueue_adds_rules",
+		QueryTmpl: "sum(workqueue_adds_total{name=\"nodereadiness-controller\",job=\"node-readiness-controller\"}) - (sum(workqueue_adds_total{name=\"nodereadiness-controller\",job=\"node-readiness-controller\"} @ %.3f) or vector(0))",
+		Unit:      "adds",
+		IsCounter: true,
+	},
+}
```
Priyankasaggu11929, line 188: we are getting this same info from ` taint_operations_add/remove` right?

Priyankasaggu11929 (commented):
some notes from my side on the metrics added so far (will also let other folks to comment/correct)

test/scale/testdata/security-agent-stage-false.yaml
vitorfloriano: Hmm...that's interesting. I ran the test with this suggestion and did not see any difference in the duration of the runs. Perhaps in CI on constrained resources this overhead on KWOK may indeed affect the metrics for NRC. 

I couldn't get the total of requests made by KWOK because it does not seem to expose this metric (the controller does not seem to use controller-runtime, which is a bit maverick). 

I only queried metrics for NRC, so overhead on KWOK went unnoticed. Thanks for pointing that out.

I applied your suggestion.

test/scale/helpers.go
vitorfloriano: Done.

test/scale/scale_suite_test.go
vitorfloriano: Done.

test/scale/helpers.go
vitorfloriano: I refactored the function into two different helpers with no returning errors and better names. Hope that looks a bit clearer.

test/scale/helpers.go
vitorfloriano: Hm...not so sure about it. AFAIK KWOK does not have a flag for setting that. See: https://kwok.sigs.k8s.io/docs/generated/kwokctl_create_cluster/

So, considering the queries would always be directed to the (configurable) port on the (not configurable) localhost, I think hardcoding the loopback IP here is fine, isn't it? I can't see why we would have a different host for that? Would that not work in Prow/DinD ? Even if we move to the `kind` runtime, I believe we would need to port-forward Prometheus in the BeforeSuite, wouldn't we? 🤔

Anyway, just for reference, here's the Prometheus config on the binary runtime, after we append the NRC job:

<details><summary>Prometheus Config</summary>
<p>

```yaml
global:
  evaluation_interval: 15s
  scrape_interval: 15s
  scrape_timeout: 10s
scrape_configs:
- enable_http2: true
  follow_redirects: true
  honor_timestamps: true
  job_name: etcd
  metrics_path: /metrics
  scheme: http
  static_configs:
  - targets:
    - 127.0.0.1:32765
- enable_http2: true
  follow_redirects: true
  honor_timestamps: true
  job_name: kube-apiserver
  metrics_path: /metrics
  scheme: https
  static_configs:
  - targets:
    - 127.0.0.1:32764
  tls_config:
    cert_file: /home/vitorfloriano/.kwok/clusters/kwok/pki/admin.crt
    insecure_skip_verify: true
    key_file: /home/vitorfloriano/.kwok/clusters/kwok/pki/admin.key
- enable_http2: true
  follow_redirects: true
  honor_timestamps: true
  job_name: kube-controller-manager
  metrics_path: /metrics
  scheme: https
  static_configs:
  - targets:
    - 127.0.0.1:32762
  tls_config:
    cert_file: /home/vitorfloriano/.kwok/clusters/kwok/pki/admin.crt
    insecure_skip_verify: true
    key_file: /home/vitorfloriano/.kwok/clusters/kwok/pki/admin.key
- enable_http2: true
  follow_redirects: true
  honor_timestamps: true
  job_name: kube-scheduler
  metrics_path: /metrics
  scheme: https
  static_configs:
  - targets:
    - 127.0.0.1:32761
  tls_config:
    cert_file: /home/vitorfloriano/.kwok/clusters/kwok/pki/admin.crt
    insecure_skip_verify: true
    key_file: /home/vitorfloriano/.kwok/clusters/kwok/pki/admin.key
- http_sd_configs:
  - url: http://127.0.0.1:32763/discovery/prometheus
  job_name: kwok-controller-metrics-discovery
- enable_http2: true
  follow_redirects: true
  honor_timestamps: true
  job_name: kwok-controller
  metrics_path: /metrics
  scheme: http
  static_configs:
  - targets:
    - 127.0.0.1:32763
- enable_http2: true
  follow_redirects: true
  honor_timestamps: true
  job_name: prometheus
  metrics_path: /metrics
  scheme: http
  static_configs:
  - targets:
    - 127.0.0.1:9090
- job_name: node-readiness-controller
  scrape_interval: 1s
  metrics_path: /metrics
  scheme: http
  static_configs:
  - targets:
    - 127.0.0.1:8080
```
</p>
</details>

test/scale/scale_suite_test.go
vitorfloriano: I remember trying to use `--extra-args` but facing an error. Didn't go down the rabbit hole though. But agreed, pkill kinda dangerous. 

Changed the flag to -HUP as suggested.

test/scale/testdata/prometheus-nrc-job.yaml
vitorfloriano: Done.

test/scale/scale_suite_test.go
vitorfloriano: Good call. I refactored both the spec and the suite to make it easier to read. I also split the helpers into spec and suite helpers.

test/scale/scale_suite_test.go
vitorfloriano: Fixed.

vitorfloriano:
@Priyankasaggu11929 I addressed your comments from your first round and refactored a bunch of stuff. I also reshaped the template for the report and refactored the test config to make it more concise.

I'll leave the Prom queries out of it for now to let others review.

PTAL.

test/scale/promqueries.go
```diff
@@ -0,0 +1,177 @@
+//go:build scale
```
ajaysundark, line 1: cc @ajaysundark review the metric signals and establish the baseline.

ajaysundark: we reviewed this offline, the metrics shape look good. there maybe some changes we could add in the future (for example, the new observability shape being discussed). but we can come back to it as needed.

test/scale/testdata/scalability_report.md.tmpl
```diff
@@ -0,0 +1,46 @@
...
+| Workqueue Metric | Node Controller | Rule Controller |
+| :--- | :--- | :--- |
+| **Workqueue Retries** | {{index .Metrics "workqueue_retries_node"}} | {{index .Metrics "workqueue_retries_rules"}} |
+| **Workqueue Adds** | {{index .Metrics "workqueue_adds_node"}} | {{index .Metrics "workqueue_adds_rules"}} |
+
+| Operational Counter | Value |
+| :--- | :--- |
+| **Taint Operations (Added)** | {{index .Metrics "taint_operations_add"}} |
+| **Taint Operations (Removed)** | {{index .Metrics "taint_operations_remove"}} |
+| **Condition Check Failures** | {{index .Metrics "condition_failures_total"}} |
```
ajaysundark, line 35: not actionable now. but I proposed to deprecate this metric in the new metric surface. The reason is that this just counts "condition-not-satisfied" at every iteration, which really are not 'failures'. At scale, this shouldn't bring out bad impression of state of things

ajaysundark:
/lgtm
/approve

Thanks @vitorfloriano! We want to use this for release pipeline to be able to continuously validate for scale for all the release versions. We should plan to see how we can integrate this into CI with kwok "in-cluster" mode. We can follow up that work in #409

ajaysundark:
cc @rawadhossain

PR #285, opened 2026-07-02
Remove duplicate bootstrap duration observation in taint removal path

rawadhossain:
## Description
cleanup: duplicate observation of node_readiness_bootstrap_duration_seconds in the bootstrap-only taint removal path.
Two consecutive observation blocks recording the same bootstrap completion event.

This is important because upcoming metrics, dashboards, and SLOs rely on this histogram for accurate calculations. Duplicate samples would skew those results.

### Files changed

| File | Change |
|------|--------|
| `internal/controller/nodereadinessrule_controller.go` | Remove duplicate Block |
| `internal/controller/nodereadinessrule_controller_test.go` | Add regression test + `histogramSampleCount` helper |

## Type of Change
/kind bug

## Testing
Added test to verify the histogram sample count increases by exactly one for a bootstrap completion event
- Ran:
  - `make test`
  - `make lint`

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/internal/controller/nodereadinessrule_controller_test.go
+++ b/internal/controller/nodereadinessrule_controller_test.go
@@ -23,6 +23,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -46,6 +47,12 @@ func counterValue(counter interface{ Write(*dto.Metric) error }) float64 {
 	metric := &dto.Metric{}
 	Expect(counter.Write(metric)).To(Succeed())
 	return metric.GetCounter().GetValue()
+}
+
+func histogramSampleCount(histogram interface{ Write(*dto.Metric) error }) uint64 {
+	metric := &dto.Metric{}
+	Expect(histogram.Write(metric)).To(Succeed())
+	return metric.GetHistogram().GetSampleCount()
 }
 
 // errorInjectingClient forces Patch to fail for selected nodes.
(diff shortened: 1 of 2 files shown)


internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -1030,6 +1037,73 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+				Status: corev1.NodeStatus{
+					Conditions: []corev1.NodeCondition{
+						{
+							Type:               "Ready",
+							Status:             corev1.ConditionTrue,
+							LastTransitionTime: transitionTime,
+						},
+					},
+				},
+			}
```
Karthik-K-N: I think it does not matter, wont the create ignore status?
Usually we create only with spec

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -1030,6 +1037,73 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+			Expect(k8sClient.Create(ctx, node)).To(Succeed())
+			defer func() { _ = k8sClient.Delete(ctx, node) }()
+
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: nodeName}, node)).To(Succeed())
+			node.Status.Conditions[0].LastTransitionTime = transitionTime
+			node.Status.Conditions[0].Status = corev1.ConditionTrue
+			Expect(k8sClient.Status().Update(ctx, node)).To(Succeed())
```
Karthik-K-N: Usually Update() call should go inside a Eventually block to have some retry, How about something like this
```
    Eventually(func() error {
        if err := k8sClient.Get(ctx, types.NamespacedName{Name: nodeName}, node); err != nil {
            return err
        }
        node.Status.Conditions = []corev1.NodeCondition{
            {
                Type:               "Ready",
                Status:             corev1.ConditionTrue,
                LastTransitionTime: transitionTime,
            },
        }
        return k8sClient.Status().Update(ctx, node)
    }, "5s", "100ms").Should(Succeed())
```

Karthik-K-N (commented):
Thank you for catching this. If possible we can do small improvement in test from the readability and stability aspect apart from that LGTM. Thank you.

internal/controller/nodereadinessrule_controller_test.go
rawadhossain: you're right. Removed the Status block, now node is created with spec only

internal/controller/nodereadinessrule_controller_test.go
rawadhossain: Thanks for pointing this out. Commiting the changes now.

rawadhossain:
@Karthik-K-N Made the changes. PTAL.
Thanks.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -1030,6 +1037,72 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+					NodeSelector: metav1.LabelSelector{
+						MatchLabels: map[string]string{"env": "bootstrap-duration"},
+					},
+					EnforcementMode: nodereadinessiov1alpha1.EnforcementModeBootstrapOnly,
+				},
+			}
+			Expect(k8sClient.Create(ctx, rule)).To(Succeed())
+			defer func() { _ = k8sClient.Delete(ctx, rule) }()
+
+			transitionTime := metav1.NewTime(time.Now().Add(5 * time.Second))
```
Karthik-K-N, line 1068: just a suggestion, its always better to declare the variable as near to the place where its being used for better readability.

Karthik-K-N (commented):
/lgtm

Thank you.

ajaysundark:
Thanks for catching this.

/lgtm
/approve

PR #286, opened 2026-07-02
feat: collector-based rule_matched_nodes and typed failure reasons

rawadhossain:
## Description

Adds `node_readiness_rule_matched_nodes` metric to the scrape-time collector pattern introduced in #398.

The metric is computed on each scrape from the cached Node list, sharing the same snapshot as `node_readiness_rule_nodes`. Dry-run and deleting rules are also included.

#### (exisitng) `node_readiness_failures_total` (new reason: `AnnotationPatchFailed`)
Also adds `AnnotationPatchFailed` as a new reason for the existing `node_readiness_failures_total` metric. Tracks failures when writing the bootstrap-completion annotation, keeping all failure reporting under the existing `failures_total` counter.

### Changes
- Moved `node_readiness_rule_matched_nodes` to the scrape-time `ReadinessCollector`.
- Shares the Node list with `node_readiness_rule_nodes`.
- Includes dry-run and deleting rules.
- Added `AnnotationPatchFailed` to `node_readiness_failures_total`.
- Failure reasons (`EvaluationError`, `AddTaintError`, `RemoveTaintError`, `AnnotationPatchFailed`) are now defined as a type with named constants. 

## Type of Change
/kind feature

## Testing
- Verified the metric against a live cluster and confirmed the counts match the current Node selectors.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -156,7 +156,7 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			log.Error(err, "Failed to evaluate rule for node",
 				"node", node.Name, "rule", rule.Name)
 			// Continue with other rules even if one fails
-			r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
+			r.recordNodeFailure(rule, node.Name, string(metrics.FailureReasonEvaluationError), err.Error())
 			errs = append(errs, err)
 			metrics.Failures.WithLabelValues(rule.Name, string(metrics.FailureReasonEvaluationError)).Inc()
 		}
@@ -483,6 +483,7 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
 	switch {
 	case err != nil:
 		log.Error(err, "Failed to mark bootstrap completed", "node", nodeName, "rule", rule.Name, "uid", rule.GetUID())
+		metrics.Failures.WithLabelValues(rule.Name, string(metrics.FailureReasonAnnotationPatchFailed)).Inc()
 	case deferred:
 		log.Info("Deferring bootstrap completion - rule taint still present on node",
 			"node", nodeName, "rule", rule.Name, "taint", rule.Spec.Taint.Key)
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -138,19 +138,9 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	// Update rule cache (after cleanup)
 	r.Controller.updateRuleCache(ctx, rule)
 
-	// Filter nodes once and update node_readiness_rule_matched_nodes.
-	var matchedNodes []corev1.Node
-	for i := range nodeList.Items {
-		if r.Controller.ruleAppliesTo(ctx, rule, &nodeList.Items[i]) {
-			matchedNodes = append(matchedNodes, nodeList.Items[i])
-		}
-	}
-	metrics.RuleMatchedNodes.WithLabelValues(rule.Name).Set(float64(len(matchedNodes)))
-	filteredList := &corev1.NodeList{Items: matchedNodes}
-
 	// Handle dry run
 	if rule.Spec.DryRun {
-		if err := r.Controller.processDryRun(ctx, rule, filteredList); err != nil {
+		if err := r.Controller.processDryRun(ctx, rule, nodeList); err != nil {
 			log.Error(err, "Failed to process dry run", "rule", rule.Name)
 			return ctrl.Result{RequeueAfter: time.Minute}, err
 		}
@@ -159,7 +149,7 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		rule.Status.DryRunResults = readinessv1alpha1.DryRunResults{}
 
 		// Process all applicable nodes for this rule
-		if err := r.Controller.processAllNodesForRule(ctx, rule, filteredList); err != nil {
+		if err := r.Controller.processAllNodesForRule(ctx, rule, nodeList); err != nil {
 			log.Error(err, "Failed to process nodes for rule", "rule", rule.Name)
 			return ctrl.Result{RequeueAfter: time.Minute}, err
 		}
@@ -223,7 +213,6 @@ func (r *RuleReconciler) reconcileDelete(ctx context.Context, rule *readinessv1a
 	metrics.BootstrapCompleted.DeleteLabelValues(rule.Name)
 	metrics.BootstrapDuration.DeleteLabelValues(rule.Name)
 	metrics.EvaluationDuration.DeleteLabelValues(rule.Name)
-	metrics.RuleMatchedNodes.DeleteLabelValues(rule.Name)
 
 	// For multi-label metrics, use DeletePartialMatch to wipe all combinations
 	metrics.NodesByState.DeletePartialMatch(ruleLabel)
@@ -293,7 +282,7 @@ func (r *RuleReadinessController) cleanupDeletedNodes(ctx context.Context, rule 
 func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule, nodeList *corev1.NodeList) error {
 	log := ctrl.LoggerFrom(ctx)
 
-	log.Info("Processing all nodes for rule", "rule", rule.Name, "matchedNodes", len(nodeList.Items))
+	log.Info("Processing all nodes for rule", "rule", rule.Name, "totalNodes", len(nodeList.Items))
 
 	var appliedNodes []string
 	for _, node := range nodeList.Items {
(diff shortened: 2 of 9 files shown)


internal/controller/nodereadinessrule_controller.go
```diff
@@ -135,6 +135,15 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	// Update rule cache (after cleanup)
 	r.Controller.updateRuleCache(ctx, rule)
 
+	// Update selector_matched_nodes_total from the node list we already fetched.
+	var matchedCount int
+	for i := range nodeList.Items {
```
AvineshTripathi: we already do the same thing inside processAllNodesForRule [here](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/nodereadinessrule_controller.go#L284)

internal/metrics/metrics.go
```diff
@@ -117,6 +117,35 @@ var (
...
+	SelectorMatchedNodes = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "node_readiness_selector_matched_nodes_total",
+			Help: "Number of nodes matched by a rule's NodeSelector",
+		},
+		[]string{"rule"},
+	)
+
+	// BootstrapCompletionErrors tracks failures writing the bootstrap-completion annotation.
+	BootstrapCompletionErrors = prometheus.NewCounterVec(
```
AvineshTripathi: I couldn't find this in the doc. Where are we planning to use these metrics? I feel counter without reason of failure won't be that useful. Please correct me if I am understanding this wrongly

internal/metrics/metrics.go
```diff
@@ -117,6 +117,35 @@ var (
...
+		prometheus.CounterOpts{
+			Name: "node_readiness_bootstrap_completion_errors_total",
+			Help: "Total failures writing the bootstrap-completion annotation",
+		},
+		[]string{"rule"},
+	)
+
+	// BootstrapNRCDuration tracks time from NRC taint application to bootstrap completion.
+	// Unlike bootstrap_duration_seconds, this excludes pre-NRC node boot time.
+	BootstrapNRCDuration = prometheus.NewHistogramVec(
```
AvineshTripathi: missing implementation here or is this the pending decision one?

internal/controller/nodereadinessrule_controller.go
```diff
@@ -406,6 +418,8 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
 				if duration > 0 {
 					metrics.BootstrapDuration.WithLabelValues(rule.Name).Observe(duration)
 				}
+				// TODO: pending decision — record BootstrapNRCDuration from readiness.k8s.io/taint-applied-<rule> to time.Now(); skip if missing.
```
AvineshTripathi: [This](https://github.com/kubernetes-sigs/node-readiness-controller/pull/224) PR introduces a new format of bootstrap annotation where the value is a JSON. Instead of creating a new annotation `readiness.k8s.io/taint-applied-<rule>` we could just add a field in the json value.

AvineshTripathi (commented):
Thanks for the PR! Left a few comments.

internal/controller/nodereadinessrule_controller.go
rawadhossain: Yeah, I intentionally kept it before the dry-run check so the gauge gets updated for both dry-run and normal rules in one place. If I move it into processAllNodesForRule, dryrun rules would never update the metric as they go through processDryRun. Open to refactor it if you think splitting it across both paths is cleaner.

internal/metrics/metrics.go
rawadhossain: yes it wasnt there in the doc, I actually found this gap while implementing the metrics.

Right now, if writing the bootstrap annotation fails after all retries are exhausted, the node gets excluded from `bootstrap_duration_seconds` with no signal anywhere. This counter makes that failure visible.
Good point on the reason label, I’ll add it so it's explicit and easy to extend later if needed.

internal/metrics/metrics.go
rawadhossain: Yes its the pending one, was waiting for the decision. 

Btw I think it's cleaner to remove this metric from the PR for now and open a separate PR once PR #224 merges. what do you say?

internal/controller/nodereadinessrule_controller.go
rawadhossain: I like that approach better too. As it was still in discussion, I went with the annotation one.

Since #224 is still pending, I'll remove `bootstrap_nrc_duration_seconds` metric from here entirely and implement JSON value at taint then open a PR once #224 merges. So the other two metrics are not blocked.

rawadhossain:
@AvineshTripathi PTAL. I'll commit the changes once you approve. 
Thanks.

internal/controller/nodereadinessrule_controller.go
AvineshTripathi: do you think a cleanup would be better? We could filter nodes before the evaluating them in dryRun or regular run

internal/metrics/metrics.go
AvineshTripathi: works for me!

AvineshTripathi:
@rawadhossain I left one comment PHAL and maybe remove the Bootstrap duration metrics till we merge https://github.com/kubernetes-sigs/node-readiness-controller/pull/224

internal/controller/nodereadinessrule_controller.go
rawadhossain: yes, that seems cleaner. I'll filter the nodes once in Reconcile(), set the gauge there, and pass the filtered list to both paths. Thanks.

internal/metrics/metrics.go
rawadhossain: Added the reason label

rawadhossain:
@AvineshTripathi I have made the changes. PTAL.
Thanks.

AvineshTripathi:
/lgtm

cc @ajaysundark for final look

ajaysundark:
Could `node_readiness_bootstrap_completion_errors_total` be folded within `failures_total` with a dedicated reason?

internal/controller/node_controller.go
```diff
@@ -392,6 +392,7 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
 	switch {
 	case err != nil:
 		log.Error(err, "Failed to mark bootstrap completed", "node", nodeName, "rule", ruleName, "uid", ruleUID)
+		metrics.BootstrapCompletionErrors.WithLabelValues(ruleName, "AnnotationPatchFailed").Inc()
```
ajaysundark: I'd suggest us to treat 'reasons' as contracts, and move from free form literals to constants. When users start using these in promql scripts and it shouldn't fail when accidentally changed later.

rawadhossain:
> Could `node_readiness_bootstrap_completion_errors_total` be folded within `failures_total` with a dedicated reason?

Yeah, agreed. Folded it into failures_total with a dedicated reason, so bootstrap_completion_errors_total has been removed.

internal/controller/node_controller.go
rawadhossain: Good point. Added a FailureReason type and moved the new reason to a constant. I'll open a follow up PR to make the existing reasons constants too.

internal/metrics/metrics.go
```diff
@@ -152,6 +153,15 @@ var (
 		},
 		[]string{"rule"},
 	)
+
+	// SelectorMatchedNodes tracks how many nodes match a rule's NodeSelector.
+	SelectorMatchedNodes = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "node_readiness_selector_matched_nodes_total",
```
AvineshTripathi: we will have to rename metric acc to the doc https://github.com/kubernetes-sigs/node-readiness-controller/pull/344/changes#diff-007dbcdc7a3ebbe3d56dc086447e169ac6f5c07f67fc0240eb500c0ca7ea430eR49

rawadhossain: yes, updated it now. Thanks.

ajaysundark:
Hi @rawadhossain can you update the description to reflect the changes?

rawadhossain:
updated the description. Thanks.

AvineshTripathi:
/lgtm

/assign @ajaysundark 
for review and approve

internal/metrics/metrics.go
```diff
@@ -152,6 +153,15 @@ var (
 		},
 		[]string{"rule"},
 	)
+
+	// RuleMatchedNodes tracks how many nodes match each rule's selector.
+	RuleMatchedNodes = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "node_readiness_rule_matched_nodes",
```
ajaysundark: We discussed this in design about having this behind the scrape-collector. I see you have a PR (https://github.com/kubernetes-sigs/node-readiness-controller/pull/398) open for that. Do you want to merge this and migrate later or you want to handle this after handling #398 first?

rawadhossain: Yes, that was the plan to get this merged first. Since this PR was raised before we introduced the scrape-time collector, I didn’t want to discard it and block the current work.

I’ll do a follow-up migration PR once this and the collector PRs are merged.

AvineshTripathi:
@rawadhossain can you please resolve the conflicts?

rawadhossain:
@AvineshTripathi resolved the conflicts. GitHub was down for quite a while 😥
PTAL.

rawadhossain:
@ajaysundark if everything looks right here, can we get it merged? 
Will do the collector migration for `rule_matched_nodes` as a follow-up PR once this land.

internal/metrics/metrics.go
ajaysundark: Now that scrape-time collector is already in, would it make sense to make these changes right now than to plan a followup?

rawadhossain: Yeah, I was thinking the same. I think it'd be better to do it now, and it'll be easier for you to review as well. I'll try to get the changes done today. Thanks.

rawadhossain:
@ajaysundark made the changes based on scrape time collector and updated the description.

Could we get #431 `(blocked_nodes)` reviewed and merged first? This PR has a small adjustment that depends on the changes in #431. Once it's merged, I'll make that change here and push the final commit.

rawadhossain:
@ajaysundark made the adjustments here after 431 landed. I think this should be good to merge. 
PTAL. Thanks!

internal/metrics/collector.go
```diff
@@ -133,4 +147,13 @@ func (c *ReadinessCollector) Collect(ch chan<- prometheus.Metric) {
 			}
 		}
 	}
+
+	matched, err := c.lister.ListRuleMatchedNodes(ctx, nodes, rules)
```
ajaysundark, line 151: @rawadhossain Could you check if we are scraping these metrics as a single-pass instead of scraping the cache individually for each of these scrape time metrics?

rawadhossain: Yeah, checked it. All three scrape time metrics scrape the node and rule cache only once, not three individual cache reads. Also tested it on a live cluster to double check, and everything matched up correctly.

PR #287, opened 2026-07-03
add options to tune concurrency, qps and burst

ajaysundark:
## Description
Add flags to the controller to tune controller-runtime knobs for scalability

## Related Issue
None

## Type of Change
/kind feature

## Testing
Manual test runs

## Checklist
- [X] `make test` passes
- [X] `make lint` passes

Yes, adds command-flags to tune concurrency, qps and burst -- https://cluster-api.sigs.k8s.io/developer/core/tuning#runtime-tuning-options

Relevant for our ongoing scalability work.

```release-note
add new commandline-flags for tuning concurrency, qps and burst limits
```

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -67,6 +67,10 @@ func main() {
 	var metricsCertDir string
 	var leaderElectionNamespace string
 	var enableNodeStateMetrics bool
+	var kubeAPIQPS float64
+	var kubeAPIBurst int
+	var nodeConcurrentReconciles int
+	var ruleConcurrentReconciles int
 
 	flag.StringVar(&metricsAddr, "metrics-bind-address", "0", "The address the metrics endpoint binds to. "+
 		"Use :8443 for HTTPS or :8080 for HTTP, or leave as 0 to disable the metrics service.")
@@ -84,6 +88,16 @@ func main() {
 	flag.StringVar(&leaderElectionNamespace, "leader-election-namespace", "", "The namespace where the leader election resource will be created.")
 	flag.BoolVar(&enableNodeStateMetrics, "enable-node-state-metrics", false,
 		"Enable aggregate node state metrics on node updates)")
+	flag.Float64Var(&kubeAPIQPS, "kube-api-qps", 20,
+		"Maximum queries per second to the API server from this client. "+
+			"Raise together with --kube-api-burst on large clusters.")
+	flag.IntVar(&kubeAPIBurst, "kube-api-burst", 30,
+		"Maximum burst for throttle between requests to the API server.")
+	flag.IntVar(&nodeConcurrentReconciles, "node-concurrent-reconciles", 1,
+		"Maximum number of Node objects reconciled concurrently. "+
+			"Raise on large clusters to reduce readiness-taint latency during node join/condition updates.")
+	flag.IntVar(&ruleConcurrentReconciles, "rule-concurrent-reconciles", 1,
+		"Maximum number of NodeReadinessRule objects reconciled concurrently.")
 
 	opts := zap.Options{
 		Development:     true,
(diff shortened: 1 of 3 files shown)


cmd/main.go
```diff
@@ -67,6 +67,10 @@ func main() {
 	var metricsCertDir string
 	var leaderElectionNamespace string
 	var enableNodeStateMetrics bool
+	var kubeAPIQPS float64
```
Karthik-K-N: Given that these vars are increasing, I think we can keep them in vars() block outside main for better code organization?

cmd/main.go
```diff
@@ -84,6 +88,16 @@ func main() {
 	flag.StringVar(&leaderElectionNamespace, "leader-election-namespace", "", "The namespace where the leader election resource will be created.")
 	flag.BoolVar(&enableNodeStateMetrics, "enable-node-state-metrics", false,
 		"Enable aggregate node state metrics on node updates)")
+	flag.Float64Var(&kubeAPIQPS, "kube-api-qps", 20,
+		"Maximum queries per second to the API server from this client. "+
+			"Raise together with --kube-api-burst on large clusters.")
+	flag.IntVar(&kubeAPIBurst, "kube-api-burst", 30,
+		"Maximum burst for throttle between requests to the API server.")
```
Karthik-K-N: isn't this makes it simpler  "maximum number of queries that should be allowed in one burst"

cmd/main.go
```diff
@@ -107,7 +121,11 @@ func main() {
 		}(),
 	}
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	restConfig := ctrl.GetConfigOrDie()
+	restConfig.QPS = float32(kubeAPIQPS)
```
Karthik-K-N, line 132: why can't we declare it directly as float32?

cmd/main.go
Karthik-K-N: is there any specific reason for keeping 20 and 30 or is it just to get started.

ajaysundark: These are the current defaults set in controller runtime. I think we could adjust this after our experiments. Maybe  I could add a comment to clarify these random numbers here

Karthik-K-N: Yeah, That would help, I looked at the default and saw its 5 and 10 https://github.com/kubernetes/client-go/blob/f16383b964b3519812bac4daf8f48fc5a529ae0f/rest/config.go#L118-L127, I am I looking at wrong place?

ajaysundark: hmm.. I referred kcm default config here: https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager.

looks like the controller-runtime had these hardcoded as defaults until recently: https://github.com/kubernetes-sigs/controller-runtime/commit/ab40409635dc0c91781f575a87c64a579680be95

vitorfloriano: Regarding the magic numbers, could we have them as constants?

vitorfloriano:
I took this branch for a spin and verified that the tuning works. 

I ran a manual test with a 1000-node kwok cluster without any parameters and NRC added all the taints in about 2m55s:

image

---

The same manual test with `--node-concurrent-reconciles 20 --kube-api-burst 1000 --kube-api-qps 50`** is performed in 1m21s (about half the time w/o any tuning):

image

I'm still learning how to query Prometheus, but I think I got this right. 

**Just picked these numbers randomly.

cmd/main.go
ajaysundark: yeah makes sense. updated to move them to a vars block

cmd/main.go
ajaysundark: flag only provides a Float64Var for parsing, so casting to float32 is unavoidable as QPS is float32: https://pkg.go.dev/k8s.io/client-go/rest#Config

ajaysundark:
@Karthik-K-N @vitorfloriano addressed your comments, ptal when you are available.

PR #288, opened 2026-07-03
feat: Add reconciliation retry observability metrics

rawadhossain:
## Description
This PR adds two new metrics

### `node_readiness_api_conflicts_total`

Tracks how many API conflicts occur during controller operations. Previously, only conflicts that exhausted all retries were visible through `node_readiness_failures_total`. If a retry succeeded, there was no signal that a conflict had happened. This metric makes those retries visible and helps identify API server contention earlier.

### `node_readiness_reconcile_requeue_total`
Tracks why a rule reconciliation is requeued. Right now, we know a rule is being requeued, but we don't know which stage caused it. This metric records the reason for the requeue, making it easier to identify where reconciliation is failing.

The metric is labeled by `reason`, currently covering:
- `status_update_error`
- `cleanup_error`
- `taint_cleanup_error`

## Type of Change
/kind feature

## Testing
- Added tests coverages.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -155,12 +155,14 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	// Update rule status
 	if err := r.Controller.updateRuleStatus(ctx, rule); err != nil {
 		log.Error(err, "Failed to update rule status", "rule", rule.Name)
+		metrics.ReconcileRequeue.WithLabelValues(rule.Name, "status_update_error").Inc()
 		return ctrl.Result{RequeueAfter: time.Minute}, err
 	}
 
 	// Clean up status for deleted nodes
 	if err := r.Controller.cleanupDeletedNodes(ctx, rule, nodeList); err != nil {
 		log.Error(err, "Failed to clean up deleted nodes", "rule", rule.Name)
+		metrics.ReconcileRequeue.WithLabelValues(rule.Name, "cleanup_error").Inc()
 		return ctrl.Result{RequeueAfter: time.Minute}, err
 	}
 
@@ -188,6 +190,7 @@ func (r *RuleReconciler) reconcileDelete(ctx context.Context, rule *readinessv1a
 	log.Info("Cleaning up taints for deleted rule", "rule", rule.Name)
 	if err := r.Controller.cleanupTaintsForRule(ctx, rule, nodeList); err != nil {
 		log.Error(err, "Failed to cleanup taints for rule", "rule", rule.Name)
+		metrics.ReconcileRequeue.WithLabelValues(rule.Name, "taint_cleanup_error").Inc()
 		return ctrl.Result{RequeueAfter: time.Minute}, err
 	}
 
@@ -216,6 +219,8 @@ func (r *RuleReconciler) reconcileDelete(ctx context.Context, rule *readinessv1a
 	metrics.ConditionEvaluationFailures.DeletePartialMatch(ruleLabel)
 	metrics.TaintOperations.DeletePartialMatch(ruleLabel)
 	metrics.ReconciliationLatency.DeletePartialMatch(ruleLabel)
+	metrics.APIConflicts.DeletePartialMatch(ruleLabel)
+	metrics.ReconcileRequeue.DeletePartialMatch(ruleLabel)
 
 	return ctrl.Result{}, nil
 }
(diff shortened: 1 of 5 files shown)


internal/controller/nodereadinessrule_controller.go
```diff
@@ -155,12 +155,14 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	// Update rule status
 	if err := r.Controller.updateRuleStatus(ctx, rule); err != nil {
 		log.Error(err, "Failed to update rule status", "rule", rule.Name)
+		metrics.ReconcileRequeue.WithLabelValues(rule.Name, "status_update_error").Inc()
```
AvineshTripathi, line 158: there are other places too in `Reconcile` where we requeue for e.g. https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/nodereadinessrule_controller.go#L151

maybe a defer function would make more sense here. Thoughts?

AvineshTripathi (commented):
I did some initial review and I had some questions:

- Are we planning to visualize these metrics? If so, then how?
- What type of answer will an operator get from these? I feel these are metrics for the controller itself, which will tell how controller is behaving in concurrent env

rawadhossain:
> I did some initial review and I had some questions:
> 
> * Are we planning to visualize these metrics? If so, then how?
> * What type of answer will an operator get from these? I feel these are metrics for the controller itself, which will tell how controller is behaving in concurrent env

That's a fair point. Both metrics are meant to land in Grafana as  alert companions  along with the existing SLOs.

It may feel a bit controller centric, but here's how I was looking at it.

`api_conflicts_total`, gives the operator an early signal that the controller is starting to experience API server write contention. Right now we only know about it once retries are exhausted and an operation fails. This metric makes those retries visible earlier. To me, that felt useful as the operator gets a chance to notice the issue before it starts affecting the controller.

For `reconcile_requeue_total`, I think you're right. It mainly tells us which stage inside the reconciler is causing the requeue, which is probably more useful for controller than operator. I'm okay dropping this metric from the PR and revisit later if we feel there's any need.

internal/controller/nodereadinessrule_controller.go
rawadhossain: I looked into using a `defer`, but I don't think it simplifies things much. We'd still need to set the `reason` at each requeue site so the deferred function knows which label to increment. It would also require named returns, handling early exits before `rule` is available, and a separate `defer` in `reconcileDelete()` for `taint_cleanup_error`, so I'm not sure if a defer would help much here.

For the line you linked (L151), we actually left that one out on purpose, `processAllNodesForRule()` always returns `nil`, so that error path is never reached. But `updateRuleStatus()`, can actually fail and return a real error, so I felt that was worth instrumenting.

AvineshTripathi: I agree but in past we have add a discussion to allow processAllNodesForRule to return error. Also there are other places too like https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/nodereadinessrule_controller.go#L122

AvineshTripathi:
I dont mind having these metrics but not sure if they should go on any dashboards. wdyt @ajaysundark

rawadhossain:
Hey @AvineshTripathi @ajaysundark, just to close the loop on this
`api_conflicts_total` does have a dashboard panel (API Conflict Rate by Operation) in the Rule Bottleneck Ranking section. The idea is to make API conflicts visible before they become a problem. 

I'm okay to drop `reconcile_requeue_total` from this PR though and revisit it later if there's a stronger operator use case.

AvineshTripathi:
From the last meeting, did we see the possibility of api_error metrics going as a log and not metrics? Intuitively I too feel being a log makes more sense

rawadhossain:
Hi @AvineshTripathi, you're right and I dug into this properly after the meeting discussion.

I also looked into moving it fully to logs, but the project has no log aggregation setup. 
So went with a different approach, I updated the existing `node_readiness_failures_total` metric with conflict specific reason labels. While doing this, I also found and fixed a missing failure metric in the status patch path.

I've raised PR [#320](https://github.com/kubernetes-sigs/node-readiness-controller/pull/320) with this approach, so I'll close this PR.

internal/metrics/metrics.go
```diff
@@ -117,6 +117,27 @@ var (
...
+				"Increments per conflict, not per exhaustion (see node_readiness_failures_total for that). " +
+				"operation: add_taint|remove_taint|mark_bootstrap|process_node|cleanup_nodes|update_status",
+		},
+		[]string{"rule", "operation"},
+	)
+
+	// Counts error driven requeues in RuleReconciler, keyed by the failing stage.
+	ReconcileRequeue = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "node_readiness_reconcile_requeue_total",
```
ajaysundark, line 135: Controller-runtime already exposes requeue metrics: https://github.com/kubernetes-sigs/controller-runtime/blob/6ec95076a056c7a36c0d8acc8316da796426df7e/pkg/internal/controller/metrics/metrics.go#L39. We dont need to duplicate them.

ajaysundark:
/close

superseded by #320

PR #291, opened 2026-07-05
feat: prevent setting defaultStatus in bootstrap-only enforcement mode within validation webhook

### LINKED ISSUE #290: [FEATURE] Block the combination of DefaultStatus and EnforcementMode="bootstrap-only"

This is an enhancement/safety-protocol on top of the newly added DefaultStatus feature #282. Under all the scenarios, we could imagine, there was not a single one where it ever felt like a user might benefit by processing a missing node condition as a matching/satisfying requirement.

In-reality, this combination can actually violate what a bootstrap-only mode is meant for. Suppose you match a Condition during bootstrap with the help of a DefaultStatus. Now it completes. And guess what? Because the node was never tainted, it is forever stuck in a loop without marking the bootstrap as completed. 

Or in the worst case, there can be a scenario there are multiple conditions of which one didn't satisfy and node was tainted. Now the condition where we applied default was incomplete but because every condition including this one has satisfied + the node was tainted, it'll mark the bootstrap process as completed. This also violates the purpose of having a bootstrap only mode.

A potential solution revolves around adding a web hook validation to reject any possible combination of presence of a DefaultStatus (which is an optional field) and EnforcementMode "bootstrap-only". 

There can be another way to go for blocking of combination DefaultStatus != "Unknown" and EnforcementMode="bootstrap-only". However from a user-perspective why should you be allowed to set limited values of DefaultStatus with a mode that gains nothing from that extra feature.

If a need arrives for use DefaultStatus in future, we may have to come back to this point.

AnuragThePathak:
## Description

Adds webhook validation to prevent defaultStatus from being set on any ConditionRequirement when enforcementMode is bootstrap-only.

There was no valid use case found for this combination. defaultStatus is useful when a condition may never appear on a node — but bootstrap-only mode exists precisely to wait for conditions to be explicitly reported before completing the bootstrap gate. The two features work against each other: depending on the configuration, using them together either lets the bootstrap gate pass for a condition that was never actually verified on the node, or prevents the bootstrap from completing indefinitely.

The check runs only on CREATE — updates are excluded via an early return since enforcementMode and conditions are effectively immutable after creation (enforced at the CEL/CRD level). Also updated the schema description with this information.

## Related Issue

Fixes #290 

## Type of Change
/kind feature

## Testing

- Unit tests (validateSpec): three isolated tests covering the early return on update, the continuous-mode skip, and the single-condition forbidden error — all using a shared BeforeEach spec
- Accumulation test: standalone test verifying that an empty nodeSelector + two conditions each with defaultStatus on bootstrap-only produces all 3 errors simultaneously (no short-circuit)
- Integration test (validateNodeReadinessRule): end-to-end coverage of the new check through the full validation stack inside Full Validation Integration

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Setting defaultStatus on a condition is now forbidden when enforcementMode is "bootstrap-only". The webhook will reject such configurations at creation time with a descriptive error on the offending spec.conditions[i].defaultStatus field.
```
## Documentation

docs/book/src/reference/api-spec.md: Updated the defaultStatus field description in the API reference table to include the same constraint
docs/book/src/user-guide/getting-started.md: Added "Must not be set in bootstrap-only mode" to the defaultStatus field description
docs/book/src/user-guide/concepts.md: Updated the callout from a WARNING to IMPORTANT and rewrote the explanation — from "do not use" guidance to a clear statement that the combination is unsupported and rejected by the admission webhook, with a concise explanation of why the two features work against each other

### DIFF
--- a/docs/book/src/user-guide/concepts.md
+++ b/docs/book/src/user-guide/concepts.md
@@ -43,9 +43,15 @@ The choice of `defaultStatus` has critical interaction with the rule's `enforcem
 > [!IMPORTANT]
 > **`defaultStatus` is not supported with `bootstrap-only` rules.**
 >
-> There is no valid use case found so far for this combination. `defaultStatus` is useful when a condition may never appear on a node — but `bootstrap-only` mode exists precisely to *wait* for conditions to be explicitly reported before completing the bootstrap gate. The two features work against each other.
+> There is no valid use case found so far for this combination. `defaultStatus` is useful when a condition may never appear on a node in its healthy state — effectively treating its absence as a known-good signal. But `bootstrap-only` mode exists precisely to *wait* for conditions to be explicitly reported before completing the bootstrap gate. The two features work against each other.
 >
-> Depending on your configuration, using them together either prevents the bootstrap from completing indefinitely (because all the conditions have always been satisfied) or lets the bootstrap gate pass for a condition that was never actually verified on the node. The admission webhook rejects any such combination.
+> Depending on your configuration, combining them breaks the state machine in one of two ways:
+>
+> **Premature Completion:** The absent condition evaluates as satisfied via `defaultStatus`, allowing the bootstrap gate to pass without the condition ever being actually verified.
+>
+> **Infinite Hang:** If the node was never registered with the taint ([as prescribed here](./getting-started.md#4-configure-the-taint)), `defaultStatus` keeps the condition satisfied, meaning no taint removal will ever trigger to mark the bootstrap as complete.
+>
+> To prevent these invalid states, the admission webhook explicitly rejects this combination.
 
 ## Enforcement Modes
 
--- a/docs/book/src/user-guide/concepts.md
+++ b/docs/book/src/user-guide/concepts.md
@@ -43,15 +43,9 @@ The choice of `defaultStatus` has critical interaction with the rule's `enforcem
 > [!IMPORTANT]
 > **`defaultStatus` is not supported with `bootstrap-only` rules.**
 >
-> There is no valid use case found so far for this combination. `defaultStatus` is useful when a condition may never appear on a node in its healthy state — effectively treating its absence as a known-good signal. But `bootstrap-only` mode exists precisely to *wait* for conditions to be explicitly reported before completing the bootstrap gate. The two features work against each other.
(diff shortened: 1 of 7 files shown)


internal/webhook/nodereadinessgaterule_webhook.go
```diff
@@ -70,6 +73,22 @@ func (w *NodeReadinessRuleWebhook) validateSpec(spec readinessv1alpha1.NodeReadi
 		allErrs = append(allErrs, field.Required(field.NewPath("spec", "nodeSelector"), "nodeSelector must not be empty"))
 	}
 
+	// skip the next checks for update
```
ajaysundark:
```suggestion
	// skip below checks for update because both `enforcementMode` and `conditions` 
	// are immutable as constrained by CEL XValidation rules
```

docs/book/src/user-guide/concepts.md
```diff
@@ -40,12 +40,18 @@ By setting `requiredStatus: False` and `defaultStatus: False`, an absent `Mainte
...
+>
+> There is no valid use case found so far for this combination. `defaultStatus` is useful when a condition may never appear on a node in its healthy state — effectively treating its absence as a known-good signal. But `bootstrap-only` mode exists precisely to *wait* for conditions to be explicitly reported before completing the bootstrap gate. The two features work against each other.
+>
+> Depending on your configuration, combining them breaks the state machine in one of two ways:
+>
+> **Premature Completion:** The absent condition evaluates as satisfied via `defaultStatus`, allowing the bootstrap gate to pass without the condition ever being actually verified.
+>
+> **Infinite Hang:** If the node was never registered with the taint ([as prescribed here](./getting-started.md#4-configure-the-taint)), `defaultStatus` keeps the condition satisfied, meaning no taint removal will ever trigger to mark the bootstrap as complete.
+>
+> To prevent these invalid states, the admission webhook explicitly rejects this combination.
```
ajaysundark:
```suggestion
> `defaultStatus` is useful when a condition may never appear on a node in its healthy state, effectively treating its absence as a known-good signal. However, `bootstrap-only` mode exists precisely to *wait* for conditions to be explicitly reported before completing the bootstrap gate.
>
 > Because these two features serve opposing purposes, using them together can lead to unintended behavior, such as completing the bootstrap phase before a condition is actually verified. To prevent  this, the admission webhook explicitly rejects this combination.
```

ajaysundark (changes requested):
lgtm, just some doc feedback. Could you handle these updates?

ajaysundark (approved):
/lgtm

PR #293, opened 2026-07-07
docs: add Problem-Gate Readiness example using defaultStatus

### LINKED ISSUE #289: [Documentation] Add a Problem-gate example using the newly introduced defaultStatus
We should also consider documenting this use-case example for the site.

_Originally posted by @ajaysundark in https://github.com/kubernetes-sigs/node-readiness-controller/pull/283#discussion_r3518707726_
            
The idea behind creating this issue is to add the test case there as a new example to the set of examples present in the NRC documentation.

We can use a format similar to what security agent or CNI installation does and fill the data with reference to this test case.

Original issue for the feature itself: #282

AZERDSQ131:
## Description

Adds a new worked example demonstrating the `defaultStatus` field introduced in #283, as requested in #289.

The example uses the exact scenario already covered by the integration test in `internal/controller/nodereadinessrule_controller_test.go` ("should satisfy a rule when an absent condition matches via defaultStatus (problem-gate scenario)") and documented conceptually in `docs/book/src/user-guide/concepts.md` ("Use Case: Problem Gating (Default-Allow)"): an NPD-style `MaintenanceRequired` condition that should default to healthy when absent, so a freshly bootstrapped node isn't gated by the reporter's startup race, following the same format as the existing CNI/Security Agent examples.

Adds:
- `examples/problem-gate-readiness/`: NPD custom-plugin ConfigMap + DaemonSet + RBAC reporting `MaintenanceRequired`, and a `NodeReadinessRule` with `requiredStatus: "False"` / `defaultStatus: "False"` / `enforcementMode: "continuous"`, plus a README with apply/verify steps.
- `docs/book/src/examples/problem-gate-readiness.md`: the full guide, linked from `SUMMARY.md` under "Examples".

Docs-only change; no Go code, CRD, or API types touched. YAML manifests validated with `yaml.safe_load_all`.

## Related Issue

Fixes #289

## Type of Change

/kind documentation

### DIFF
--- a/docs/book/src/examples/problem-gate-readiness.md
+++ b/docs/book/src/examples/problem-gate-readiness.md
@@ -1,0 +1,141 @@
+# Problem-Gate Readiness (`defaultStatus`)
+
+This guide demonstrates how to use the `defaultStatus` field on a
+`ConditionRequirement` to build a **problem gate**: a rule guarded by a
+problem-oriented Node condition — reported in the style of
+[Node Problem Detector (NPD)](https://github.com/kubernetes/node-problem-detector)
+— that should default to healthy when the condition hasn't been reported yet,
+rather than blocking the node.
+
+## The Problem
+
+Problem-oriented conditions (like `MemoryPressure`, `DiskPressure`, or a
+custom `MaintenanceRequired` condition reported by NPD) are only ever written
+to a Node's status *when a problem is detected*. On a freshly bootstrapped
+node, NPD may not have run yet, so the condition is completely absent from
+`status.conditions` — not `False`, just missing.
+
+By default, the Node Readiness Controller resolves an absent condition to
+`Unknown`. If a rule requires `requiredStatus: "False"` on that condition,
+`Unknown` does not satisfy it, and the node stays tainted until NPD reports
+the condition at least once. For a problem gate, that's backwards: the
+absence of a problem report should mean "no known problem," not "blocked."
+
+## The Solution
+
+Set `defaultStatus: "False"` alongside `requiredStatus: "False"` on the
+guarded condition:
+
+- If the condition is **absent**, it evaluates to `False` (the configured
+  default) — satisfying `requiredStatus: "False"` — so the node is not
+  gated by the reporter's startup race.
+- If NPD later reports the condition as `True` (a problem was detected), the
+  rule is no longer satisfied and the taint is (re-)applied.
+- The controller still reports the *observed* status (`Unknown` when
+  absent) separately in `status.nodeEvaluations[].conditionResults[].currentStatus`,
+  so you retain full visibility into whether a value came from the Node or
+  from the configured default.
+
+See [Core Concepts: Default Condition Status](../user-guide/concepts.md#default-condition-status-defaultstatus)
+for the full semantics, including why `defaultStatus` is rejected by the
+validating webhook when `enforcementMode` is `bootstrap-only`.
+
+> [!NOTE]
+> All manifests referenced in this guide are available in the
+> [`examples/problem-gate-readiness`](https://github.com/kubernetes-sigs/node-readiness-controller/tree/main/examples/problem-gate-readiness)
+> directory.
+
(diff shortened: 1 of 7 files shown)


docs/book/src/examples/problem-gate-readiness.md
```diff
@@ -0,0 +1,141 @@
+# Problem-Gate Readiness (`defaultStatus`)
```
AnuragThePathak, line 1: If you see the titles of rest of the examples, including defaultStatus in a title wasn't a best idea perhaps. Also "problem-gate readiness" indicates something totally different than what we are doing here. Something simpler like problem gateway is a better choice. Feel free to include default allow or something extra. It's good to be innovative.

docs/book/src/examples/problem-gate-readiness.md
```diff
@@ -0,0 +1,141 @@
...
+- If NPD later reports the condition as `True` (a problem was detected), the
+  rule is no longer satisfied and the taint is (re-)applied.
+- The controller still reports the *observed* status (`Unknown` when
+  absent) separately in `status.nodeEvaluations[].conditionResults[].currentStatus`,
+  so you retain full visibility into whether a value came from the Node or
+  from the configured default.
+
+See [Core Concepts: Default Condition Status](../user-guide/concepts.md#default-condition-status-defaultstatus)
+for the full semantics, including why `defaultStatus` is rejected by the
+validating webhook when `enforcementMode` is `bootstrap-only`.
```
AnuragThePathak, line 41: inclusion of bootstrap-only mode looks very random to me. I would suggest in favor of simply avoiding it as Problem-gates are likely to work continuously.

examples/problem-gate-readiness/README.md
AnuragThePathak, line 1: I see that this file is present in CNI readiness example too but that's not the case in other too examples. Probably wait for for other reviews, I just wanted to highlight it.

But if that happens then please update the Problem-gate readiness title here as well.

AnuragThePathak (commented):
Didn't verify the yaml files, rest looks good.

docs/book/src/examples/problem-gate-readiness.md
```diff
@@ -0,0 +1,141 @@
...
+This guide demonstrates how to use the `defaultStatus` field on a
+`ConditionRequirement` to build a **problem gate**: a rule guarded by a
+problem-oriented Node condition — reported in the style of
+[Node Problem Detector (NPD)](https://github.com/kubernetes/node-problem-detector)
+— that should default to healthy when the condition hasn't been reported yet,
+rather than blocking the node.
+
+## The Problem
+
+Problem-oriented conditions (like `MemoryPressure`, `DiskPressure`, or a
```
ajaysundark, line 12: These conditions are not relevant for NRC as these have special meaning in Kubernetes, and control-plane / node life-cycle controller will react to these.

docs/book/src/examples/problem-gate-readiness.md
```diff
@@ -0,0 +1,141 @@
...
+custom `MaintenanceRequired` condition reported by NPD) are only ever written
+to a Node's status *when a problem is detected*. On a freshly bootstrapped
+node, NPD may not have run yet, so the condition is completely absent from
+`status.conditions` — not `False`, just missing.
```
ajaysundark, line 16: I'd prefer to not mention this as specific to NPD not-run situation. The feature is for addressing NRC for 'problem-gating', which is a different use-case pattern than 'readiness-gate'. The reporter can be NPD or anything else.

ajaysundark (commented):
@AZERDSQ131 Thanks for your PR!

I think it'd be a good idea to write this tied to a use-case than a generic flow.

One option I came across is https://github.com/NVIDIA/NVSentinel/blob/main/demos/local-fault-injection-demo/README.md to simulate 'gpu error condition' in a Kind cluster. Is that something you are available to explore?

ajaysundark:
/close 

Duplicate to #317

PR #294, opened 2026-07-08
feat: Add bootstrap NRC duration metric

rawadhossain:
## Description

This PR adds `node_readiness_bootstrap_hold_duration_seconds`, which measures how long a node is held by a bootstrap-only rule's taint, from when the taint is first present on the node until bootstrap completes.

The status object now tracks two anchor timestamps on `NodeEvaluation`:

- `taintAppliedAt` — when NRC itself applies the taint
- `taintObservedAt` — when NRC first observes the taint present, whether applied by NRC or already present (e.g. via kubelet `-register-with-taints`)

The metric carries a new `controller` label (`applied`/`adopted`) so the two cases are distinguishable: anchored on `taintAppliedAt` when NRC applied the taint, falling back to `taintObservedAt` for the adopted case, where `taintAppliedAt` is never set.

## Type of Change

/kind feature

## Testing

- Unit tests for `isBootstrapCompleted`, including backward compatibility and in progress states
- Integration tests added

## Checklist

- [x]  `make test` passes
- [x]  `make lint` passes


internal/controller/helper.go
```diff
@@ -29,10 +30,17 @@ const (
...
 	// Full key format: readiness.k8s.io/bootstrap-completed-<ruleUID>
-	// Value format:    {"rule-name":"<ruleName>"}   (for human readability)
+	// Value format:    {"rule-name":"<ruleName>"}   (for human readability).
 	bootstrapAnnotationPrefix = "readiness.k8s.io/bootstrap-completed-"
 )
 
+// bootstrapAnnotationPayload is the JSON value stored in a bootstrap-completion annotation.
+type bootstrapAnnotationPayload struct {
+	RuleName       string `json:"rule-name"`
+	TaintAppliedAt string `json:"taintAppliedAt,omitempty"`
```
ajaysundark: What if the taint was not applied by NRC, rather it adopted the taints added by kubelet (via `--register-with-taints`), which is the preferred mode of operation for bootstrap-only mode? How does this handle the case?

internal/controller/helper.go
```diff
@@ -29,10 +30,17 @@ const (
 	// ~36 chars), which is immutable for the object's lifetime and globally unique.
 	//
 	// Full key format: readiness.k8s.io/bootstrap-completed-<ruleUID>
-	// Value format:    {"rule-name":"<ruleName>"}   (for human readability)
+	// Value format:    {"rule-name":"<ruleName>"}   (for human readability).
 	bootstrapAnnotationPrefix = "readiness.k8s.io/bootstrap-completed-"
 )
 
+// bootstrapAnnotationPayload is the JSON value stored in a bootstrap-completion annotation.
+type bootstrapAnnotationPayload struct {
```
ajaysundark, line 39: IMO, TaintAppliedAt and Completed are better suited for the dedicated NodeEvaluation 'status' api we planned. Can we discuss this during our next sync on how we could shape it in the alpha2 api?

AvineshTripathi:
@rawadhossain can you resolve the conflicts please!

rawadhossain:
Hi @ajaysundark @AvineshTripathi  
Sorry for the late response.

Updated the PR. Moved the timestamp from node metadata to status.nodeEvaluations, following the current API shape as we discussed. Also handled the adopted taint (--register-with-taints) case by only stamping the anchor when the taint is applied by NRC.

rawadhossain:
While validating this metric for [Section 2 Grafana dashboard](https://docs.google.com/document/d/1K13io8xVz4TW6Q9RLCXVwogqfa4RURsOph1k-KuzgoE/edit?tab=t.dt5ewq11ayqn), I found that it wasn't emitting any data. It turned out `TaintAppliedAt` in `status.nodeEvaluations` exposed an issue with stale cached rules in `NodeReconciler`, which could cause the anchor to be lost during the next status update.

I added a fix to recover the missing anchor when needed and make sure it isn't lost again during later status updates. Since the Section 2 dashboard depends on this metric, I wanted to fix this before moving forward.

cc @ajaysundark @AvineshTripathi

internal/metrics/metrics.go
```diff
@@ -77,6 +77,17 @@ var (
 		[]string{"rule"},
 	)
 
+	// BootstrapNRCDuration tracks the time from NRC taint application to bootstrap completion.
+	// Measures NRC hold time only; skipped when no taintAppliedAt anchor.
+	BootstrapNRCDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "node_readiness_bootstrap_nrc_duration_seconds",
+			Help:    "Time from NRC taint application to bootstrap completion. Measures only NRC-attributable hold time and excludes pre-NRC node boot time.",
+			Buckets: []float64{1, 5, 10, 30, 60, 120, 300, 600, 1200},
```
ajaysundark: extend this to 1800/3600. I'd suggest the same for `BootstrapDuration` above

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -246,6 +246,11 @@ type NodeEvaluation struct {
 	//
 	// +required
 	LastEvaluationTime metav1.Time `json:"lastEvaluationTime,omitempty,omitzero"`
+
+	// taintAppliedAt is the timestamp when the controller applied the readiness taint to this Node.
+	//
+	// +optional
+	TaintAppliedAt metav1.Time `json:"taintAppliedAt,omitempty,omitzero"`
```
ajaysundark, line 282: cc @Karthik-K-N for note as we should keep this field name and semantics same when we split NodeEvaluation status

AvineshTripathi:
@rawadhossain can we two separate PRs? Maybe the fix can go in the different PR and we can try merge this. Please let me know your thoughts

api/v1alpha1/nodereadinessrule_types.go
ajaysundark, line 282: Note that "TaintAppliedAt" is not applicable to the most-common use-case for NRC - users register the taints with kubelet api (--register-with-taints) and NRC only removes the taint. In this case, the taint is only 'adopted' by NRC, would it make sense to create a separate field like `TaintAdoptedAt`? I asked this question at https://docs.google.com/document/d/1DOP1G6i__nQN8qbhlSvdswkUquMeSKQY4LU554AbcgM/edit?tab=t.0 as well where we were discussing the API.

rawadhossain: I looked into this and ran some burst scale tests (50 and 200 concurrent nodes with pre-existing taints). Under load, there's a noticeable delay before NRC first reconciles a node (p50 ~29s at 50 nodes, ~105s at 200), so I see the concern with `TaintAppliedAt` for the `--register-with-taints` case.

I'm still not sure whether introducing a separate `TaintAdoptedAt` field would make sense, or if the existing ones are sufficient. i think i need your more thoughts on this.

(Also, I still don't have access to the doc you linked. Could you please check it?)

rawadhossain:
> @rawadhossain can we two separate PRs? Maybe the fix can go in the different PR and we can try merge this. Please let me know your thoughts

That makes sense and its cleaner. My only concern is that the cache fix is tied to this metric, and without it the metric doesn't reliably emit data. That's why I kept them together. If you prefer splitting them, I'm happy to do that.

AvineshTripathi:
@rawadhossain i see there are conflicts, can you resolve it please

rawadhossain:
Done, resolved the merge conflicts. Thanks.

api/v1alpha1/nodereadinessrule_types.go
ajaysundark: this is likely resolved, jfyi, the new API shape is here: https://github.com/kubernetes-sigs/node-readiness-controller/pull/345

internal/metrics/metrics.go
```diff
@@ -107,11 +73,22 @@ var (
...
 
+	// BootstrapHoldDuration tracks the time from the readiness taint's anchor timestamp to bootstrap completion.
+	// Measures NRC hold time only; skipped when no anchor is available.
+	BootstrapHoldDuration = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "node_readiness_bootstrap_hold_duration_seconds",
+			Help:    "Time from readiness taint application or observation to bootstrap completion. Measures only NRC-attributable hold time.",
+			Buckets: []float64{1, 5, 10, 30, 60, 120, 300, 600, 1200, 1800, 3600},
+		},
+		[]string{"rule", "controller"}, // controller: applied, adopted
```
ajaysundark: can you x-ref the design on this one? `taint_origin` for this label is better with values adopted vs controller, to show initial ownership.

rawadhossain: Updated it. taint_origin now uses controller and adopted to show the taint ownership.

ajaysundark:
please resolve conflicts

rawadhossain:
@ajaysundark resolved the conflicts. PTAL. Thanks.

AvineshTripathi:
/ok-to-test

@rawadhossain is this a dependecy on any other PR? Are we waiting for review here?

rawadhossain:
@AvineshTripathi this is ready for review. No dependency related to this one.

rawadhossain:
@ajaysundark can we land this one if everything looks good?

api/v1alpha1/nodereadinessrule_types.go
ajaysundark: cc @rawadhossain could you confirm and resolve this after you compare and verify against the #345 API shape so it's a lift and shift later when it lands?

rawadhossain: sure, i'll do it this week.

rawadhossain:
cc @ajaysundark for review. Thanks.

PR #296, opened 2026-07-10
fix: enhance uninstall target to wait for full deletion of CRDs

vitorfloriano:
## Description

This PR adds a wait step to the uninstall target in Makefile. 

Not waiting for a full deletion was causing e2e to fail when trying to create a CRD in a given step while the previous step hadn't finished uninstalling the CRD:

```
STEP: applying the bootstrap-only rule @ 07/10/26 00:41:41.226
  running: "kubectl apply -f test/e2e/testdata/bootstrap-only-rule.yaml"
  [FAILED] in [It] - /home/vitorfloriano/go/src/github.com/vitorfloriano/node-readiness-controller/test/e2e/e2e_test.go:232 @ 07/10/26 00:41:41.318

FAILED] [9.539 seconds]
Manager NodeReadinessRule [It] should handle bootstrap-only mode correctly
/home/vitorfloriano/go/src/github.com/vitorfloriano/node-readiness-controller/test/e2e/e2e_test.go:202

  [FAILED] Unexpected error:
      <*fmt.wrapError | 0xc0003c43a0>: 
      "kubectl apply -f test/e2e/testdata/bootstrap-only-rule.yaml" failed with error "Error from server (Forbidden): error when creating \"test/e2e/testdata/bootstrap-only-rule.yaml\": nodereadinessrules.readiness.node.x-k8s.io \"bootstrap-test-rule\" is forbidden: create not allowed while custom resource definition is terminating\n": exit status 1
      {
          msg: "\"kubectl apply -f test/e2e/testdata/bootstrap-only-rule.yaml\" failed with error \"Error from server (Forbidden): error when creating \\\"test/e2e/testdata/bootstrap-only-rule.yaml\\\": nodereadinessrules.readiness.node.x-k8s.io \\\"bootstrap-test-rule\\\" is forbidden: create not allowed while custom resource definition is terminating\\n\": exit status 1",
          err: <*exec.ExitError | 0xc0003c4380>{
              ProcessState: {
                  pid: 796641,
                  status: 256,
                  rusage: {
                      Utime: {Sec: 0, Usec: 91127},
                      Stime: {Sec: 0, Usec: 27276},
                      Maxrss: 47684,
                      Ixrss: 0,
                      Idrss: 0,
                      Isrss: 0,
                      Minflt: 4382,
                      Majflt: 0,
                      Nswap: 0,
                      Inblock: 0,
                      Oublock: 24,
                      Msgsnd: 0,
                      Msgrcv: 0,
                      Nsignals: 0,
[...]

PR #298, opened 2026-07-10
cleanup(e2e): reorder uninstall CRD to avoid getting stuck

vitorfloriano:
## Description

This PR swaps the order in which the e2e test uninstalls the CRD and undeploys the controller, to avoid getting stuck in a pending state, causing flakes.

See: https://node-readiness-controller.sigs.k8s.io/user-guide/installation.html?highlight=finalizer#uninstallation

## Related Issue

None

## Type of Change

/kind cleanup
/kind failing-test
/kind flake

## Testing

Ran e2e locally.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

None

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -103,12 +103,12 @@ var _ = Describe("Manager", Ordered, func() {
 			}
 		}
 
-		By("undeploying the controller-manager")
-		cmd = exec.Command("make", "undeploy")
-		_, _ = utils.Run(cmd)
-
 		By("uninstalling CRDs")
 		cmd = exec.Command("make", "uninstall")
+		_, _ = utils.Run(cmd)
+
+		By("undeploying the controller-manager")
+		cmd = exec.Command("make", "undeploy")
 		_, _ = utils.Run(cmd)
 
 		By("removing manager namespace")


ajaysundark:
Is this addressing the same issue as https://github.com/kubernetes-sigs/node-readiness-controller/pull/296, or different?

I remember @AvineshTripathi had a look at similar issue long back, where the issue was pending finalizers on the nodes?

@vitorfloriano could you help evaluate if this is release blocking?

Avinesh, can you help take a look at this!

vitorfloriano:
>Is this addressing the same issue as #296, or different?

They could be related, but they fail for different reasons.

In #296, I observed that the next step fails to create the CRD because of the terminating state of the deletion, but here I observed that uninstall gets stuck and times out. 

But yeah, probably reording the uninstall will also solve #296.

AvineshTripathi:
> But yeah, probably reording the uninstall will also solve https://github.com/kubernetes-sigs/node-readiness-controller/pull/296.

I agree! This fix is a better I feel. I donot see any side effects here

/lgtm

ajaysundark:
Thanks @vitorfloriano and @AvineshTripathi !

/approve

PR #300, opened 2026-07-11
infra(makefile): add kind to hack tools

### LINKED ISSUE #299: [FAILING TEST] e2e periodics are failing in CI
All periodic jobs are failing at the build step:

- https://testgrid.k8s.io/sig-node-node-readiness-controller#periodic-node-readiness-controller-test-e2e-main-1-33
- https://testgrid.k8s.io/sig-node-node-readiness-controller#periodic-node-readiness-controller-test-e2e-main-1-34
- https://testgrid.k8s.io/sig-node-node-readiness-controller#periodic-node-readiness-controller-test-e2e-main-1-35
- https://testgrid.k8s.io/sig-node-node-readiness-controller#periodic-node-readiness-controller-test-e2e-main-1-36

At first glance, it seems that kind needs to be installed.

We are using the kubekins-e2e image. Shouldn't that come with kind baked in?

https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/node-readiness-controller/node-readiness-controller-periodics.yaml

vitorfloriano:
## Description

This PR adds `kind` to the suite of hack tools, following the pattern already used in the Makefile for other tools.

It also changes a few targets to use the binary in `hack/tools`.

## Related Issue

Fixes #299 

## Type of Change

/kind feature
/kind failing-test

## Testing

Deleted all `kind` binaries from PATH (`local/bin` and `go/bin`), then ran `make test-e2e-kind`.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

This adds kind to the hack tools, so UX may be improved (less env setup).

```release-note
Add kind to hack tools.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,11 @@ export PATH := $(abspath $(TOOLS_BIN_DIR)):$(PATH)
 # Binaries.
 #
 KUBECTL ?= kubectl
-KIND ?= kind
+
+KIND_VER ?= v0.32.0
+KIND_BIN := kind
+KIND ?= $(abspath $(TOOLS_BIN_DIR)/$(KIND_BIN)-$(KIND_VER))
+KIND_PKG := sigs.k8s.io/kind
 
 KUSTOMIZE_VER := v5.7.0
 KUSTOMIZE_BIN := kustomize
@@ -220,19 +224,19 @@ docker-buildx: ## Build and push docker image for the manager for cross-platform
 	- $(CONTAINER_TOOL) buildx rm nrrcontroller-builder
 
 .PHONY: kind-load
-kind-load: ## Load the built image into kind cluster
+kind-load: $(KIND) ## Load the built image into kind cluster
 ifeq ($(CONTAINER_TOOL),podman)
 	@echo "Loading Podman image into kind cluster: $(KIND_CLUSTER)"
 	@echo "Saving image to temporary tar archive..."
 	@$(CONTAINER_TOOL) save -o /tmp/controller-image.tar localhost/$(IMG_PREFIX):$(IMG_TAG)
 	@echo "Loading tar archive into kind cluster..."
-	@kind load image-archive /tmp/controller-image.tar --name $(KIND_CLUSTER)
+	@$(KIND) load image-archive /tmp/controller-image.tar --name $(KIND_CLUSTER)
 	@echo "Cleaning up temporary tar archive..."
 	@rm /tmp/controller-image.tar
 	@echo "Image loaded successfully!"
 else
 	@echo "Loading Docker image into kind cluster: $(KIND_CLUSTER)"
-	@kind load docker-image $(IMG_PREFIX):$(IMG_TAG) --name $(KIND_CLUSTER)
+	@$(KIND) load docker-image $(IMG_PREFIX):$(IMG_TAG) --name $(KIND_CLUSTER)
 endif
 
 .PHONY: docker-build-reporter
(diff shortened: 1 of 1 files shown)


ajaysundark:
@vitorfloriano can we hold on to this change? The kubekins-e2e image indeed has Kind as you expected, so this likely not needed.

Also I dont think we should be going in the direction of hardcoding a k8s version on the e2e runner. We should be able to run parallel e2e tests across supported versions, so this would not work for that interest.

vitorfloriano:
>Also I dont think we should be going in the direction of hardcoding a k8s version on the e2e runner. We should be able to run parallel e2e tests across supported versions, so this would not work for that interest.

We're only pinning the kind CLI version. The K8s version is passed as env var in the Prowjob config and captured by the e2e-test hack script.

That being said: 

It seems that we are using a deprecated version of kubekins-e2e in our periodics: https://github.com/search?q=repo%3Akubernetes%2Ftest-infra%20path%3Aconfig%2Fjobs%2Fkubernetes-sigs%2Fnode-readiness-controller%2Fnode-readiness-controller-periodics.yaml%20content%3Aimage%3A&type=code

The GCR image is the deprecated one. See: https://github.com/kubernetes/test-infra/tree/master/images/kubekins-e2e#deprecated

The kubekins-e2e v2 image is available at another registry. See: https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e-v2/README.md#kubekins-e2e-v2

So, perhaps changing the image solves that?

Edit: I've seen some jobs using the v2 image and still installing kind on top, though: 

- https://github.com/kubernetes/test-infra/blob/a69f25f9161afdc3f40dce6173d892e11b073310/config/jobs/kubernetes/sig-k8s-infra/periodics.yaml#L124-L129
- https://github.com/kubernetes/test-infra/blob/a69f25f9161afdc3f40dce6173d892e11b073310/config/jobs/kubernetes/sig-testing/kubernetes-kind-ci.yaml#L607-L613

ajaysundark:
Ah okay, passing the API version makes sense.

And not depending on what's packaged on the image and just installing ourselves is better. 

Something for us to followup: it's important to document how these dependencies are injected and managed, let us try if we can find time part of our release-improvements. 

Thanks for this PR!

/lgtm

Makefile
```diff
@@ -35,7 +35,11 @@ export PATH := $(abspath $(TOOLS_BIN_DIR)):$(PATH)
 # Binaries.
 #
 KUBECTL ?= kubectl
-KIND ?= kind
+
+KIND_VER ?= v0.32.0
+KIND_BIN := kind
+KIND ?= $(abspath $(TOOLS_BIN_DIR)/$(KIND_BIN)-$(KIND_VER))
```
Priyankasaggu11929, line 41: Q: `$(KIND_BIN)-$(KIND_VER))` - why are we doing a check on the *versioned* binary path? and not just `kind`?

Priyankasaggu11929:
> 
> It seems that we are using a deprecated version of kubekins-e2e in our periodics: https://github.com/search?q=repo%3Akubernetes%2Ftest-infra%20path%3Aconfig%2Fjobs%2Fkubernetes-sigs%2Fnode-readiness-controller%2Fnode-readiness-controller-periodics.yaml%20content%3Aimage%3A&type=code
> 
> The GCR image is the deprecated one. See: https://github.com/kubernetes/test-infra/tree/master/images/kubekins-e2e#deprecated
> 
> The kubekins-e2e v2 image is available at another registry. See: https://github.com/kubernetes/test-infra/blob/master/images/kubekins-e2e-v2/README.md#kubekins-e2e-v2


created PR https://github.com/kubernetes/test-infra/pull/37447 to address

 
> So, perhaps changing the image solves that?

~sorry, i didn't catch what problem we're trying to solve?~

my mistake. checked https://github.com/kubernetes-sigs/node-readiness-controller/issues/299

Priyankasaggu11929:
Also, thank you folks for checking/fixing it. :pray: 
I truly missed checking testgrid after adding the jobs initially.

@ajaysundark, did you get emails for these failing jobs earlier? just to confirm if testgrid email alerting on NRC job failing is working fine

ajaysundark:
> did you get emails for these failing jobs earlier? just to confirm if testgrid email alerting on NRC job failing

Thanks @Priyankasaggu11929 that's what I wanted to discuss too. I don't get any emails. We should ensure we are observing these signals and set a process to follow up on this when there are failures.

Priyankasaggu11929:
> > did you get emails for these failing jobs earlier? just to confirm if testgrid email alerting on NRC job failing
> 
> Thanks @Priyankasaggu11929 that's what I wanted to discuss too. I don't get any emails. We should ensure we are observing these signals and set a process to follow up on this when there are failures.

Thanks for the confirmation. 

I am assuming that happened because the job just failed since inception (and email alert is triggered when the job status (Failing/passing) changes).

Will test it though.

PR #304, opened 2026-07-12
fix gcloud image tag (backport of #303 to release-v0.4.0)

ajaysundark:
## Summary
- Backports #303 to `release-v0.4.0`.
- `gcr.io/k8s-staging-test-infra/gcb-docker-gcloud:v20251211-3eba3d0954` was deleted from the registry, breaking the post-submit image push job (`post-node-readiness-controller-push-images`) on both `main` and `release-v0.4.0`.
- The fix landed on `main` via #303 but `release-v0.4.0` still has the broken pin, so the v0.4.0 staging image push has been failing since the release cut.

## Test plan
- [x] Cherry-picked cleanly from `main` (d45f1f4)
- [x] Confirm the post-submit image push job succeeds once merged and staging images for v0.4.0 appear in `k8s-staging-images/node-readiness-controller`

PR #186, opened 2026-04-17
feat: add govulncheck gh workflows

AvineshTripathi:
ref: https://github.com/kubernetes/kubernetes/blob/master/hack/verify-govulncheck.sh

PR #305, opened 2026-07-12
fix: verify-govulncheck should handle Prow's shallow checkout and non-main base branches

ajaysundark:
## Summary
`hack/verify-govulncheck.sh` hardcoded a fallback base branch of `main` and assumed it already existed as a local git ref. That holds for the GitHub Actions `govulncheck` workflow (NRC_VERIFY_GIT_BRANCH` set from `github.base_ref`), but does not work for Prow.

eg: in  #304  `pull-node-readiness-controller-verify-all` job failed on `release-0.4.0`: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_node-readiness-controller/304/pull-node-readiness-controller-verify-all/2076116037765435392.

```
Creating worktree for base branch 'main'...
fatal: invalid reference: main
```

This is because`NRC_VERIFY_GIT_BRANCH` is never set

Fix:

- Prefer Prow's `PULL_BASE_REF` env var when `NRC_VERIFY_GIT_BRANCH` isn't set.
- Fetch the base branch if it isn't already present as a local ref and Prow's checkout doesn't have it.

This was root-caused and patch generated by LLM, I verified env value set in pod yaml

```
    - name: PULL_BASE_REF
      value: release-v0.4.0
```

## Test plan
- [x] `bash -n hack/verify-govulncheck.sh`
- [x] Same fix applied to `release-v0.4.0` in #304 to unblock that PR's `verify-all` job

### DIFF
--- a/hack/verify-govulncheck.sh
+++ b/hack/verify-govulncheck.sh
@@ -26,8 +26,15 @@ if ! command -v govulncheck &>/dev/null; then
   go install "golang.org/x/vuln/cmd/govulncheck@${GOVULNCHECK_VERSION}"
 fi
 
-# NRC_VERIFY_GIT_BRANCH is populated in verify CI jobs (e.g. GITHUB_BASE_REF).
-BRANCH="${NRC_VERIFY_GIT_BRANCH:-main}"
+# NRC_VERIFY_GIT_BRANCH is populated in verify CI jobs (e.g. GITHUB_BASE_REF
+# for GitHub Actions, PULL_BASE_REF for Prow).
+BRANCH="${NRC_VERIFY_GIT_BRANCH:-${PULL_BASE_REF:-main}}"
+
+# Prow (and other shallow/single-branch checkouts) may not have the base
+# branch available as a local ref, so fetch it if needed.
+if ! git show-ref --verify --quiet "refs/heads/${BRANCH}"; then
+  git fetch --quiet origin "${BRANCH}:${BRANCH}"
+fi
 
 # Create a temp directory and clean it up on exit.
 TMPDIR="$(mktemp -d)"


vitorfloriano:
Apparently we are running govulncheck in GitHub Actions and in the verify all presubmit in Prow. 

Is there a reason for that? Or could we remove the GHA when we get a green signal in Prow?

ajaysundark:
> Apparently we are running govulncheck in GitHub Actions and in the verify all presubmit in Prow.
> 
> Is there a reason for that? Or could we remove the GHA when we get a green signal in Prow?

@vitorfloriano This check was added in https://github.com/kubernetes-sigs/node-readiness-controller/pull/186 when we were looking for a way to strengthen vulnerability handling for the controller. The script itself was ported from k/k.

There is no specific reason for this to be in presubmit, but by then we didnt have a periodic job and there's no way to escalate any vulnerabilities if found to us. Do you have a better alternative on how this could be handled for us?

vitorfloriano:
>Do you have a better alternative on how this could be handled for us?

There's been an ongoing effort to convert all GHA workflows to Prow jobs. See: https://github.com/kubernetes-sigs/node-readiness-controller/pull/210#issuecomment-4427490667 and https://github.com/kubernetes/test-infra/pull/37035

So maybe we could split the `verify-all` job into separate jobs that are triggered by touching only parts that are relevant to the verification (e.g.: only run govulncheck if Go code is changed, only verify links when the book is touched, etc).

Once we get them green in Prow, we could remove the govulncheck GHA workflow. 

WDYT?

Priyankasaggu11929:
> > Do you have a better alternative on how this could be handled for us?
> 
> There's been an ongoing effort to convert all GHA workflows to Prow jobs. See: [#210 (comment)](https://github.com/kubernetes-sigs/node-readiness-controller/pull/210#issuecomment-4427490667) and [kubernetes/test-infra#37035](https://github.com/kubernetes/test-infra/pull/37035)
> 
> So maybe we could split the `verify-all` job into separate jobs that are triggered by touching only parts that are relevant to the verification (e.g.: only run govulncheck if Go code is changed, only verify links when the book is touched, etc).
> 
> Once we get them green in Prow, we could remove the govulncheck GHA workflow.
> 
> WDYT?


(before we split any verify script to a separate presubmit) I think the first question here is - which all verify-* script signals we want to treat as blocking for PRs? All of them can stay part of the verify-all script in the list here - https://github.com/kubernetes-sigs/node-readiness-controller/blob/a4492b6e058e7aca624dc4433f836c5e7aaf5856/hack/verify-all.sh#L21-L24

re: removing verify-govulncheck from presubmit, just remove this line - https://github.com/kubernetes-sigs/node-readiness-controller/blob/a4492b6e058e7aca624dc4433f836c5e7aaf5856/hack/verify-all.sh#L24, and we can add a new periodics for running govulncheck.



---


> Apparently we are running govulncheck in GitHub Actions and in the verify all presubmit in Prow.
> Is there a reason for that? Or could we remove the GHA when we get a green signal in Prow?

+1 to cleanup GH action workflow.

Priyankasaggu11929:
Although, https://github.com/kubernetes-sigs/node-readiness-controller/pull/305#issuecomment-4955359393 is non-blocking for this PR btw.

this fix is needed (regardless of govulncheck stays part of presubmit or become a new periodics)

/lgtm

/hold
(for any further review if needed)

ajaysundark:
we should create a followup issue so this doesn't get lost-

https://github.com/kubernetes-sigs/node-readiness-controller/pull/305#issuecomment-4955359393

/unhold

PR #307, opened 2026-07-12
infra(CI): remove test-e2e GitHub Actions workflow

vitorfloriano:
## Description

This PR removes the e2e test workflow. The e2e tests have been migrated to Prow as periodic jobs. 

See: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/node-readiness-controller/node-readiness-controller-periodics.yaml

Follow-up to https://github.com/kubernetes/test-infra/pull/37165
Follow-up to https://github.com/kubernetes-sigs/node-readiness-controller/pull/210#issuecomment-4427490667 

## Type of Change

/kind cleanup


vitorfloriano:
/cc @ajaysundark 
/cc @Priyankasaggu11929 

Now that the e2e tests are green in Prow (see the [testgrid](https://testgrid.k8s.io/sig-node-node-readiness-controller#periodic-node-readiness-controller-test-e2e-main-1-36)), perhaps we could remove them from GHA workflows. 

The only question I have is if we shouldn't have at least one presubmit job with the most recent k8s version to help spot changes that may break e2e. WDYT? 

/hold

Priyankasaggu11929:
> The only question I have is if we shouldn't have at least one presubmit job with the most recent k8s version to help spot changes that may break e2e. WDYT?

yes, the idea is to move the existing gh workflow tests to prow.
so, we should add one presubmit.
this one can be used - https://github.com/kubernetes/test-infra/blob/9b391474f607e28a3dc789b32841a3cd776b4c04/config/jobs/kubernetes-sigs/node-readiness-controller/node-readiness-controller-periodics.yaml#L1-L41


---

Also, we have a new make target `test-e2e-kind` which is doing more or less same as `test-e2e` - so, i think we can clean up `test-e2e` target now -

had left a note here earlier for this cleanup - https://github.com/kubernetes-sigs/node-readiness-controller/pull/270#issue-4570974296

vitorfloriano:
> so, we should add one presubmit.

Opened a PR: https://github.com/kubernetes/test-infra/pull/37459

vitorfloriano:
Testgrid looking good: https://testgrid.k8s.io/sig-node-node-readiness-controller#pull-node-readiness-controller-test-e2e

Shall we move on with this one?

/unhold

Priyankasaggu11929:
> Also, we have a new make target test-e2e-kind which is doing more or less same as test-e2e - so, i think we can clean up test-e2e target now -
>
> had left a note here earlier for this cleanup - https://github.com/kubernetes-sigs/node-readiness-controller/pull/270#issue-4570974296

@vitorfloriano, could you please include this change as well. Thanks!

After that yes, ready to merge.

Makefile
vitorfloriano, line 1: Hi @Priyankasaggu11929 

I removed the redundant `test-e2e` target and the setup and cleanup as well since `test-e2e-kind` takes care of the entire lifecycle of the `nrr-test` kind cluster. 

PTAL.

ajaysundark:
/lgtm

Thanks @vitorfloriano , @Priyankasaggu11929 for this successful migration!

ajaysundark:
@vitorfloriano There are still some open comments. I'll wait for @Priyankasaggu11929 final lgtm before this can be merged.

Priyankasaggu11929:
Thank you.

LGTM from my side as well. (confirming the comments are addressed)

PR #310, opened 2026-07-16
Adding READMEs to improve discoverability of condition reporting patterns

arnab-logs:
Add READMEs to examples directory

This PR adds README files to the `examples/` directory to make it clear for users how node conditions can be reported to NRC, and what options are available.

NRC does not restrict which component reports a condition and anything that writes to `node.status.conditions` works. The existing examples already demonstrate multiple reporting mechanisms, but this was not surfaced clearly to users browsing the examples.

It addresses feedback that it was unclear how conditions can be reported to NRC and that the examples should show different options explicitly. With these READMEs, a user landing on the examples directory can immediately see the available reporting mechanisms and pick the example that fits their use case.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,0 +1,69 @@
+# Node Readiness Controller Examples
+
+This directory contains examples demonstrating different ways to report node conditions that can be consumed by the Node Readiness Controller (NRC).
+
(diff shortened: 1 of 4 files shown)


ajaysundark:
Thanks for looking into this, @arnab-logs .

I'll find time to look into content, but my first thought was that it'd be beneficial to write this for the mdBook than READMEs.

ajaysundark (approved):
/lgtm

Issue #313, opened 2026-07-17
[FEATURE] Support anyOf condition aggregation policy alongside the existing allOf behavior


vishnukothakapu:
### Is your feature request related to a problem or existing issue? Please describe.

NRC evaluates conditions with implicit AND logic all conditions must pass before a taint is removed. That works for most cases, but I've hit a wall with it.

We have nodes that can boot with either a hardware GPU driver *or* a software fallback. If either is ready, the node should schedule GPU workloads. There's no clean way to express this today. Two rules with the same taint key hits the webhook conflict check. The only real escape is writing a sidecar to synthesize a third aggregated condition which is extra infrastructure just to paper over a gap in the rule API.

Didn't find this covered in existing issues. #272 is about rule ordering (different), #282 is about `defaultStatus` for missing conditions (also different).

### Describe the solution you'd like

Add an optional `conditionPolicy` field `allOf` (default, existing behavior) or `anyOf` (taint removed when at least one condition is satisfied):

```yaml
spec:
  conditionPolicy: anyOf
  conditions:
    - type: "gpu.example.com/HardwareDriverReady"
      requiredStatus: "True"
    - type: "gpu.example.com/SoftwareFallbackReady"
      requiredStatus: "True"
  taint:
    key: "readiness.k8s.io/GPUReady"
    effect: "NoSchedule"
  enforcementMode: "continuous"
```

### P.S. - Post-review update:

**PR [#315](https://github.com/kubernetes-sigs/node-readiness-controller/pull/315) review identified an edge case:** when `conditionPolicy` is `anyOf` and `any` condition defines `defaultStatus == requiredStatus`, the rule silently short-circuits because the absent condition is defaulted to a value that already satisfies the `anyOf` check. A follow-up PR will add a Go webhook validation to reject this misconfiguration and update the documentation accordingly.

ajaysundark:
An `anyOf` optional behavior sounds useful.

My only concerns are that

1. It kinda breaks the homogeneity of the group of nodes evaluated by a rule.
2. It's not straightforward for the user to know what to expect of the node after bootstrap - how would cluster-operator will be able to reason about the node  satisfying the declared readiness requirements?

I welcome you to discuss your use case further if you are available to bring this to one of our biweekly calls

vishnukothakapu:
Hi @ajaysundark, thanks for the thoughtful feedback.

[...]

PR #315, opened 2026-07-20
feat: anyOf readiness condition requirements

vishnukothakapu:
- Add ConditionPolicy type (allOf/anyOf) to NodeReadinessRuleSpec
- Implement IsConditionsSatisfied() with allOf/anyOf evaluation logic
- Update controller to use conditionPolicy when evaluating conditions
- Add webhook validation to reject conditionPolicy with bootstrap-only mode
- Expand controller test coverage for conditionPolicy scenarios
- Update CRD schema to include conditionPolicy field

## Description
Introduces a conditionPolicy field (allOf | anyOf) to NodeReadinessRuleSpec that controls how the list of node conditions is aggregated when evaluating a rule.

- allOf (default): every condition must match its requiredStatus before the taint is removed — preserves existing behavior.
- anyOf: at least one condition matching its requiredStatus is sufficient to remove the taint.

## Type of Change
/kind feature /kind api-change

## Testing
- Added 4 new controller integration tests covering anyOf taint removal, anyOf taint addition, explicit allOf behavior, and missing-condition resolution under anyOf.
- Existing test suite continues to pass (default allOf path unchanged).

## Checklist
- [X] `make test` passes
- [X] `make lint` passes

Added `conditionPolicy` field to `NodeReadinessRuleSpec` (enum: `allOf` | `anyOf`).
`allOf` (default) preserves existing behavior; `anyOf` removes the taint when at least one condition is satisfied.
`anyOf` cannot be combined with `enforcementMode: bootstrap-only`.

### DIFF
--- a/api/v1alpha1/nodereadinessrule_types.go
+++ b/api/v1alpha1/nodereadinessrule_types.go
@@ -57,14 +57,14 @@ const (
 	TaintStatusAbsent TaintStatus = "Absent"
 )
 
+// Note for Developers: conditionPolicy immutability validation is placed at the
+// NodeReadinessRuleSpec level instead of conditionPolicy because conditionPolicy is optional.
+// When transitioning between omitted and explicit "allOf", field-level transition rules are
+// bypassed since CEL only evaluates them when both self and oldSelf are present. Evaluating at
+// the struct level allows using has() to normalize absent values.
+
 // NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
 //
-// We put the conditionPolicy immutability validation at `NodeReadinessRuleSpec` level instead of
-// putting it on `conditionPolicy`. This is required because conditionPolicy is optional field. If a
-// user transitions from an omitted field to an explicit "allOf" or vice-versa, field-level
-// transition rules are not evaluated, since validations are only performed only when both self and
-// oldSelf are present. By evaluating it at the struct level, we can safely use the `has()` macro to
-// normalize absent values and force evaluation under any condition.
 // +kubebuilder:validation:XValidation:rule="(!has(oldSelf.conditionPolicy) ? 'allOf' : oldSelf.conditionPolicy) == (!has(self.conditionPolicy) ? 'allOf' : self.conditionPolicy)",message="conditionPolicy is immutable"
 type NodeReadinessRuleSpec struct {
 	// conditions contains a list of the Node conditions that defines the specific
--- a/api/v1alpha1/nodereadinessrule_types.go
+++ b/api/v1alpha1/nodereadinessrule_types.go
@@ -33,6 +33,18 @@ const (
 	EnforcementModeContinuous EnforcementMode = "continuous"
 )
 
+// ConditionPolicy defines how the list of conditions is aggregated when evaluating a rule.
+// +kubebuilder:validation:Enum=allOf;anyOf
+type ConditionPolicy string
+
+const (
+	// ConditionPolicyAllOf requires ALL conditions to match their requiredStatus (default).
+	ConditionPolicyAllOf ConditionPolicy = "allOf"
+
+	// ConditionPolicyAnyOf requires at least ONE condition to match its requiredStatus.
+	ConditionPolicyAnyOf ConditionPolicy = "anyOf"
+)
+
 // TaintStatus specifies status of the Taint on Node.
 // +kubebuilder:validation:Enum=Present;Absent
 type TaintStatus string
@@ -46,6 +58,14 @@ const (
 )
 
 // NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
+//
+// We put the conditionPolicy immutability validation at `NodeReadinessRuleSpec` level instead of
+// putting it on `conditionPolicy`. This is required because conditionPolicy is optional field. If a
+// user transitions from an omitted field to an explicit "allOf" or vice-versa, field-level
+// transition rules are not evaluated, since validations are only performed only when both self and
+// oldSelf are present. By evaluating it at the struct level, we can safely use the `has()` macro to
+// normalize absent values and force evaluation under any condition.
+// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.conditionPolicy) ? 'allOf' : oldSelf.conditionPolicy) == (!has(self.conditionPolicy) ? 'allOf' : self.conditionPolicy)",message="conditionPolicy is immutable"
 type NodeReadinessRuleSpec struct {
 	// conditions contains a list of the Node conditions that defines the specific
 	// criteria that must be met for taints to be managed on the target Node.
@@ -98,6 +118,15 @@ type NodeReadinessRuleSpec struct {
 	// +required
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="nodeSelector is immutable"
 	NodeSelector metav1.LabelSelector `json:"nodeSelector,omitempty,omitzero"`
+
+	// conditionPolicy controls how the conditions list is evaluated.
+	// "allOf" (default) requires every condition to match its requiredStatus before the taint is removed.
+	// "anyOf" requires at least one condition to match its requiredStatus.
+	//
+	// anyOf cannot be used with enforcementMode: bootstrap-only.
+	//
+	// +optional
+	ConditionPolicy ConditionPolicy `json:"conditionPolicy,omitempty"` // Use GetConditionPolicy() for safe access; field may be empty even when allOf applies.
 
 	// dryRun when set to true, The controller will evaluate Node conditions and log intended taint modifications
 	// without persisting changes to the cluster. Proposed actions are reflected in the resource status.
@@ -377,6 +406,20 @@ func (c *ConditionRequirement) GetDefaultStatus() corev1.ConditionStatus {
 		return corev1.ConditionUnknown
 	}
 	return c.DefaultStatus
+}
+
+// GetConditionPolicy returns the effective condition policy, defaulting to allOf
+// when the field is not explicitly set.
+//
+// Always use this method instead of reading ConditionPolicy directly. The field
+// is intentionally left without an OpenAPI schema default (kubebuilder:default
+// is forbidden by project policy) and the Spec is immutable, so defaulting
+// must happen at read time via this accessor.
+func (spec *NodeReadinessRuleSpec) GetConditionPolicy() ConditionPolicy {
+	if spec.ConditionPolicy == "" {
+		return ConditionPolicyAllOf
+	}
+	return spec.ConditionPolicy
 }
 
 func init() {
--- a/config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml
+++ b/config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml
@@ -62,6 +62,17 @@ spec:
           spec:
             description: spec defines the desired state of NodeReadinessRule
             properties:
+              conditionPolicy:
+                description: |-
+                  conditionPolicy controls how the conditions list is evaluated.
+                  "allOf" (default) requires every condition to match its requiredStatus before the taint is removed.
+                  "anyOf" requires at least one condition to match its requiredStatus.
+
+                  anyOf cannot be used with enforcementMode: bootstrap-only.
+                enum:
+                - allOf
+                - anyOf
+                type: string
               conditions:
                 description: |-
                   conditions contains a list of the Node conditions that defines the specific
@@ -251,6 +262,10 @@ spec:
             - nodeSelector
             - taint
             type: object
+            x-kubernetes-validations:
+            - message: conditionPolicy is immutable
+              rule: '(!has(oldSelf.conditionPolicy) ? ''allOf'' : oldSelf.conditionPolicy)
+                == (!has(self.conditionPolicy) ? ''allOf'' : self.conditionPolicy)'
           status:
             description: status defines the observed state of NodeReadinessRule
             minProperties: 1
--- a/docs/book/src/reference/api-spec.md
+++ b/docs/book/src/reference/api-spec.md
@@ -27,10 +27,10 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `type` _string_ | type corresponds to the Node condition type being evaluated. |  | MaxLength: 316 <br />MinLength: 1 <br />Required: \{\} <br /> |
-| `currentStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | currentStatus is the actual status value observed on the Node, one of True, False, Unknown. |  | Enum: [True False Unknown] <br />Required: \{\} <br /> |
-| `requiredStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | requiredStatus is the status value defined in the rule that must be matched, one of True, False, Unknown. |  | Enum: [True False Unknown] <br />Required: \{\} <br /> |
-| `defaultStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | defaultStatus is the status a condition is evaluated to if the condition<br />is not found in a node. Reflects the defaultStatus configured in the rule<br />spec. |  | Enum: [True False Unknown] <br />Optional: \{\} <br /> |
+| `type` _string_ | type corresponds to the Node condition type being evaluated. |  | MaxLength: 316 <br />MinLength: 1 <br /> |
+| `currentStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | currentStatus is the actual status value observed on the Node, one of True, False, Unknown. |  | Enum: [True False Unknown] <br /> |
+| `requiredStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | requiredStatus is the status value defined in the rule that must be matched, one of True, False, Unknown. |  | Enum: [True False Unknown] <br /> |
+| `defaultStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | defaultStatus is the status a condition is evaluated to if the condition<br />is not found in a node. Reflects the defaultStatus configured in the rule<br />spec. |  | Enum: [True False Unknown] <br /> |
 
 
 #### ConditionPolicy
@@ -66,9 +66,9 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `type` _string_ | type of Node condition<br />Following kubebuilder validation is referred from https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition |  | MaxLength: 316 <br />MinLength: 1 <br />Required: \{\} <br /> |
-| `requiredStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | requiredStatus is status of the condition, one of True, False, Unknown. |  | Enum: [True False Unknown] <br />Required: \{\} <br /> |
-| `defaultStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | defaultStatus is the status a condition is evaluated to if the condition<br />is not found in a node.<br />Accepted values are True, False, Unknown. It is optional.<br />When omitted, the effective default is Unknown, applied transparently by<br />the controller at evaluation time.<br />Note: This field must not be set when enforcementMode is bootstrap-only. |  | Enum: [True False Unknown] <br />Optional: \{\} <br /> |
+| `type` _string_ | type of Node condition<br />Following kubebuilder validation is referred from https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition |  | MaxLength: 316 <br />MinLength: 1 <br /> |
+| `requiredStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | requiredStatus is status of the condition, one of True, False, Unknown. |  | Enum: [True False Unknown] <br /> |
+| `defaultStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | defaultStatus is the status a condition is evaluated to if the condition<br />is not found in a node.<br />Accepted values are True, False, Unknown. It is optional.<br />When omitted, the effective default is Unknown, applied transparently by<br />the controller at evaluation time.<br />Note: This field must not be set when enforcementMode is bootstrap-only. |  | Enum: [True False Unknown] <br /> |
 
 
 #### DryRunResults
@@ -85,11 +85,11 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `affectedNodes` _integer_ | affectedNodes is the total count of Nodes that match the rule's criteria. |  | Minimum: 0 <br />Optional: \{\} <br /> |
-| `taintsToAdd` _integer_ | taintsToAdd is the number of Nodes that currently lack the specified taint and would have it applied. |  | Minimum: 0 <br />Optional: \{\} <br /> |
-| `taintsToRemove` _integer_ | taintsToRemove is the number of Nodes that currently possess the<br />taint but no longer meet the criteria, leading to its removal. |  | Minimum: 0 <br />Optional: \{\} <br /> |
-| `riskyOperations` _integer_ | riskyOperations represents the count of Nodes where required conditions<br />are missing entirely, potentially indicating an ambiguous node state. |  | Minimum: 0 <br />Optional: \{\} <br /> |
-| `summary` _string_ | summary provides a human-readable overview of the dry run evaluation,<br />highlighting key findings or warnings. |  | MaxLength: 4096 <br />MinLength: 1 <br />Required: \{\} <br /> |
+| `affectedNodes` _integer_ | affectedNodes is the total count of Nodes that match the rule's criteria. |  | Minimum: 0 <br /> |
+| `taintsToAdd` _integer_ | taintsToAdd is the number of Nodes that currently lack the specified taint and would have it applied. |  | Minimum: 0 <br /> |
+| `taintsToRemove` _integer_ | taintsToRemove is the number of Nodes that currently possess the<br />taint but no longer meet the criteria, leading to its removal. |  | Minimum: 0 <br /> |
+| `riskyOperations` _integer_ | riskyOperations represents the count of Nodes where required conditions<br />are missing entirely, potentially indicating an ambiguous node state. |  | Minimum: 0 <br /> |
+| `summary` _string_ | summary provides a human-readable overview of the dry run evaluation,<br />highlighting key findings or warnings. |  | MaxLength: 4096 <br />MinLength: 1 <br /> |
 
 
 #### EnforcementMode
@@ -123,10 +123,10 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `nodeName` _string_ | nodeName is the name of the evaluated Node. |  | MaxLength: 253 <br />MinLength: 1 <br />Pattern: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$` <br />Required: \{\} <br /> |
-| `conditionResults` _[ConditionEvaluationResult](#conditionevaluationresult) array_ | conditionResults provides a detailed breakdown of each condition evaluation<br />for this Node. This allows for granular auditing of which specific<br />criteria passed or failed during the rule assessment. |  | MaxItems: 5000 <br />Required: \{\} <br /> |
-| `taintStatus` _[TaintStatus](#taintstatus)_ | taintStatus represents the taint status on the Node, one of Present, Absent. |  | Enum: [Present Absent] <br />Required: \{\} <br /> |
-| `lastEvaluationTime` _[Time](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#time-v1-meta)_ | lastEvaluationTime is the timestamp when the controller last assessed this Node. |  | Required: \{\} <br /> |
+| `nodeName` _string_ | nodeName is the name of the evaluated Node. |  | MaxLength: 253 <br />MinLength: 1 <br />Pattern: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$` <br /> |
+| `conditionResults` _[ConditionEvaluationResult](#conditionevaluationresult) array_ | conditionResults provides a detailed breakdown of each condition evaluation<br />for this Node. This allows for granular auditing of which specific<br />criteria passed or failed during the rule assessment. |  | MaxItems: 5000 <br /> |
+| `taintStatus` _[TaintStatus](#taintstatus)_ | taintStatus represents the taint status on the Node, one of Present, Absent. |  | Enum: [Present Absent] <br /> |
+| `lastEvaluationTime` _[Time](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#time-v1-meta)_ | lastEvaluationTime is the timestamp when the controller last assessed this Node. |  |  |
 
 
 #### NodeFailure
@@ -142,10 +142,10 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `nodeName` _string_ | nodeName is the name of the failed Node.<br />Following kubebuilder validation is referred from<br />https://github.com/kubernetes/apimachinery/blob/84d740c9e27f3ccc94c8bc4d13f1b17f60f7080b/pkg/util/validation/validation.go#L198 |  | MaxLength: 253 <br />MinLength: 1 <br />Pattern: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$` <br />Required: \{\} <br /> |
-| `reason` _string_ | reason provides a brief explanation of the evaluation result. |  | MaxLength: 256 <br />MinLength: 1 <br />Optional: \{\} <br /> |
-| `message` _string_ | message is a human-readable message indicating details about the evaluation. |  | MaxLength: 10240 <br />MinLength: 1 <br />Optional: \{\} <br /> |
-| `lastEvaluationTime` _[Time](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#time-v1-meta)_ | lastEvaluationTime is the timestamp of the last rule check failed for this Node. |  | Required: \{\} <br /> |
+| `nodeName` _string_ | nodeName is the name of the failed Node.<br />Following kubebuilder validation is referred from<br />https://github.com/kubernetes/apimachinery/blob/84d740c9e27f3ccc94c8bc4d13f1b17f60f7080b/pkg/util/validation/validation.go#L198 |  | MaxLength: 253 <br />MinLength: 1 <br />Pattern: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$` <br /> |
+| `reason` _string_ | reason provides a brief explanation of the evaluation result. |  | MaxLength: 256 <br />MinLength: 1 <br /> |
+| `message` _string_ | message is a human-readable message indicating details about the evaluation. |  | MaxLength: 10240 <br />MinLength: 1 <br /> |
+| `lastEvaluationTime` _[Time](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#time-v1-meta)_ | lastEvaluationTime is the timestamp of the last rule check failed for this Node. |  |  |
 
 
 #### NodeReadinessRule
@@ -162,9 +162,9 @@ NodeReadinessRule is the Schema for the NodeReadinessRules API.
 | --- | --- | --- | --- |
 | `apiVersion` _string_ | `readiness.node.x-k8s.io/v1alpha1` | | |
 | `kind` _string_ | `NodeReadinessRule` | | |
-| `metadata` _[ObjectMeta](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#objectmeta-v1-meta)_ | Refer to Kubernetes API documentation for fields of `metadata`. |  | Optional: \{\} <br /> |
-| `spec` _[NodeReadinessRuleSpec](#nodereadinessrulespec)_ | spec defines the desired state of NodeReadinessRule |  | Required: \{\} <br /> |
-| `status` _[NodeReadinessRuleStatus](#nodereadinessrulestatus)_ | status defines the observed state of NodeReadinessRule |  | MinProperties: 1 <br />Optional: \{\} <br /> |
+| `metadata` _[ObjectMeta](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#objectmeta-v1-meta)_ | Refer to Kubernetes API documentation for fields of `metadata`. |  |  |
+| `spec` _[NodeReadinessRuleSpec](#nodereadinessrulespec)_ | spec defines the desired state of NodeReadinessRule |  |  |
+| `status` _[NodeReadinessRuleStatus](#nodereadinessrulestatus)_ | status defines the observed state of NodeReadinessRule |  | MinProperties: 1 <br /> |
 
 
 #### NodeReadinessRuleSpec
@@ -173,13 +173,6 @@ NodeReadinessRule is the Schema for the NodeReadinessRules API.
 
 NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
 
-We put the conditionPolicy immutability validation at `NodeReadinessRuleSpec` level instead of
-putting it on `conditionPolicy`. This is required because conditionPolicy is optional field. If a
-user transitions from an omitted field to an explicit "allOf" or vice-versa, field-level
-transition rules are not evaluated, since validations are only performed only when both self and
-oldSelf are present. By evaluating it at the struct level, we can safely use the `has()` macro to
-normalize absent values and force evaluation under any condition.
-
 
 
 _Appears in:_
@@ -187,12 +180,12 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `conditions` _[ConditionRequirement](#conditionrequirement) array_ | conditions contains a list of the Node conditions that defines the specific<br />criteria that must be met for taints to be managed on the target Node.<br />The presence or status of these conditions directly triggers the application or removal of Node taints. |  | MaxItems: 32 <br />MinItems: 1 <br />Required: \{\} <br /> |
-| `enforcementMode` _[EnforcementMode](#enforcementmode)_ | enforcementMode specifies how the controller maintains the desired state.<br />enforcementMode is one of bootstrap-only, continuous.<br />"bootstrap-only" applies the configuration once during initial setup.<br />"continuous" ensures the state is monitored and corrected throughout the resource lifecycle. |  | Enum: [bootstrap-only continuous] <br />Required: \{\} <br /> |
-| `taint` _[Taint](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#taint-v1-core)_ | taint defines the specific Taint (Key, Value, and Effect) to be managed<br />on Nodes that meet the defined condition criteria.<br />The taint key must follow Kubernetes qualified name format: prefix/name<br />where prefix is 'readiness.k8s.io' (DNS subdomain) and name is a qualified<br />name (max 63 chars, alphanumeric, '-', '_', '.', must start and end with alphanumeric).<br />ref: git.k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/api/validate/content/kube.go#L24-L72<br />Supported effects: NoSchedule, PreferNoSchedule, NoExecute.<br />Caution: NoExecute evicts existing pods and can cause significant disruption<br />when combined with continuous enforcement mode. Prefer NoSchedule for most use cases. |  | Required: \{\} <br /> |
-| `nodeSelector` _[LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#labelselector-v1-meta)_ | nodeSelector limits the scope of this rule to a specific subset of Nodes. |  | Required: \{\} <br /> |
-| `conditionPolicy` _[ConditionPolicy](#conditionpolicy)_ | conditionPolicy controls how the conditions list is evaluated.<br />"allOf" (default) requires every condition to match its requiredStatus before the taint is removed.<br />"anyOf" requires at least one condition to match its requiredStatus.<br />anyOf cannot be used with enforcementMode: bootstrap-only. |  | Enum: [allOf anyOf] <br />Optional: \{\} <br /> |
-| `dryRun` _boolean_ | dryRun when set to true, The controller will evaluate Node conditions and log intended taint modifications<br />without persisting changes to the cluster. Proposed actions are reflected in the resource status. |  | Optional: \{\} <br /> |
+| `conditions` _[ConditionRequirement](#conditionrequirement) array_ | conditions contains a list of the Node conditions that defines the specific<br />criteria that must be met for taints to be managed on the target Node.<br />The presence or status of these conditions directly triggers the application or removal of Node taints. |  | MaxItems: 32 <br />MinItems: 1 <br /> |
+| `enforcementMode` _[EnforcementMode](#enforcementmode)_ | enforcementMode specifies how the controller maintains the desired state.<br />enforcementMode is one of bootstrap-only, continuous.<br />"bootstrap-only" applies the configuration once during initial setup.<br />"continuous" ensures the state is monitored and corrected throughout the resource lifecycle. |  | Enum: [bootstrap-only continuous] <br /> |
+| `taint` _[Taint](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#taint-v1-core)_ | taint defines the specific Taint (Key, Value, and Effect) to be managed<br />on Nodes that meet the defined condition criteria.<br />The taint key must follow Kubernetes qualified name format: prefix/name<br />where prefix is 'readiness.k8s.io' (DNS subdomain) and name is a qualified<br />name (max 63 chars, alphanumeric, '-', '_', '.', must start and end with alphanumeric).<br />ref: git.k8s.io/kubernetes/staging/src/k8s.io/apimachinery/pkg/api/validate/content/kube.go#L24-L72<br />Supported effects: NoSchedule, PreferNoSchedule, NoExecute.<br />Caution: NoExecute evicts existing pods and can cause significant disruption<br />when combined with continuous enforcement mode. Prefer NoSchedule for most use cases. |  |  |
+| `nodeSelector` _[LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#labelselector-v1-meta)_ | nodeSelector limits the scope of this rule to a specific subset of Nodes. |  |  |
+| `conditionPolicy` _[ConditionPolicy](#conditionpolicy)_ | conditionPolicy controls how the conditions list is evaluated.<br />"allOf" (default) requires every condition to match its requiredStatus before the taint is removed.<br />"anyOf" requires at least one condition to match its requiredStatus.<br />anyOf cannot be used with enforcementMode: bootstrap-only. |  | Enum: [allOf anyOf] <br /> |
+| `dryRun` _boolean_ | dryRun when set to true, The controller will evaluate Node conditions and log intended taint modifications<br />without persisting changes to the cluster. Proposed actions are reflected in the resource status. |  |  |
 
 
 #### NodeReadinessRuleStatus
@@ -209,11 +202,11 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `observedGeneration` _integer_ | observedGeneration reflects the generation of the most recently observed NodeReadinessRule by the controller. |  | Minimum: 1 <br />Optional: \{\} <br /> |
-| `appliedNodes` _string array_ | appliedNodes lists the names of Nodes where the taint has been successfully managed.<br />This provides a quick reference to the scope of impact for this rule. |  | MaxItems: 5000 <br />items:MaxLength: 253 <br />Optional: \{\} <br /> |
-| `failedNodes` _[NodeFailure](#nodefailure) array_ | failedNodes lists the Nodes where the rule evaluation encountered an error.<br />This is used for troubleshooting configuration issues, such as invalid selectors during node lookup. |  | MaxItems: 5000 <br />Optional: \{\} <br /> |
-| `nodeEvaluations` _[NodeEvaluation](#nodeevaluation) array_ | nodeEvaluations provides detailed insight into the rule's assessment for individual Nodes.<br />This is primarily used for auditing and debugging why specific Nodes were or<br />were not targeted by the rule. |  | MaxItems: 5000 <br />Optional: \{\} <br /> |
-| `dryRunResults` _[DryRunResults](#dryrunresults)_ | dryRunResults captures the outcome of the rule evaluation when DryRun is enabled.<br />This field provides visibility into the actions the controller would have taken,<br />allowing users to preview taint changes before they are committed. |  | MinProperties: 1 <br />Optional: \{\} <br /> |
+| `observedGeneration` _integer_ | observedGeneration reflects the generation of the most recently observed NodeReadinessRule by the controller. |  | Minimum: 1 <br /> |
+| `appliedNodes` _string array_ | appliedNodes lists the names of Nodes where the taint has been successfully managed.<br />This provides a quick reference to the scope of impact for this rule. |  | MaxItems: 5000 <br />items:MaxLength: 253 <br /> |
+| `failedNodes` _[NodeFailure](#nodefailure) array_ | failedNodes lists the Nodes where the rule evaluation encountered an error.<br />This is used for troubleshooting configuration issues, such as invalid selectors during node lookup. |  | MaxItems: 5000 <br /> |
+| `nodeEvaluations` _[NodeEvaluation](#nodeevaluation) array_ | nodeEvaluations provides detailed insight into the rule's assessment for individual Nodes.<br />This is primarily used for auditing and debugging why specific Nodes were or<br />were not targeted by the rule. |  | MaxItems: 5000 <br /> |
+| `dryRunResults` _[DryRunResults](#dryrunresults)_ | dryRunResults captures the outcome of the rule evaluation when DryRun is enabled.<br />This field provides visibility into the actions the controller would have taken,<br />allowing users to preview taint changes before they are committed. |  | MinProperties: 1 <br /> |
 
 
 #### TaintStatus
--- a/docs/book/src/reference/api-spec.md
+++ b/docs/book/src/reference/api-spec.md
@@ -27,10 +27,28 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `type` _string_ | type corresponds to the Node condition type being evaluated. |  | MaxLength: 316 <br />MinLength: 1 <br /> |
-| `currentStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | currentStatus is the actual status value observed on the Node, one of True, False, Unknown. |  | Enum: [True False Unknown] <br /> |
-| `requiredStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | requiredStatus is the status value defined in the rule that must be matched, one of True, False, Unknown. |  | Enum: [True False Unknown] <br /> |
-| `defaultStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | defaultStatus is the fallback status value configured in the rule spec, which is used for evaluation if the condition is not present on the Node. |  | Enum: [True False Unknown] <br /> |
+| `type` _string_ | type corresponds to the Node condition type being evaluated. |  | MaxLength: 316 <br />MinLength: 1 <br />Required: \{\} <br /> |
+| `currentStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | currentStatus is the actual status value observed on the Node, one of True, False, Unknown. |  | Enum: [True False Unknown] <br />Required: \{\} <br /> |
+| `requiredStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | requiredStatus is the status value defined in the rule that must be matched, one of True, False, Unknown. |  | Enum: [True False Unknown] <br />Required: \{\} <br /> |
+| `defaultStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | defaultStatus is the status a condition is evaluated to if the condition<br />is not found in a node. Reflects the defaultStatus configured in the rule<br />spec. |  | Enum: [True False Unknown] <br />Optional: \{\} <br /> |
+
+
+#### ConditionPolicy
+
+_Underlying type:_ _string_
+
+ConditionPolicy defines how the list of conditions is aggregated when evaluating a rule.
+
+_Validation:_
+- Enum: [allOf anyOf]
+
+_Appears in:_
+- [NodeReadinessRuleSpec](#nodereadinessrulespec)
+
+| Field | Description |
+| --- | --- |
+| `allOf` | ConditionPolicyAllOf requires ALL conditions to match their requiredStatus (default).<br /> |
+| `anyOf` | ConditionPolicyAnyOf requires at least ONE condition to match its requiredStatus.<br /> |
 
 
 #### ConditionRequirement
@@ -38,7 +56,8 @@ _Appears in:_
 
 
 ConditionRequirement defines a specific Node condition and the status value
-required to trigger the controller's action.
+required to trigger the controller's action. It also contains an optional
+default status value.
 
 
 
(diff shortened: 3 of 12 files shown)


ajaysundark:
@vishnukothakapu is this ready for review?

@AnuragThePathak The code-changes are close to what you already worked on for `defaultStatus`. Are you available for a first pass review?

AnuragThePathak:
> @vishnukothakapu is this ready for review?
> 
> @AnuragThePathak The code-changes are close to what you already worked on for `defaultStatus`. Are you available for a first pass review?

Sure, I'll review within next 24 hours.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -99,6 +111,17 @@ type NodeReadinessRuleSpec struct {
...
+	// conditionPolicy controls how the conditions list is evaluated.
+	// "allOf" (default) requires every condition to match its requiredStatus before the taint is removed.
+	// "anyOf" requires at least one condition to match its requiredStatus.
+	//
+	// Cannot be used with enforcementMode: bootstrap-only.
+	//
+	// +optional
+	// +kubebuilder:default=allOf
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="conditionPolicy is immutable"
+	ConditionPolicy ConditionPolicy `json:"conditionPolicy,omitempty"`
```
Karthik-K-N: Few things
1. Lets avoid kubebuilder default marker, It will complicate whether the user has really set the value vs api defaulted, when we are doing api migration.
2. The current approach is simple and serves the purpose, but if we want to support some advanced logics like (conditionA AND conditionB) OR (conditionC AND conditionD) we can't achieve with this model.
If we anticipate these kind of use cases then we can take reference from how NodeAffinity or PodAffinity is handled

vishnukothakapu: Good point. The current `allOf`/`anyOf` enum is intentionally simple for the first iteration, so it doesn't support compound logic.

I see two options:

- **Keep the current design.** For `v1alpha1`, a simple API seems reasonable. Since the field is immutable, we could introduce a richer `conditionRequirements` structure in a future version without breaking existing users.
- **Redesign it now.** Instead of a flat list with `conditionPolicy`, we could use a `NodeSelector`-style expression tree, for example:

```yaml
conditionRequirements:
  anyOf:
    - allOf:
        - type: CondA
          requiredStatus: True
        - type: CondB
          requiredStatus: True
    - allOf:
        - type: CondC
          requiredStatus: True
```
I prefer the first option because the immediate use case in #313 only requires a flat anyOf. If we expect more complex condition logic in the near future, I'm happy to prototype the expression-tree approach instead.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -379,6 +401,50 @@ func (c *ConditionRequirement) GetDefaultStatus() corev1.ConditionStatus {
...
+// policy defined in the spec.
+//
+//   - allOf: every result's effective status must equal its RequiredStatus.
+//   - anyOf: at least one result's effective status must equal its RequiredStatus.
+//
+// The effective status is CurrentStatus, unless the condition was not found on
+// the Node (CurrentStatus == Unknown) and a DefaultStatus is configured — in
+// that case DefaultStatus is used as the fallback, matching the original
+// per-condition evaluation logic in the controller.
+func (spec *NodeReadinessRuleSpec) IsConditionsSatisfied(results []ConditionEvaluationResult) bool {
```
AnuragThePathak: This code must not belong to API package. This is a business logic.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -379,6 +401,50 @@ func (c *ConditionRequirement) GetDefaultStatus() corev1.ConditionStatus {
 	return c.DefaultStatus
 }
 
+// GetConditionPolicy returns the effective condition policy, defaulting to allOf
+// when the field is not explicitly set.
```
AnuragThePathak, line 412: I would encourage adding a notice on why developers should not use the raw condition policy value directly, as we did in default status

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -379,6 +401,50 @@ func (c *ConditionRequirement) GetDefaultStatus() corev1.ConditionStatus {
...
+//   - allOf: every result's effective status must equal its RequiredStatus.
+//   - anyOf: at least one result's effective status must equal its RequiredStatus.
+//
+// The effective status is CurrentStatus, unless the condition was not found on
+// the Node (CurrentStatus == Unknown) and a DefaultStatus is configured — in
+// that case DefaultStatus is used as the fallback, matching the original
+// per-condition evaluation logic in the controller.
+func (spec *NodeReadinessRuleSpec) IsConditionsSatisfied(results []ConditionEvaluationResult) bool {
+	effectiveStatus := func(r ConditionEvaluationResult) corev1.ConditionStatus {
+		if r.CurrentStatus == corev1.ConditionUnknown && r.DefaultStatus != "" {
```
AnuragThePathak: Once you move it to right package you will be able to use already defined method replacing it.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -357,12 +355,12 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
 			"satisfied", satisfied)
 	}
 
-	// Determine taint action
-	shouldRemoveTaint := allConditionsSatisfied
+	// Determine taint action using the rule's conditionPolicy (allOf or anyOf)
+	shouldRemoveTaint := rule.Spec.IsConditionsSatisfied(conditionResults)
```
AnuragThePathak: I don't think it's the right choice to rely on the results to finally conclude conditions satisfied or not. If there is any bug during result calculation/forming the result to show it will propagate to taint apply/remove process.

Also I think spec is what we should process not status. Status values should not be used for reconciliation process.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -604,9 +602,9 @@ func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readi
 
 		affectedNodes++
 
-		// Simulate rule evaluation
-		allConditionsSatisfied := true
+		// Simulate rule evaluation using the rule's conditionPolicy
 		missingConditions := 0
+		dryRunResults := make([]readinessv1alpha1.ConditionEvaluationResult, 0, len(rule.Spec.Conditions))
```
AnuragThePathak: This unnecessary usage of results array here will also be solved by https://github.com/kubernetes-sigs/node-readiness-controller/pull/315/changes#r3640651436

manifests.yaml
AnuragThePathak, line 1: Couldn't figure out why this file is appearing in the PR, just double check once.

AnuragThePathak (changes requested):
Haven't reviewed tests yet but I think some of these changes are necessary.

.gitignore
```diff
@@ -38,3 +38,5 @@ hack/test-workloads/ignore-pending/*
 CLAUDE.md
 GEMINI.md
 
+# Build artifacts (generated by make build-installer / kustomize build)
+manifests.yaml
```
AnuragThePathak: I genuinely don't know if one's supposed to put manifests.yaml in .gitignore

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -379,6 +401,36 @@ func (c *ConditionRequirement) GetDefaultStatus() corev1.ConditionStatus {
...
+
+// GetDefaultStatus returns the effective default status for a condition evaluation
+// result whose condition was not found on the node. If the field is unset (empty
+// string), it falls back to corev1.ConditionUnknown.
+//
+// Always use this method instead of reading DefaultStatus directly. The field
+// is intentionally left without an OpenAPI schema default (kubebuilder:default
+// is forbidden by project policy) and the Spec is immutable, so defaulting
+// must happen at read time via this accessor.
+func (r *ConditionEvaluationResult) GetDefaultStatus() corev1.ConditionStatus {
```
AnuragThePathak: unused method.

internal/controller/node_controller.go
```diff
@@ -246,19 +246,6 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 	return errors.Join(errs...)
 }
 
-// getConditionStatus gets the status of a condition on a node.
```
AnuragThePathak, line 250: don't think this was the right call.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -605,25 +602,15 @@ func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readi
 
 		affectedNodes++
 
-		// Simulate rule evaluation
-		allConditionsSatisfied := true
+		// Simulate rule evaluation using the rule's conditionPolicy
 		missingConditions := 0
-
 		for _, condReq := range rule.Spec.Conditions {
```
AnuragThePathak, line 623: What I was thinking earlier, was to leverage this loop only to complete our evaluation process. If done perfectly, I think this loop is more than enough to accommodate your new changes, while also retaining the node evaluation results for all conditions. Though you have to be decisive to get it done in a clean way with minimal changes which is our goal.

AnuragThePathak (commented):
@vishnukothakapu I noticed that you deleted your previous commit and created a completely new one. From next time, prefer not removing the already pushed commits unless you know it won't have massive impact. Retaining what you already did helps review most of the time.

In any case, for my new comments, I won't suggest you to hurry. Instead I would encourage you to take some time and think (feel free to leverage AI either for ideas), then come up with something that allows you achieve the goal with minimal changes in what's already there following the existing conventions.

With that being said, also give others as well some time to reply, I am also curious to see if others also agree with my opinion of not introducing a completely new method for calculating is condition is satisfied or not.

P.S. if anything is not clear or you need more help, feel free to ask questions than applying some changes randomly.

.gitignore
vishnukothakapu: Good catch. I checked the Makefile and build-manifests-temp generates bin/build/manifests.yaml, which is already ignored by bin/* in .gitignore. The manifests.yaml in the repository root is a tracked release artifact, not a local build file. I added it to .gitignore by mistake while working with a locally generated copy. I'll remove that entry.

AnuragThePathak:
@vishnukothakapu great job. You have perfectly nailed the direction, I was referring to. I have some minor feedbacks which I'll address today if possible or tomorrow at worst. In the meantime you can proceed updating the test cases probably as they are failing now and also docs if applicable. I don't anticipate any further functional changes.

I am excluding, the point Karthik mentioned here by the way, I'll leave that up to you.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -321,9 +320,10 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
 	defer timer.ObserveDuration()
 	log := ctrl.LoggerFrom(ctx)
 
-	// Evaluate all conditions (ALL logic)
-	allConditionsSatisfied := true
+	// Evaluate all conditions, accumulating the policy result alongside per-condition results.
 	conditionResults := make([]readinessv1alpha1.ConditionEvaluationResult, 0, len(rule.Spec.Conditions))
+	conditionPolicy := rule.Spec.GetConditionPolicy()
+	allSatisfied, anySatisfied := true, false
```
AnuragThePathak: Let's split it into two lines. That would be more readable, and more aligned with go conventions

internal/controller/nodereadinessrule_controller.go
```diff
@@ -333,6 +333,13 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
 		)
 		satisfied := effectiveStatus == condReq.RequiredStatus
 
+		if !satisfied {
```
AnuragThePathak, line 340: This part can be merged with !satisfied block below (line no 342 in main, 350 in your change) only. Put an else to do the other part.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -605,25 +614,29 @@ func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readi
 
 		affectedNodes++
 
-		// Simulate rule evaluation
-		allConditionsSatisfied := true
+		// Simulate rule evaluation using the rule's conditionPolicy
+		conditionPolicy := rule.Spec.GetConditionPolicy()
 		missingConditions := 0
+		allSatisfied, anySatisfied := true, false
```
AnuragThePathak: same splitting here.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -605,25 +614,29 @@ func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readi
...
 
 		for _, condReq := range rule.Spec.Conditions {
```
AnuragThePathak, line 614: Unnecessary change from git diff point of view.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -605,25 +614,29 @@ func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readi
...
+			if !found {
 				missingConditions++
 			}
-			if currentStatus != condReq.RequiredStatus {
-				allConditionsSatisfied = false
+			satisfied := effectiveStatus == condReq.RequiredStatus
+			if !satisfied {
+				allSatisfied = false
+			}
+			if satisfied {
```
AnuragThePathak: use else

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2178,7 +2180,7 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 			Expect(failedNames).To(ContainElement("fail-path-node"))
 		})
 
-		It("should remove stale failedNodes entry when evaluation succeeds and include the node in appliedNodes", func() {
```
AnuragThePathak, line 2251: I genuinely feel this line was more descriptive and helpful when someone sees this test case failing. Any reason you decided to replace it?

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -1043,30 +1043,32 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 				},
 			}
 
+			r := &RuleReadinessController{}
```
AnuragThePathak: This is also unnecessary change. And below readinessController.getConditionStatus updates.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2235,4 +2237,158 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 			Expect(failedNames).NotTo(ContainElement("stale-recovery-node"))
 		})
 	})
+
+	Context("ConditionPolicy", func() {
```
AnuragThePathak, line 2318: This is something I won't label as something you need to change right away, but personally I just felt trying to stick close to other neighborhood tests was a more sensible choice overall in my case.

It's just a suggestion from my side, to evaluate once if your newly introduced test cases doesn't create a different style/pattern for adding new test cases. Usually we try to minimize anomalies to avoid making things harder for future contributors and maintainers but feel free to keep the your current way as well if you feel you adhere to what I said and it's the right way to do it.

AnuragThePathak (commented):
Didn't review the test cases themselves as not passing as of now. Apply the changes I have suggested, fix the failing test cases, and add docs. We should be good.

internal/controller/nodereadinessrule_controller.go
AnuragThePathak: Hi @vishnukothakapu you didn't undo this part 😐. In particular it's not a problem at all, but when someone opens up this PR/stashed commit to see the changes was made here, they will unnecessary stare at this section for some 5-20 seconds.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2235,4 +2235,159 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+		var (
+			anyOfController *RuleReadinessController
+			anyOfNode       *corev1.Node
+		)
+
+		BeforeEach(func() {
+			anyOfController = &RuleReadinessController{
```
AnuragThePathak: I was looking at readinessController defined at the top (description section of tests), seems like it already is what you are defining a new anyOfController for.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2235,4 +2235,159 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+				clientset:     fakeClientset,
+				ruleCache:     make(map[string]*nodereadinessiov1alpha1.NodeReadinessRule),
+				EventRecorder: events.NewFakeRecorder(10),
+			}
+			anyOfNode = &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:   "anyof-test-node",
+					Labels: map[string]string{"anyof-test": "true"},
+				},
+				Spec:   corev1.NodeSpec{},
```
AnuragThePathak: Rather than leaving spec and status empty in the original node resource object, you can assign some default value here. That way fewer reassignment will also be needed in later parts.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2235,4 +2235,159 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+
+			Expect(k8sClient.Create(ctx, anyOfNode)).To(Succeed())
+			defer func() { Expect(k8sClient.Delete(ctx, anyOfNode)).To(Succeed()) }()
+			Expect(anyOfController.evaluateRuleForNode(ctx, rule, anyOfNode)).To(Succeed())
+
+			// Taint must remain because CondB is still False
+			Expect(anyOfController.hasTaintBySpec(anyOfNode, rule.Spec.Taint)).To(BeTrue())
+		})
+
+		It("anyOf: missing condition evaluated via defaultStatus does not auto-satisfy", func() {
```
AnuragThePathak: This one is like testing if default status is implemented right way or not which is the job of default status related tests not conditionPolicy. I think we should not require this.

AnuragThePathak (commented):
Apply the suggested changes and also update the documentation/website material and we should be sorted.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -341,7 +349,6 @@ func (r *RuleReadinessController) evaluateRuleForNode(ctx context.Context, rule
 		}
 
 		if !satisfied {
-			allConditionsSatisfied = false
 			metrics.ConditionEvaluationFailures.WithLabelValues(rule.Name, condReq.Type).Inc()
```
AnuragThePathak: Please combine this in the if block at line 338

docs/book/src/user-guide/concepts.md
```diff
@@ -1,128 +1,156 @@
...
+
+
+### Condition Evaluation Policy (conditionPolicy)
+
+When a rule specifies multiple conditions, the conditionPolicy determines how they are evaluated collectively to decide if the node should be tainted:
+
+- **llOf (Default)**: Every single condition listed in the rule must match its 
equiredStatus. If even one condition fails, the node is tainted. This is the standard behavior for ensuring all critical dependencies are healthy.
+- **nyOf**: At least ONE condition must match its 
equiredStatus. If no conditions match, the node is tainted. This is particularly useful for hardware that may be satisfied by multiple drivers, or systems with active/passive fallbacks.
```
AnuragThePathak: I see a lot of character issues. Not sure how you are running into this on the first place but fix across the whole PR.

docs/book/src/user-guide/concepts.md
```diff
@@ -1,128 +1,156 @@
...
-#### Use Case: Problem Gating (Default-Allow)
-By default, the controller operates in a "default-deny" fashion. If a guarded condition is missing from a new node, it resolves to `Unknown`, which typically does not satisfy `requiredStatus: True` or `requiredStatus: False`, causing the node to be tainted.
-
-However, for problem-detection conditions (like those reported by Node Problem Detector), you want a "default-allow" behavior: the node should start as healthy, and only be tainted if the problem condition explicitly triggers (e.g., `MaintenanceRequired=True`). 
-
-By setting `requiredStatus: False` and `defaultStatus: False`, an absent `MaintenanceRequired` condition evaluates to `False`. It matches the required status, allowing the node to bootstrap without being blocked by race conditions or initialization delays.
-
-#### Important Implication: Enforcement Modes
-The choice of `defaultStatus` has critical interaction with the rule's `enforcementMode`:
+# Concepts
```
AnuragThePathak: I was wondering how these unchanged changes are appearing in git diff?

docs/book/src/user-guide/getting-started.md
```diff
@@ -181,3 +181,5 @@ The Node Readiness Controller only 'reacts' to observed conditions on the Node o
 3.  **External Controllers**: Any tool that can patch Node status can trigger these rules.
 
 For a full example of setting up a custom condition for a security agent, see the [Security Agent Readiness Example](../examples/security-agent-readiness.md).
+
```
AnuragThePathak: same here. zero changes but two line insertion out of nowhere.

internal/controller/nodereadinessrule_controller.go
AnuragThePathak: move it to if !satisfied block line no 339

internal/controller/nodereadinessrule_controller.go
```diff
@@ -723,3 +737,4 @@ func (r *RuleReadinessController) getPreviousNodeEvaluation(rule *readinessv1alp
 	}
 	return nil
 }
+
```
AnuragThePathak: be wary of this extra lines. if actually needed fine but audit these extra lines across all your changes and remove if any. I see a major inconsistency of these.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -362,7 +362,7 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 
 		It("should count taintsToAdd when conditions unmet and taint absent", func() {
 			// Node matches selector, condition NOT satisfied, no pre-existing taint
-			// → shouldRemoveTaint=false, currentlyHasTaint=false → taintsToAdd++
+			// â†’ shouldRemoveTaint=false, currentlyHasTaint=false â†’ taintsToAdd++
```
AnuragThePathak: fix these character issues, I won't comment on rest of such occurrences.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2238,36 +2238,31 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
-				Status: corev1.NodeStatus{},
+				Spec: corev1.NodeSpec{
+					Taints: []corev1.Taint{
+						{Key: "readiness.k8s.io/GPUReady", Effect: corev1.TaintEffectNoSchedule},
+					},
+				},
+				Status: corev1.NodeStatus{
+					Conditions: []corev1.NodeCondition{
+						{Type: "gpu.example.com/HardwareDriverReady", Status: corev1.ConditionFalse},
+						{Type: "gpu.example.com/SoftwareFallbackReady", Status: corev1.ConditionFalse},
```
AnuragThePathak, line 2338: let's keep one of these true by default.

AnuragThePathak (changes requested):
Didn't review the actual doc updates due to git diff being impacted random + invalid character.

With that being said, do not ignore the comments on code.

And also note that you need to update docs/book/src/reference/api-spec.md file as well

docs/book/src/user-guide/concepts.md
vishnukothakapu: My editor accidentally evaluated the markdown backticks as escape sequences and added a UTF-8 BOM when saving, which caused the missing characters and messed up the diff. I've stripped the BOM, fixed the line endings, and restored the missing characters across all the doc files

docs/book/src/user-guide/concepts.md
vishnukothakapu: This happened because my editor changed the file's line endings from LF to CRLF while saving, so Git showed the whole file as modified. I've changed it back to LF, and the diff is clean now.

docs/book/src/user-guide/concepts.md
AnuragThePathak: This is a weird situation to be in but no worries. Take some time, apply the changes and review yourself before requesting review from others. Just a basic eye hover should work. Line by line, review again is not needed as long you properly audit everything before git commit.

docs/book/src/reference/api-spec.md
```diff
@@ -104,10 +123,10 @@ _Appears in:_
 
 | Field | Description | Default | Validation |
 | --- | --- | --- | --- |
-| `nodeName` _string_ | nodeName is the name of the evaluated Node. |  | MaxLength: 253 <br />MinLength: 1 <br />Pattern: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$` <br /> |
-| `conditionResults` _[ConditionEvaluationResult](#conditionevaluationresult) array_ | conditionResults provides a detailed breakdown of each condition evaluation<br />for this Node. This allows for granular auditing of which specific<br />criteria passed or failed during the rule assessment. |  | MaxItems: 5000 <br /> |
-| `taintStatus` _[TaintStatus](#taintstatus)_ | taintStatus represents the taint status on the Node, one of Present, Absent. |  | Enum: [Present Absent] <br /> |
-| `lastEvaluationTime` _[Time](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#time-v1-meta)_ | lastEvaluationTime is the timestamp when the controller last assessed this Node. |  |  |
+| `nodeName` _string_ | nodeName is the name of the evaluated Node. |  | MaxLength: 253 <br />MinLength: 1 <br />Pattern: `^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$` <br />Required: \{\} <br /> |
```
AnuragThePathak: May I know what purpose this \{\} is serving? It is added to every random line making it hard to review actual changes

docs/book/src/user-guide/concepts.md
```diff
@@ -133,18 +170,3 @@ When `spec.dryRun: true` is set on a rule:
 *   The intended actions are reported in the `status.dryRunResults` field of the `NodeReadinessRule`.
 
 This allows you to preview exactly which nodes would be affected and identifying any potential misconfigurations (like a typo in a label selector) before they impact your cluster.
-
-## Selecting Rules
```
AnuragThePathak, line 137: Why would you wipe this part?

internal/controller/nodereadinessrule_controller.go
```diff
@@ -606,25 +613,29 @@ func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readi
...
-		// Simulate rule evaluation
-		allConditionsSatisfied := true
+		// Simulate rule evaluation using the rule's conditionPolicy
+		conditionPolicy := rule.Spec.GetConditionPolicy()
 		missingConditions := 0
+		allSatisfied := true
+		anySatisfied := false
 
 		for _, condReq := range rule.Spec.Conditions {
-			currentStatus, conditionFound := r.getConditionStatus(
```
AnuragThePathak, line 615: I have already told to revert it as doesn't add any benefit in this PR other than introducing noise. It was split into two lines, because one single line overflows for smaller screens.

docs/book/src/user-guide/concepts.md
```diff
@@ -24,6 +24,43 @@ Typical taint keys look like:
 
 The segment after `readiness.k8s.io/` should describe the dependency or subsystem whose readiness is being guarded (for example, a CNI plugin, storage backend, or security agent). Treat this domain as reserved for the controller and closely related components, and avoid reusing it for unrelated taints.
 
+
+### Default Condition Status (`defaultStatus`)
```
AnuragThePathak: This part may be duplicated.

AnuragThePathak (changes requested):
Resolve each comment only after addressing them properly and verifying at your end.

AnuragThePathak:
@ajaysundark needed a help. How can we make netlify bot post a deployment preview URL?

docs/book/src/reference/api-spec.md
vishnukothakapu: Sorry about that. I regenerated api-spec.md using a different version of the docs generator, which introduced those Required: {} / Optional: {} entries across the file.  I'll revert those unrelated changes and keep the documentation update limited to the `conditionPolicy` additions. Thanks for catching it.

docs/book/src/user-guide/concepts.md
vishnukothakapu: This section got accidentally dropped during the merge conflict resolution with upstream/main. I've restored the original Selecting Rules section. Fixed now.

ajaysundark:
@AnuragThePathak AFAIU, it skips build if there are no changes since last build on the doc directory. Let me look into the PR status. One option is to unblock PR is to rerun the checks

ajaysundark:
Looks like the netlify checks are succeeding to me

AnuragThePathak:
For me it shows deploy preview canceled still and can't find a link for preview either.

docs/book/src/user-guide/concepts.md
AnuragThePathak: Did you forget to push your commits @vishnukothakapu?

docs/book/src/user-guide/getting-started.md
```diff
@@ -54,6 +54,7 @@ The `conditions` list defines the criteria. The controller watches the Node's st
 *   `type`: The exact string matching the NodeCondition type.
 *   `requiredStatus`: The status required (`True`, `False`, or `Unknown`).
 *   `defaultStatus`: (Optional) The status to assume if the condition is completely missing from the Node's status. (`True`, `False`, or `Unknown`, defaults to `Unknown` if not provided). Must not be set in `bootstrap-only` mode. (See [Concepts](./concepts.md#default-condition-status-defaultstatus) for usage implications).
+*   `conditionPolicy`: (Optional) Determines how conditions are evaluated. Set to `allOf` (default) to require every condition to match, or `anyOf` to require at least one condition to match. Must not be `anyOf` in `bootstrap-only` mode.
```
AnuragThePathak: I think this is not the right place to put conditionPolicy. It's not a child of ConditionRequirement after all.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -606,9 +613,11 @@ func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readi
 
 		affectedNodes++
 
-		// Simulate rule evaluation
-		allConditionsSatisfied := true
+		// Simulate rule evaluation using the rule's conditionPolicy
+		conditionPolicy := rule.Spec.GetConditionPolicy()
 		missingConditions := 0
+		allConditionsSatisfied := true
```
AnuragThePathak: For the sake of consistency, remove condition from the name here or add condition to other 3. All 4 of line no 328, 329, 619, and 620 should follow the same convention.

AnuragThePathak (changes requested):

AnuragThePathak (changes requested):
Almost there, hopefully last change needed 👍

internal/controller/nodereadinessrule_controller.go
vishnukothakapu: I standardized the variable names in both evaluateRuleForNode and processDryRun to use allSatisfied and anySatisfied consistently across both functions. Thanks!

docs/book/src/user-guide/getting-started.md
```diff
@@ -55,15 +55,20 @@ The `conditions` list defines the criteria. The controller watches the Node's st
 *   `requiredStatus`: The status required (`True`, `False`, or `Unknown`).
 *   `defaultStatus`: (Optional) The status to assume if the condition is completely missing from the Node's status. (`True`, `False`, or `Unknown`, defaults to `Unknown` if not provided). Must not be set in `bootstrap-only` mode. (See [Concepts](./concepts.md#default-condition-status-defaultstatus) for usage implications).
 
-### 3. Choose an Enforcement Mode
+### 3. Choose a Condition Policy
```
AnuragThePathak: Should be good but can make it slightly better by mentioning optional here only.

AnuragThePathak (commented):
/lgtm

Over to you @ajaysundark 

/assign @ajaysundark

docs/book/src/user-guide/getting-started.md
vishnukothakapu: Done! I've moved the (Optional) flag directly into the section header (### 3. Choose a Condition Policy (Optional)) so it's super clear right away. Thanks again for all the thorough reviews on this!

AnuragThePathak: Respect to you as well for having patience to make those changes again and again. Hope this journey helps you get better on the developer experience side going forward.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -99,6 +111,16 @@ type NodeReadinessRuleSpec struct {
...
 	NodeSelector metav1.LabelSelector `json:"nodeSelector,omitempty,omitzero"`
 
+	// conditionPolicy controls how the conditions list is evaluated.
+	// "allOf" (default) requires every condition to match its requiredStatus before the taint is removed.
+	// "anyOf" requires at least one condition to match its requiredStatus.
+	//
+	// Cannot be used with enforcementMode: bootstrap-only.
+	//
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="conditionPolicy is immutable"
```
ajaysundark: Hmm. This guardrail is weak. This seems to protect only when the field exists in both old and new objects. So a user can add conditionPolicy: anyOf to an existing rule (where the field doesn't exist). also can delete it from a rule using anyOf to flip behavior.

config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml
```diff
@@ -62,6 +62,20 @@ spec:
           spec:
             description: spec defines the desired state of NodeReadinessRule
             properties:
+              conditionPolicy:
+                description: |-
+                  conditionPolicy controls how the conditions list is evaluated.
+                  "allOf" (default) requires every condition to match its requiredStatus before the taint is removed.
+                  "anyOf" requires at least one condition to match its requiredStatus.
+
+                  Cannot be used with enforcementMode: bootstrap-only.
```
ajaysundark: nit: make it read explicitly as "anyOf cannot be used with bootstrap-only mode". You are not validating setting the field but only this combination

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2235,4 +2235,119 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 			Expect(failedNames).NotTo(ContainElement("stale-recovery-node"))
 		})
 	})
+
```
ajaysundark, line 2317: Can we test a scenario where an anyOf rule, with a single condition, configured with defaultStatus: True and requiredStatus: True. If the condition is missing what's the expected behavior?

internal/webhook/nodereadinessgaterule_webhook.go
```diff
@@ -88,6 +88,19 @@ func (w *NodeReadinessRuleWebhook) validateSpec(
 			}
 		}
 	}
+
+	// validate anyOf conditionPolicy is not combined with bootstrap-only mode.
```
ajaysundark, line 92: Can we also add tests for validating the webhook behavior?

ajaysundark (commented):
Thanks for looking into this feature, @vishnukothakapu. I had some comments, ptal!

Great work, @AnuragThePathak on the heavy lifting on the review!

internal/controller/nodereadinessrule_controller_test.go
AnuragThePathak: Regarding this one, actually the author added that initially and then it was my idea to drop it saying, getConditionStatus method anchors the ConditionRequirement as a unit and we have already unit tested all possible paths for that. And definitely unit tests for defaultStatus are already there.

Hence we can safely trust those tests do their job and only keep conditionPolicy restricted to what we code we write in conditionPolicy only. Probably something worth clarifying for me as well if that's not how test cases works.

internal/webhook/nodereadinessgaterule_webhook.go
AnuragThePathak: Speaking of this, I also remembered that we should update comment on line 74-75 with the addition of conditionPolicy below the return block.

api/v1alpha1/nodereadinessrule_types.go
vishnukothakapu: You're right, that's a gap in the CEL check. I've moved the validation into ValidateUpdate and now compare the effective policy via GetConditionPolicy(), so the implicit default case is covered as well.

internal/controller/nodereadinessrule_controller_test.go
vishnukothakapu: @AnuragThePathak is exactly right here, I actually had tests covering the cross-product of anyOf and defaultStatus initially, but we agreed to drop them to avoid test explosion. getConditionStatus perfectly encapsulates the defaultStatus logic and is heavily unit tested on its own, so ConditionPolicy just safely aggregates whatever boolean results bubble up from that. Let's keep the scope restricted to avoid unnecessary test bloat.

internal/webhook/nodereadinessgaterule_webhook.go
vishnukothakapu: @ajaysundark , I've added comprehensive test coverage to the webhook test suite verifying that anyOf + bootstrap-only is rejected, and that the new immutability checks work correctly.

@AnuragThePathak Done! I've updated that comment on line 74 to include conditionPolicy as well.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -115,10 +115,9 @@ type NodeReadinessRuleSpec struct {
 	// "allOf" (default) requires every condition to match its requiredStatus before the taint is removed.
 	// "anyOf" requires at least one condition to match its requiredStatus.
 	//
-	// Cannot be used with enforcementMode: bootstrap-only.
+	// anyOf cannot be used with enforcementMode: bootstrap-only.
 	//
 	// +optional
-	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="conditionPolicy is immutable"
```
AnuragThePathak:
```suggestion
// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.conditionPolicy) ? 'allOf' : oldSelf.conditionPolicy) == (!has(self.conditionPolicy) ? 'allOf' : self.conditionPolicy)",message="conditionPolicy is immutable"
```

Don't forget to add a comment while introducing something not everyone is not familiar with/not understandable at the first glance

internal/webhook/nodereadinessgaterule_webhook.go
```diff
@@ -211,6 +212,13 @@ func (w *NodeReadinessRuleWebhook) ValidateUpdate(ctx context.Context, oldRule,
 		return nil, fmt.Errorf("validation failed: %v", allErrs)
 	}
 
+	// Validate conditionPolicy immutability.
+	// CEL validation is insufficient because conditionPolicy has an implicit default (allOf) when omitted,
+	// and transitioning from omitted to 'anyOf' bypasses a simple `self == oldSelf` check.
+	if oldRule.Spec.GetConditionPolicy() != newRule.Spec.GetConditionPolicy() {
```
AnuragThePathak: Temporarily remove this and tests corresponding to it. We may ask to revert back to it, again.

internal/webhook/nodereadinessgaterule_webhook_test.go
```diff
@@ -173,6 +173,36 @@ var _ = Describe("NodeReadinessRule Validation Webhook", func() {
 			})
 		})
 
+		Context("conditionPolicy with bootstrap-only", func() {
```
AnuragThePathak, line 176: Whatever tests you added under this context is fine, but a few more tests are needed/modified. Refer to the tests here https://github.com/kubernetes-sigs/node-readiness-controller/pull/291/changes

internal/webhook/nodereadinessgaterule_webhook_test.go
```diff
@@ -173,6 +173,36 @@ var _ = Describe("NodeReadinessRule Validation Webhook", func() {
...
+
+			BeforeEach(func() {
+				spec = readinessv1alpha1.NodeReadinessRuleSpec{
+					NodeSelector: metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
+					ConditionPolicy: readinessv1alpha1.ConditionPolicyAnyOf,
+					EnforcementMode: readinessv1alpha1.EnforcementModeBootstrapOnly,
+				}
+			})
+
+			It("should forbid anyOf conditionPolicy with bootstrap-only enforcement", func() {
```
AnuragThePathak: Also it's a personal preference but I just feel ordering tests in the order of execution helps, like early return first and so on...

AnuragThePathak (changes requested):

internal/webhook/nodereadinessgaterule_webhook_test.go
vishnukothakapu: @AnuragThePathak Good point, reordered them to follow the execution flow (early return first, then continuous, then the actual failure case).

internal/webhook/nodereadinessgaterule_webhook_test.go
vishnukothakapu: @AnuragThePathak You're right, that makes sense. I went ahead and added a full test object for the anyOf + bootstrap-only rejection scenario alongside the defaultStatus one.

internal/webhook/nodereadinessgaterule_webhook.go
vishnukothakapu: @AnuragThePathak Done! I've removed the manual webhook validation and its tests for now. The new struct-level CEL validation handles this declaratively, so we don't need both. We can always add it back later if needed.

api/v1alpha1/nodereadinessrule_types.go
vishnukothakapu: @AnuragThePathak Thanks for pointing that out! I ended up moving the XValidation to the parent NodeReadinessRuleSpec since has() doesn't work as expected at the field level. Added a comment explaining why, and removed the redundant webhook validation and tests since the CEL rule covers it now.

AnuragThePathak:
@vishnukothakapu can you squash the 3 latest commits into one? That way we would be able to review the latest changes in one piece.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -46,6 +58,13 @@ const (
 )
 
 // NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
+//
+// We validate conditionPolicy immutability here at the struct level rather than the field level.
+// This is required because conditionPolicy has an implicit default ("allOf") when omitted. If a user
+// transitions from an omitted field to an explicit "anyOf", field-level transition rules are not
+// evaluated (since the field was previously absent), allowing the mutation to bypass validation.
+// By evaluating it at the struct level, we can safely use the `has()` macro to normalize absent values.
+// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.conditionPolicy) ? 'allOf' : oldSelf.conditionPolicy) == (!has(self.conditionPolicy) ? 'allOf' : self.conditionPolicy)",message="conditionPolicy is immutable"
```
AnuragThePathak, line 68: Not sure why would put this check here. It was meant to be used on ConditionPolicy as it's optional.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -99,6 +118,15 @@ type NodeReadinessRuleSpec struct {
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="nodeSelector is immutable"
 	NodeSelector metav1.LabelSelector `json:"nodeSelector,omitempty,omitzero"`
 
+	// conditionPolicy controls how the conditions list is evaluated.
```
AnuragThePathak, line 122: Also one heads-up: when you move the validation to the right place, see how you can manage Go-docs, comments, validation, and k8s manifest descriptions separate. That often is a challenging task to do.

AnuragThePathak (changes requested):

AnuragThePathak:
When you make changes next time, please amend to this latest commit only.

api/v1alpha1/nodereadinessrule_types.go
vishnukothakapu: @AnuragThePathak I agree it would be cleaner on the field, but this rule has to live at the struct level because has() needs self to be the parent object. On the field, self becomes the string value, which causes a CEL type-checking error. Putting it on the struct lets the validation work as intended.

api/v1alpha1/nodereadinessrule_types.go
vishnukothakapu: @AnuragThePathak Good point! Since the validation has to stay at the struct level anyway, the field docs remain clean. I also separated the explanation with a blank // so controller-gen doesn't include those implementation details in the generated CRD description.

api/v1alpha1/nodereadinessrule_types.go
AnuragThePathak: Ok no issue. Leave it as it is I'll review it tomorrow.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -58,6 +58,13 @@ const (
 )
 
 // NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
+//
+// We validate conditionPolicy immutability here at the struct level rather than the field level.
```
AnuragThePathak: Please update the comment to this

```We put the conditionPolicy immutability validation at `NodeReadinessRuleSpec` level instead of putting it on `conditionPolicy`. This is required because conditionPolicy is optional field. If a user transitions from an omitted field to an explicit "allOf" or vice-versa, field-level transition rules are not evaluated, since validations are only performed only when both self and oldSelf are present. By evaluating it at the struct level, we can safely use the `has()` macro to normalize absent values and force evaluation under any condition.```

Take care of the line breaks at your end.

File: api/v1alpha1/nodereadinessrule_types.go (hunk 44 above)
Comment (AnuragThePathak) [reply api/v1alpha1/nodereadinessrule_types.go:68]: You can mark it as resolve, seems like GitHub won't let me. I am assuming you tried tested running in a kubernetes cluster (locally) and tweaking the manifests.

File: internal/webhook/nodereadinessgaterule_webhook.go [hunk 47]
```diff
@@ -88,6 +88,19 @@ func (w *NodeReadinessRuleWebhook) validateSpec(
 			}
 		}
 	}
+
+	// validate anyOf conditionPolicy is not combined with bootstrap-only mode.
+	// The semantics are undefined: once the taint is permanently removed after
```
AnuragThePathak: Can you explain me briefly what you are trying to capture in line 93-96? I thought line 92 is more than enough.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2235,4 +2235,119 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+			Expect(k8sClient.Create(ctx, anyOfNode)).To(Succeed())
+			defer func() { Expect(k8sClient.Delete(ctx, anyOfNode)).To(Succeed()) }()
+			Expect(readinessController.evaluateRuleForNode(ctx, rule, anyOfNode)).To(Succeed())
+
+			// Taint should have been added because neither condition is satisfied
+			Expect(readinessController.hasTaintBySpec(anyOfNode, rule.Spec.Taint)).To(BeTrue())
+		})
+
+		It("allOf (explicit): keeps taint when only one of two conditions is satisfied", func() {
```
AnuragThePathak: keeps taint when not only conditions are conditions are satisfied

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2235,4 +2235,119 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+			anyOfNode.Status.Conditions = []corev1.NodeCondition{
```
AnuragThePathak: Declaring new conditions was not essential, we can simply use hardware software conditions.

Update rule too accordingly, same apply to taint.

One more idea I got while reviewing, we can actually declare the rule at context level and utilize beforeEach to assign a default value and perform minimal changes during tests. Should not take much time with a decent AI agent I think but worth simplifying.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2235,4 +2235,119 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+			}
+			readinessController.updateRuleCache(ctx, rule)
+
+			Expect(k8sClient.Create(ctx, anyOfNode)).To(Succeed())
+			defer func() { Expect(k8sClient.Delete(ctx, anyOfNode)).To(Succeed()) }()
+			Expect(readinessController.evaluateRuleForNode(ctx, rule, anyOfNode)).To(Succeed())
+
+			// Taint must remain because CondB is still False
+			Expect(readinessController.hasTaintBySpec(anyOfNode, rule.Spec.Taint)).To(BeTrue())
+		})
```
AnuragThePathak, line 2405: One more request let's add a test for allOf(explicit) removes taint when all conditions are satisfied. This would be a good addition because it future we may change default condition policy or multiple things can happen. So then this current setup will fumble.

AnuragThePathak (changes requested):
@vishnukothakapu we are in a very good shape to merge this. Just a little refactoring/improvement and we would be sorted. Thanks for being patient this far. I see we are nearing 100 conversations including bot comments 👀

internal/webhook/nodereadinessgaterule_webhook.go
vishnukothakapu: I mainly added that comment to explain why this combination is blocked for future readers. With bootstrap-only, the taint is removed permanently as soon as one condition is met, so combining it with anyOf could let a node bootstrap before the remaining components are initialized. If you'd rather keep the comments minimal, I'm happy to trim it down or remove it entirely.

api/v1alpha1/nodereadinessrule_types.go
AnuragThePathak: @vishnukothakapu sorry minor update on the first line please.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2318,36 +2308,34 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
 
-		It("allOf (explicit): keeps taint when only one of two conditions is satisfied", func() {
-			anyOfNode.Status.Conditions = []corev1.NodeCondition{
-				{Type: "example.com/CondA", Status: corev1.ConditionTrue},
-				{Type: "example.com/CondB", Status: corev1.ConditionFalse},
-			}
-			anyOfNode.Spec.Taints = []corev1.Taint{
-				{Key: "readiness.k8s.io/MultiCond", Effect: corev1.TaintEffectNoSchedule},
-			}
+		It("allOf (explicit): keeps taint when not all conditions are satisfied", func() {
```
AnuragThePathak: I saw that you are going for add taint instead of keeps above. Let's follow the same strategy here as well.

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2318,36 +2308,34 @@ var _ = Describe("NodeReadinessRule Controller", func() {
...
+			rule.Name = "allof-explicit-rule"
```
AnuragThePathak, line 2391: I would suggest name taint to condition-policy or something commonly usable. It's a headache to update it for each test.

AnuragThePathak (changes requested):
LGTM to just add these changes in a new commit. Including the above two in two resolved comments (https://github.com/kubernetes-sigs/node-readiness-controller/pull/315#discussion_r3737180811 https://github.com/kubernetes-sigs/node-readiness-controller/pull/315#discussion_r3737884062)

internal/webhook/nodereadinessgaterule_webhook.go
AnuragThePathak: Just trim it imo. The justification belongs in a different place not there. @vishnukothakapu

internal/controller/nodereadinessrule_controller_test.go
vishnukothakapu: Done! Cleared the taint first and updated the test to verify it adds it back.

internal/controller/nodereadinessrule_controller_test.go
vishnukothakapu: Good call, done! Renamed the taint to readiness.k8s.io/condition-policy.

docs/book/src/user-guide/getting-started.md
```diff
@@ -55,15 +55,20 @@ The `conditions` list defines the criteria. The controller watches the Node's st
 *   `requiredStatus`: The status required (`True`, `False`, or `Unknown`).
 *   `defaultStatus`: (Optional) The status to assume if the condition is completely missing from the Node's status. (`True`, `False`, or `Unknown`, defaults to `Unknown` if not provided). Must not be set in `bootstrap-only` mode. (See [Concepts](./concepts.md#default-condition-status-defaultstatus) for usage implications).
 
-### 3. Choose an Enforcement Mode
+### 3. Choose a Condition Policy (Optional)
+The `conditionPolicy` determines how the list of conditions is evaluated.
+*   **`allOf` (Default)**: Require every condition to match its required status.
+*   **`anyOf`**: Require at least one condition to match its required status. Must not be used in `bootstrap-only` mode.
+
```
AnuragThePathak, line 62: Put a link to ConditionPolicy in concepts like conditions and taint sections does.

AnuragThePathak (approved):
Rest looks good to me. Add this change in a new commit and request a review from Ajay from once done.

This PR surely will help me get better and efficient at reviewing large sized PRs going forward. Hope it also helps you be a better programmer (I have already seen significant improvement from start till end)

vishnukothakapu:
> Rest looks good to me. Add this change in a new commit and request a review from Ajay from once done.
> 
> This PR surely will help me get better and efficient at reviewing large sized PRs going forward. Hope it also helps you be a better programmer (I have already seen significant improvement from start till end)


@AnuragThePathak Thank you for the approval and for all the thoughtful reviews! I really appreciate the time and guidance throughout this PR. I learned a lot about Kubernetes controllers and writing better tests along the way. I just pushed the final docs commit and requested a review from @ajaysundark .

docs/book/src/user-guide/getting-started.md
AnuragThePathak: @vishnukothakapu just saw the update, I think `how conditions are evaluated` can be slightly misguiding saying something like "more details on conditionPolicy" or something similar can be better as how conditions are evaluated can also indicate how individual conditions are evaluated

vishnukothakapu: @AnuragThePathak Ah, good point! Changed the wording to explicitly say "more details on conditionPolicy" to avoid any ambiguity. Amended and pushed!

vishnukothakapu:
@ajaysundark The latest changes are ready for review. Anurag has reviewed and approved the PR, and suggested requesting your review as the next step. When you get a chance, could you please take a look? Thank you!

ajaysundark:
> Rest looks good to me. Add this change in a new commit and request a review from Ajay from once done.

Thanks for the review, @AnuragThePathak. This callout missed my attention as I filtering on the lgtm tag. prow commands expect a `/lgtm` to be tagged correctly.

ajaysundark:
@vishnukothakapu can you resolve conflicts?

api/v1alpha1/nodereadinessrule_types.go
ajaysundark: nice validation. this is better!

docs/book/src/reference/api-spec.md
```diff
@@ -165,6 +183,7 @@ _Appears in:_
...
 | `taint` _[Taint](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#taint-v1-core)_ | taint defines the specific Taint (Key, Value, and Effect) to be managed<br />on Nodes that meet the defined condition criteria. |  |  |
 | `nodeSelector` _[LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#labelselector-v1-meta)_ | nodeSelector limits the scope of this rule to a specific subset of Nodes. |  |  |
+| `conditionPolicy` _[ConditionPolicy](#conditionpolicy)_ | conditionPolicy controls how the conditions list is evaluated.<br />"allOf" (default) requires every condition to match its requiredStatus before the taint is removed.<br />"anyOf" requires at least one condition to match its requiredStatus.<br />Cannot be used with enforcementMode: bootstrap-only. |  | Enum: [allOf anyOf] <br /> |
```
ajaysundark: @vishnukothakapu does this need a regen again?

internal/controller/nodereadinessrule_controller_test.go
```diff
@@ -2235,4 +2235,110 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 			Expect(failedNames).NotTo(ContainElement("stale-recovery-node"))
 		})
 	})
+
```
ajaysundark, line 2317: @vishnukothakapu can we add a test case for the new CEL rule?

docs/book/src/reference/api-spec.md
vishnukothakapu: Yeah!, I forgot to run crd-ref-docs. I just regenerated the API spec docs and pushed the update.

docs/book/src/user-guide/concepts.md
```diff
@@ -24,6 +24,19 @@ Typical taint keys look like:
...
+
+- **`allOf` (Default)**: Every single condition listed in the rule must match its `requiredStatus`. If even one condition fails, the node is tainted. This is the standard behavior for ensuring all critical dependencies are healthy.
+- **`anyOf`**: At least ONE condition must match its `requiredStatus`. If no conditions match, the node is tainted. This is particularly useful for hardware that may be satisfied by multiple drivers, or systems with active/passive fallbacks.
+
+> [!IMPORTANT]
+> **`anyOf` is not supported with `bootstrap-only` rules.**
+>
+> `bootstrap-only` mode exists to verify that specific components have finished initializing. Using `anyOf` with `bootstrap-only` could lead to the node bootstrapping prematurely if just one component is ready, completely ignoring the initialization status of the others. The admission webhook enforces this restriction.
+
+
```
ajaysundark, line 39: Just noting the misconfiguration risk, I'm not sure if this is worth adding in a warning or adding a webhook validation (may be in a separate PR if we feel the same way, @AnuragThePathak) --

in anyOf rules, if defaultStatus and requiredStatus are both set as true, this could short-circuit the rule, without anyone realizing it. eg: HardwareDriverReady is missing but default status set as True (matching requiredStatus=True) would never taint the node.

internal/controller/nodereadinessrule_controller_test.go
vishnukothakapu: I just added a test case for the conditionPolicy immutability CEL rule and pushed the update.

docs/book/src/user-guide/concepts.md
vishnukothakapu: @ajaysundark Thanks for pointing that out!, You're totally right. If a condition defaults to its `requiredStatus` in an `anyOf` rule, the whole thing would silently short-circuit and never taint the node.

I'd be happy to add a CEL validation for this, making sure `defaultStatus` doesn't equal `requiredStatus` when the policy is `anyOf`.

@AnuragThePathak Do you want me to add this validation to this PR, or would you prefer we track it as a new issue and handle it in a follow-up PR so we can get this one merged first? Happy to do either!

AnuragThePathak: @vishnukothakapu I would be in favor of updating the issue with this information-> adding webhook validation (CEL also there but not straight-forward at all + we have failurePolicy = fail, hence webhook is guaranteed to guard) -> updating the docs.

Better to keep things atomic imo

cc @ajaysundark 

by the way regarding the lgtm label, apparently I do not have permission to use it so far. probably need to join the org, idk who they allow. Feel free to enlighten me if the process is not as complex as I am imagining.

vishnukothakapu: @AnuragThePathak Makes total sense! I agree that keeping this atomic is the best approach. I'll update the original issue with @ajaysundark 's edge case and we can tackle the Go webhook validation and docs update in a follow-up PR.

Assuming everything else looks good, is this PR good to go?

AnuragThePathak: Oh about this PR. Personally I was in favor of adding the validation and doc updates in this PR only so that if this feature goes live, people use it the right way since day 1. Were you two in favor of decoupling both changes?

AnuragThePathak: I have read your latest message on the issue. Looks good, wording can be improved (slightly lacks the precision) but rest ok. With that being said, also updating the issue adding a post script or something like that might be a good idea, just in case people decide to read just the top text.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -46,6 +58,14 @@ const (
 )
 
 // NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
+//
+// We put the conditionPolicy immutability validation at `NodeReadinessRuleSpec` level instead of
+// putting it on `conditionPolicy`. This is required because conditionPolicy is optional field. If a
+// user transitions from an omitted field to an explicit "allOf" or vice-versa, field-level
+// transition rules are not evaluated, since validations are only performed only when both self and
+// oldSelf are present. By evaluating it at the struct level, we can safely use the `has()` macro to
+// normalize absent values and force evaluation under any condition.
```
AnuragThePathak: Can we add a line break here?

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -99,6 +119,15 @@ type NodeReadinessRuleSpec struct {
...
 	NodeSelector metav1.LabelSelector `json:"nodeSelector,omitempty,omitzero"`
 
+	// conditionPolicy controls how the conditions list is evaluated.
+	// "allOf" (default) requires every condition to match its requiredStatus before the taint is removed.
+	// "anyOf" requires at least one condition to match its requiredStatus.
+	//
+	// anyOf cannot be used with enforcementMode: bootstrap-only.
+	//
+	// +optional
+	ConditionPolicy ConditionPolicy `json:"conditionPolicy,omitempty"` // Use GetConditionPolicy() for safe access; field may be empty even when allOf applies.
```
AnuragThePathak, line 129: One thing I was thinking that the schema doesn't state anywhere that conditionPolicy and defaultStatus together should not be used at the same time. Though genuinely not sure, if there is any right place to deliver that message within schema.

docs/book/src/reference/api-spec.md
```diff
@@ -27,18 +27,37 @@ _Appears in:_
...
+| `type` _string_ | type corresponds to the Node condition type being evaluated. |  | MaxLength: 316 <br />MinLength: 1 <br />Required: \{\} <br /> |
+| `currentStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | currentStatus is the actual status value observed on the Node, one of True, False, Unknown. |  | Enum: [True False Unknown] <br />Required: \{\} <br /> |
+| `requiredStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | requiredStatus is the status value defined in the rule that must be matched, one of True, False, Unknown. |  | Enum: [True False Unknown] <br />Required: \{\} <br /> |
+| `defaultStatus` _[ConditionStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.35/#conditionstatus-v1-core)_ | defaultStatus is the status a condition is evaluated to if the condition<br />is not found in a node. Reflects the defaultStatus configured in the rule<br />spec. |  | Enum: [True False Unknown] <br />Optional: \{\} <br /> |
```
AnuragThePathak: again these weird symbols 😐

docs/book/src/reference/api-spec.md
```diff
@@ -27,18 +27,37 @@ _Appears in:_
...
+
+
+#### ConditionPolicy
```
AnuragThePathak, line 36: This part looks good. But you have to deal with those unwanted characters

docs/book/src/user-guide/concepts.md
```diff
@@ -24,6 +24,19 @@ Typical taint keys look like:
...
 
+### Condition Evaluation Policy (`conditionPolicy`)
+
+When a rule specifies multiple conditions, the `conditionPolicy` determines how they are evaluated collectively to decide if the node should be tainted:
+
+- **`allOf` (Default)**: Every single condition listed in the rule must match its `requiredStatus`. If even one condition fails, the node is tainted. This is the standard behavior for ensuring all critical dependencies are healthy.
+- **`anyOf`**: At least ONE condition must match its `requiredStatus`. If no conditions match, the node is tainted. This is particularly useful for hardware that may be satisfied by multiple drivers, or systems with active/passive fallbacks.
+
+> [!IMPORTANT]
+> **`anyOf` is not supported with `bootstrap-only` rules.**
```
AnuragThePathak, line 35: Should update docs saying conditionPolicy anyOf and defaultStatus are not to be used at the same time (or whatever rule you set in the validation).

AnuragThePathak (commented):
Seems like I ended up reviewing earlier by mistake. But fix the doc issue and one extra line mentioned (go through all comments once)

api/v1alpha1/nodereadinessrule_types.go
vishnukothakapu: @AnuragThePathak , I think the most natural place would be to add a note to the conditionPolicy field comment here, something like "When using anyOf, ensure that no condition defines a defaultStatus equal to its requiredStatus, as this would cause the rule to silently short-circuit." I'll include that along with the webhook validation and docs update in the follow-up PR.

docs/book/src/reference/api-spec.md
vishnukothakapu: @AnuragThePathak Yeah, those symbols are coming from the crd-ref-docs tool itself. this file is fully auto-generated, so I can't clean them up without the changes being overwritten on the next regen. This seems like a pre-existing issue with the tool's markdown renderer.

vishnukothakapu: Update: Fixed! The \{\} symbols were being injected by the newer version of crd-ref-docs (v0.3.0). I regenerated the file and ran a post-processing step to strip those artifacts.

docs/book/src/user-guide/concepts.md
vishnukothakapu: @AnuragThePathak Noted! I'll add that to the docs as part of the follow-up PR alongside the webhook validation.

docs/book/src/reference/api-spec.md
AnuragThePathak: Hope you didn't manually remove the symbols @vishnukothakapu. Because we need a permanent solution not a temporary one

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -59,12 +59,6 @@ const (
 
 // NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
 //
-// We put the conditionPolicy immutability validation at `NodeReadinessRuleSpec` level instead of
```
AnuragThePathak: Was it anywhere suggested to remove this altogether?

docs/book/src/reference/api-spec.md
vishnukothakapu: @AnuragThePathak ,I stripped them via a post-processing step after regeneration, not a permanent fix. The \{\} symbols are introduced by crd-ref-docs v0.3.0 and weren't present in older versions.

I think the cleanest permanent fix would be to add a sed post-processing step to the crd-ref-docs Makefile target so the symbols are automatically stripped on every regen. Happy to open a separate issue/PR for that.

api/v1alpha1/nodereadinessrule_types.go
vishnukothakapu: @AnuragThePathak , that was completely my mistake during the docs cleanup. I was trying to move the implementation comment out of the // doc block so it wouldn’t end up in the auto-generated api-spec.md, but I accidentally deleted it instead of changing it to /* */. I’ve restored it properly now. Sorry about the confusion!

AnuragThePathak: I don't think that comment is captured by any auto generation tools. They rather capture the comment when you use that struct (the field where that structure serves as a type).

docs/book/src/reference/api-spec.md
AnuragThePathak: I see. In that case, confirm once if crd-ref-docs has some bug, or this is actually how it's meant to be. If it's how it's meant to be.

If that's how it's meant to be then maybe we will require one PR purely to run crd tool only so git diff remains clean on future PRs.

If bug then I would prefer using older crd-ref-doc version. Also as current solution if older version works better, otherwise what you have should also works.

api/v1alpha1/nodereadinessrule_types.go
vishnukothakapu: @AnuragThePathak I actually thought the same thing at first! But when I ran the crd-ref-docs generator locally, I noticed it was pulling that entire implementation paragraph into the public api-spec.md under the NodeReadinessRuleSpec section. Wrapping it in /* */ seemed to do the trick to hide it from the generator while keeping the context in the code.

docs/book/src/reference/api-spec.md
vishnukothakapu: @AnuragThePathak I looked into it, and it appears to be a rendering bug in newer versions of the crd-ref-docs markdown template when it encounters empty default values or markers.

Since the manual strip I just pushed perfectly cleans up the api-spec.md without polluting the git diff for this PR, let's stick with this for now! I'll investigate the best permanent fix (either pinning to an older working version or adding a sed command to the Makefile) when we tackle the webhook follow-up PR.

With that fixed, I believe everything else is fully addressed! Let me know if there's anything else you'd like me to look at.

api/v1alpha1/nodereadinessrule_types.go
AnuragThePathak: Use of /* */ doesn't seem necessary. We can simply it to top keeping a proper gap and it should be fine.

```go
// Note for Developers: conditionPolicy immutability validation is placed at the
// NodeReadinessRuleSpec level instead of conditionPolicy because conditionPolicy is optional.
// When transitioning between omitted and explicit "allOf", field-level transition rules are
// bypassed since CEL only evaluates them when both self and oldSelf are present. Evaluating at
// the struct level allows using has() to normalize absent values.

// NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
//
// +kubebuilder:validation:XValidation:rule="(!has(oldSelf.conditionPolicy) ? 'allOf' : oldSelf.conditionPolicy) == (!has(self.conditionPolicy) ? 'allOf' : self.conditionPolicy)",message="conditionPolicy is immutable"
type NodeReadinessRuleSpec struct {
```

docs/book/src/reference/api-spec.md
AnuragThePathak: Ok cool. Not a blocker for this PR in that case. I will handle that separately.

Regarding your question, you still have to restore the comment before NodeReadinessRuleSpec

AnuragThePathak (approved):
LGTM. I would suggest creating a new issue for defaultStatus + conditionPolicy follow-up when you see this message considering the existing issue will automatically be closed once we merge this PR.

vishnukothakapu:
> LGTM. I would suggest creating a new issue for defaultStatus + conditionPolicy follow-up when you see this message considering the existing issue will automatically be closed once we merge this PR.

@AnuragThePathak Thank you for the review and approval! I will open the new issue for the `defaultStatus + conditionPolicy` follow-up as soon as this is merged.

AnuragThePathak:
@vishnukothakapu ideally speaking there's no rule like having to wait for this PR to get merged to actually create the new issue. But anyways it's your preference, since the details are logged properly under current issue, I won't push further on creating the issue right away.

For your information @ajaysundark maybe you can take a look whenever possible. I am still figuring out on accessing the prow commands by the way. @vishnukothakapu is in favor of creating a separate PR for the restriction on using defaultStatus and conditionPolicy at the same time. I think that will be convenient for us to review as well. Hopefully that's not an issue for release?

internal/controller/nodereadinessrule_controller.go
ajaysundark, line 361: @vishnukothakapu is it possible to capture the rule is satisfied with anyOf requirement here in this log (or a separate log?)

ajaysundark (approved):
@vishnukothakapu Thanks for your contribution! 

/lgtm
/approve

internal/controller/nodereadinessrule_controller.go
vishnukothakapu: @ajaysundark , The Evaluation result log that prints immediately after this loop actually captures exactly that (it explicitly logs both "conditionsSatisfied": true and "conditionPolicy": "anyOf").

However, if you think it would be helpful to also have conditionPolicy printed inside this per-condition log for easier debugging, I can definitely add that in the next follow-up PR! Let me know what you prefer.

vishnukothakapu:
@AnuragThePathak @ajaysundark Thank you both for your thorough reviews and guidance throughout this PR! I really appreciate the time you took to help shape this feature. I've opened the follow-up issue for the `defaultStatus` webhook validation and will get started on that next!

PR #318, opened 2026-07-21
Node Readiness Daemon

ajaysundark:
## Description

NodeReadinessDaemon - a low-latency node-local daemon subscribing pod-readiness information from kubelet over the `/pods-api` grpc channel and updates the node-conditions.

Dependencies:
- k8s 1.37
- controller-runtime v0.24

Doc: https://docs.google.com/document/d/1r0n0c8wb-8PY9zeBDjG_JJluDVtfSOHW0xKdlOtOWaU/edit?tab=t.0#heading=h.4rwasj1qznfq

Assisted by AI: Claude-Code and Gemini

## Related Issue

None

## Type of Change

/kind feature
/kind design

## Testing
Tested in kind cluster

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
feat: node-readiness-daemon for low-latency reliable updates from the node. 
```
Doc <TBD>


DsThakurRawat:
Ran this branch locally against `upstream/main` (`af1f57d`) — `go vet`, `staticcheck`, and `go test -race ./internal/daemon/...` are all clean. One reachable panic on the flag path:

**`--reconcile-interval` and `--resync-interval` accept negative durations, which panic the daemon at startup.**

`New` normalises only the zero value:

```go
// internal/daemon/daemon.go
if opts.ReconcileInterval == 0 { opts.ReconcileInterval = DefaultReconcileInterval }
if opts.ResyncInterval == 0    { opts.ResyncInterval    = DefaultResyncInterval }
```

so a negative flag flows through untouched into:

```go
ticker := d.clk.NewTicker(d.reconcileInterval)
resync := d.clk.NewTicker(d.resyncInterval)
```

and `time.NewTicker` panics on `d <= 0`. Reproduced on go1.26.5:

```
=== RUN   TestReproNegativeIntervalPanics/negative_reconcile-interval
    PANIC: non-positive interval for NewTicker
=== RUN   TestReproNegativeIntervalPanics/negative_resync-interval
    PANIC: non-positive interval for NewTicker
```

Worth fixing rather than filing under "don't do that": the daemon is a DaemonSet, so a single `--reconcile-interval=-10s` in a manifest puts *every* node's daemon into CrashLoopBackOff, and the only diagnostic is a bare runtime panic with no mention of the flag that caused it.

It is also inconsistent with the config path in this same PR, which already gets it right:

```go
// internal/daemon/config.go — parseThreshold
if d < 0 { return 0, fmt.Errorf("must not be negative") }
```

Durations parsed from YAML are validated; durations parsed from flags are not.

Smallest fix that keeps `New`'s existing normalise-in-place convention:

```go
if opts.ReconcileInterval <= 0 { opts.ReconcileInterval = DefaultReconcileInterval }
if opts.ResyncInterval    <= 0 { opts.ResyncInterval    = DefaultResyncInterval }
```

That silently corrects, matching what `New` already does for `0`. If you would rather fail loudly on an operator typo — arguably better, since a silently-ignored interval is its own surprise — the alternative is a check in `run()` alongside the existing `node-name` validation. Happy to send either as a PR if it's useful.

AvineshTripathi:
Could we consolidate RCR + NRD into a single daemon?

Both daemons do the same thing: publish node conditions from different sources. They duplicate a lot of core logic (signal handling, reconciliation loop, node condition writes, metrics, graceful shutdown).

Would it make sense to have one daemon with evaluator modes? e.g., `--mode=pod` for pod health (NRD), `--mode=http` for HTTP checks (RCR), `--mode=pod,http` to run both.
 
**Benefits**: less code duplication, single DaemonSet instead of 2, unified config/metrics/logging, easier to add new health sources in the future.

p.s. maybe NPD integration to the daemon as new source could be a thing

ajaysundark:
Thanks for the review @AvineshTripathi .

> Could we consolidate RCR + NRD into a single daemon?
> Both daemons do the same thing: publish node conditions from different sources

It is intentional that they are functionally identical. I think the goal for NRD is to eventually supersede [readiness-reporter](https://github.com/kubernetes-sigs/node-readiness-controller/tree/main/cmd/readiness-condition-reporter).

The current readiness reporter has few design down-sides:
1. When deployed as DS: it requires host networking, also requiring the pod healthz endpoint to be exposed on the host, which is not ideal
2. When deployed as a side-car: the readiness monitoring is tightly 'coupled' with the target component, which has risks as when the component crashes it also brings down the health-watcher with it.

That said, there may be use-cases where a simple readiness-reporter could still be beneficial eg: NRD is closely tied to Kubelet's pod grpc api, however readiness-reporter can probe 'any' http endpoint (but that is also overlapping with NPD capabilities). In the long-run, the readiness-reporter could benefit as a standard http-probe directly within the Node Problem Detector (avoids double deployment) project itself. see ref: https://github.com/kubernetes/node-problem-detector/issues/991

cc @hakman @Priyankasaggu11929

ajaysundark:
> --reconcile-interval and --resync-interval accept negative durations, which panic the daemon at startup.

Thanks for the catch, @DsThakurRawat . I'll look into it!

PR #320, opened 2026-07-22
fix: record conflict failures and fix missing status patch metric

rawadhossain:
## Description

Adds `node_readiness_api_conflicts_total`, tracking every conflict and retry across 7 places: 
- adding/removing a taint
- marking bootstrap done
- adding/removing the rule finalizer
- updating a node's status or the whole rule's status.

Earlier it was just logs and most of these conflicts couldn't actually happen. Now that #343 is in, conflicts are real, so bringing the metric to handle them.

### What's added
- `node_readiness_api_conflicts_total` with all 7 operations, each tested with a real forced conflict.
- New failure reason `RuleStatusRuleSweepConflictExhausted`: full rule status update can exhaust its retries under load, and that wasn't tracked before.
- `monitoring.md` updated.

## Type of Change

/kind feature

## Checklist

- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -153,8 +154,15 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			"ruleResourceVersion", rule.ResourceVersion)
 
 		if err := r.evaluateRuleForNode(ctx, rule, node); err != nil {
+			operation := "unknown"
+			switch {
+			case errors.Is(err, errAddTaintFailed):
+				operation = "add_taint"
+			case errors.Is(err, errRemoveTaintFailed):
+				operation = "remove_taint"
+			}
 			log.Error(err, "Failed to evaluate rule for node",
-				"node", node.Name, "rule", rule.Name)
+				"node", node.Name, "rule", rule.Name, "operation", operation)
 			// Continue with other rules even if one fails
 			r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
 			errs = append(errs, err)
@@ -223,6 +231,11 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 		})
 
 		if err != nil {
+			reason := "StatusPatchError"
+			if apierrors.IsConflict(err) {
+				reason = "StatusPatchConflictExhausted"
+			}
+			metrics.Failures.WithLabelValues(rule.Name, reason).Inc()
 			log.Error(err, "Failed to update rule status after node evaluation",
 				"node", node.Name,
 				"rule", rule.Name,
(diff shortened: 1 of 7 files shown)


internal/controller/node_controller.go
```diff
@@ -153,8 +154,15 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			"ruleResourceVersion", rule.ResourceVersion)
 
 		if err := r.evaluateRuleForNode(ctx, rule, node); err != nil {
+			operation := "unknown"
```
AvineshTripathi: can we make a util function here?

internal/controller/node_controller.go
```diff
@@ -153,8 +154,15 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
...
+			switch {
+			case errors.Is(err, errAddTaintFailed):
+				operation = "add_taint"
+			case errors.Is(err, errRemoveTaintFailed):
+				operation = "remove_taint"
+			}
 			log.Error(err, "Failed to evaluate rule for node",
```
AvineshTripathi, line 157: I don't think adding operations would add value. We already are logging the error. If the idea here is to standardise error messaging, then that is a whole different discussion.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -286,7 +293,14 @@ func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, ru
...
 			if err := r.evaluateRuleForNode(ctx, rule, &node); err != nil {
-				log.Error(err, "Failed to evaluate node for rule", "rule", rule.Name, "node", node.Name)
+				operation := "unknown"
+				switch {
+				case errors.Is(err, errAddTaintFailed):
+					operation = "add_taint"
+				case errors.Is(err, errRemoveTaintFailed):
+					operation = "remove_taint"
+				}
+				log.Error(err, "Failed to evaluate node for rule", "rule", rule.Name, "node", node.Name, "operation", operation)
```
AvineshTripathi: same as above. we already log errors.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -668,7 +690,7 @@ func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readi
 func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule, nodeList *corev1.NodeList) error {
 	log := ctrl.LoggerFrom(ctx)
 
-	var errors []string
+	var cleanupErrors []string
```
AvineshTripathi: any particular reason to change this?

AvineshTripathi (commented):
left my initial thoughts here!

AvineshTripathi:
cc @Priyankasaggu11929 I remember you had some thoughts on this

rawadhossain:
Hi @AvineshTripathi
gave this a fresh look. Actually built `api_conflicts_total` locally and live tested it to see if it was worth keeping. Most of what it would show either duplicates the `ConflictExhausted` reasons already in this PR, or just stay at zero where conflicts don't happen.
So I reworked this PR to go with the logs approach instead, which also addresses your earlier comments. PTAL.

ajaysundark:
The largest source of api conflicts (409s) don't rise up because they are handled as merge patch. We should check this after #343 is merged, only after when status conflicts will 'surface'

rawadhossain:
cc @ajaysundark for some comments to get holistic context.

rawadhossain:
@ajaysundark
added the metric and possible conflict operations. Tested every scenario, and working as expected. 
Lmk if any changes needed. Thanks!

rawadhossain:
cc @ajaysundark for review. Thanks.

PR #322, opened 2026-07-24
add --pprof-bind-address flag to NRC controller

Priyankasaggu11929:
With the new flag, we can expose the NRC controller's pprof HTTP endpoints, and capture heap and CPU profiles from the running NRC instance during our scale test harness.

controller-runtime project provides this with [PprofBindAddress](https://github.com/kubernetes-sigs/controller-runtime/blob/efff3590e02a2a27059a57a9535e87d829c3cc77/pkg/manager/manager.go#L248-L253) (so, we don't need to setup custom server)

Many  projects in the kubernetes ecosystem are using it - https://cs.k8s.io/?q=PprofBindAddress&i=nope&literal=nope&files=&excludeFiles=&repos=

---

A tiny tutorial on how to use it - https://book.kubebuilder.io/reference/pprof-tutorial.html, cc: @vitorfloriano

It will be useful for diagnosing memory growth and CPU hotspots during scale tests.

## Type of Change

/kind feature

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -64,6 +64,7 @@ var (
 	metricsCertDir           string
 	leaderElectionNamespace  string
 	enableNodeStateMetrics   bool
+	pprofAddr                string
 	kubeAPIQPS               float64
 	kubeAPIBurst             int
 	nodeConcurrentReconciles int
@@ -87,6 +88,7 @@ func main() {
 	flag.StringVar(&metricsCertDir, "metrics-cert-dir", "",
 		"The directory where the certificates for metrics are located.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
+	flag.StringVar(&pprofAddr, "pprof-bind-address", "", "The address the pprof endpoint binds to. Leave empty to disable.")
 	flag.BoolVar(&enableLeaderElection, "leader-elect", false,
 		"Enable leader election for controller manager. "+
 			"Enabling this will ensure there is only one active controller manager.")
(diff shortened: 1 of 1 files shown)


vitorfloriano:
Very small patch for so much insightful data!

As a total newbie in profiling, I just wonder how we would add the profile data to the report, since it seems that we would need `go tool pprof` to curl the endpoint and serve it or output it in a format to be interpreted by us. I'll need to research. 🤔

Priyankasaggu11929:
> As a total newbie in profiling, I just wonder how we would add the profile data to the report, since it seems that we would need `go tool pprof` to curl the endpoint and serve it or output it in a format to be interpreted by us. I'll need to research. 🤔

using the same running example prow job, i saw them exporting pprof artifacts here - https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kops/18537/pull-kops-gce-master-scale-kindnet-performance-100/2073598788685008896/artifacts/

for visualization purpose, we will hopefully be displaying them as part of the same UI (for other metrics).


we can start with a much much simpler page but for idea sake - it would look similar to this - https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_agent-sandbox/1184/presubmit-agent-sandbox-benchmarks-kops-gcp/2077450067362451456/artifacts/stress-report/pprof.html

ajaysundark:
/lgtm
/approve

Thanks! we should document how to work with pprof somewhere as well.

PR #219, opened 2026-05-06
fix: increment metrics.Failures counter on node reconciler evaluation error

mohanakatari119-bit:
`processNodeAgainstAllRules` already called `recordNodeFailure` on evaluation errors but never incremented the `metrics.Failures` counter, so the `node_readiness_failures_total` metric was not updated for this path.

Added the missing `metrics.Failures.WithLabelValues(rule.Name, "EvaluationError").Inc()` call right after `recordNodeFailure`, matching the pattern used elsewhere in the controller.

The new test covers this path by intercepting `Patch` on node objects to force an evaluation error, then reading the counter before and after `processNodeAgainstAllRules` to confirm it incremented.

PR #328, opened 2026-07-30
fix lint failure

ajaysundark:
Fixing lint-error from #219

### DIFF
--- a/internal/controller/node_controller_test.go
+++ b/internal/controller/node_controller_test.go
@@ -1144,7 +1144,8 @@ var _ = Describe("Node Controller", func() {
 			_ = metrics.Failures.WithLabelValues(rule.Name, "EvaluationError").Write(beforeM)
 			before := beforeM.GetCounter().GetValue()
 
-			controller.processNodeAgainstAllRules(ctx, node)
+			err := controller.processNodeAgainstAllRules(ctx, node)
+			Expect(err).To(HaveOccurred())
 
 			afterM := &dto.Metric{}
 			_ = metrics.Failures.WithLabelValues(rule.Name, "EvaluationError").Write(afterM)


AnuragThePathak (approved):
LGTM

internal/controller/node_controller_test.go
```diff
@@ -1144,7 +1144,8 @@ var _ = Describe("Node Controller", func() {
 			_ = metrics.Failures.WithLabelValues(rule.Name, "EvaluationError").Write(beforeM)
 			before := beforeM.GetCounter().GetValue()
 
-			controller.processNodeAgainstAllRules(ctx, node)
+			err := controller.processNodeAgainstAllRules(ctx, node)
+			Expect(err).To(HaveOccurred())
```
vitorfloriano, line 1148: Shouldn't this be:
```suggestion
			Expect(err).NotTo(HaveOccurred())
```
Also, interesting discussion on using this pattern: https://github.com/kubernetes/kubernetes/issues/109600 

We're not using the e2e test framework, but still, something to ponder.

vitorfloriano: /hold

vitorfloriano: Oh I hadn't read the whole It block. My apologies. 

/unhold

ajaysundark: hmm. `ExpectNoError(err error, contextMsg string, explain ...interface{})` is interesting. I wasn't aware, thanks!
That'd be a good clean-up story fit for a good-first-issue for the project, if it's already available.

ajaysundark:
> LGTM

Prow require to use `/lgtm` (slashed prefix) to be detected :)

PR #244, opened 2026-05-13
fix: add rule label to node_readiness_evaluation_duration_seconds metric

Shreya2005-2005:
## Description
node_readiness_evaluation_duration_seconds was defined as a plain
Histogram with no labels. All other per-rule metrics include a rule
label. Without it, operators cannot identify which rule is slow in
multi-rule clusters.

## Related Issue
Fixes #243 

## Type of Change
/kind bug

## Testing
- go build ./... passes
- make lint passes (0 issues)
- go test ./internal/webhook/... passes (25/25)

## Checklist
- [x] make test passes
- [x] make lint passes

## Does this PR introduce a user-facing change?
```release-note
adds rule label to node_readiness_evaluation_duration_seconds metric
```

Issue #269, opened 2026-05-30
testing: Add Test Coverage for Prometheus Metrics


chinma-yyy:
## Problem
We're exposing 5 Prometheus metrics from the controller, but only 1 of them (`BootstrapCompleted`) has test coverage. This makes it harder to catch regressions when metrics behavior changes.

## Current Metrics Status

- ✅ `BootstrapCompleted` - Has tests
- ❌ `RulesTotal` - No tests
- ❌ `TaintOperations` - No tests  
- ❌ `Failures` - No tests
- ❌ `EvaluationDuration` - No tests (and not even used in code yet)

### 1. Add tests for `RulesTotal` metric
This gauge tracks the total number of active rules. We should test:
- Value increases when a rule is added to the cache
- Value decreases when a rule is removed from the cache

### 2. Add tests for `TaintOperations` metric
This counter tracks taint add/remove operations. We should test:
- Counter increments when adding a taint (with labels: `rule`, `operation="add"`)
- Counter increments when removing a taint (with labels: `rule`, `operation="remove"`)

### 3. Add tests for `Failures` metric
This counter tracks operational failures. We should test:
- Counter increments on evaluation errors (label: `reason="EvaluationError"`)
- Counter increments on taint operation failures (labels: `reason="AddTaintError"` or `reason="RemoveTaintError"`)

### 4. Implement and test `EvaluationDuration` metric
This histogram is defined but never used. We should:
- Add instrumentation to the `evaluateRuleForNode()` function to record evaluation duration
- Add tests to verify the histogram records evaluations

## Acceptance Criteria

- [ ] All 5 metrics have unit test coverage
- [ ] Tests verify metrics behave correctly in both success and failure scenarios
- [ ] All tests pass with `make test`

ajaysundark:
cc @rawadhossain could you help take a first look at this? Thanks

ajaysundark:
cc @AvineshTripathi for guidance

rawadhossain:
Hi, thanks for raising the issue!
I took a look at the current implementation and noticed a few things.

The description contains a factual error regarding `EvaluationDuration`. The metric is already instrumented in `evaluateRuleForNode()` using `prometheus.NewTimer(metrics.EvaluationDuration)` and `ObserveDuration()`, so the work here would be adding test coverage rather than implementing the metric.

[...]

PR #330, opened 2026-07-31
fix: clean up evaluation duration metric on rule deletion

LightCreator1007:
## Description
When a rule is deleted, `reconcileDelete` removes that rule's metrics so they don't pile up.
`node_readiness_evaluation_duration_seconds` is missing from that list.

It gained a `rule` label in #244. Before that it had no labels and nothing per-rule to remove,
so leaving it out was correct. Once it became per-rule it needed removing too, but the list
wasn't updated.

So the metric for a deleted rule is never dropped: the memory is never freed, and deleted rules
keep showing up,

This adds it to the existing list.
## Related Issue

None

## Type of Change

/kind bug
/kind regression

## Testing

verified with a temporary test (not included in this PR): create a rule, reconcile it,
delete it, then check what is still exported for that rule name. `RuleLastReconciliationTime`
serves as a control, since it is already in the cleanup list.

Before the fix:
before delete: evaluation_duration=1   last_reconciliation=1
after  delete: evaluation_duration=1   last_reconciliation=0

After the fix, `evaluation_duration` drops to 0 on deletion, matching the control.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fixed a leak where node_readiness_evaluation_duration_seconds was kept for deleted
NodeReadinessRules
```

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -212,6 +212,7 @@ func (r *RuleReconciler) reconcileDelete(ctx context.Context, rule *readinessv1a
 	metrics.RuleLastReconciliationTime.DeleteLabelValues(rule.Name)
 	metrics.BootstrapCompleted.DeleteLabelValues(rule.Name)
 	metrics.BootstrapDuration.DeleteLabelValues(rule.Name)
+	metrics.EvaluationDuration.DeleteLabelValues(rule.Name)
 
 	// For multi-label metrics, use DeletePartialMatch to wipe all combinations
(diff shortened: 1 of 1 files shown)


ajaysundark:
Thanks for catching this!

/lgtm

cc @rawadhossain @AvineshTripathi  another reason to improve metrics tests (#269).

ajaysundark:
> verified with a temporary test (not included in this PR

@LightCreator1007 can you add this test case?

ajaysundark:
Ill merge this for now.  @rawadhossain please handle the metrics test followup together in #269

PR #332, opened 2026-08-01
fix: isolate rule status cache per rule-reconcile workers

bhuvan-somisetty:
## Description

This PR resolves data races and memory leaks in the controller when `--node-concurrent-reconciles` is set to > 1:

1. **Rule Cache Data Race Fix**: Updated `RuleReadinessController.getApplicableRulesForNode` to return `rule.DeepCopy()` pointers. Previously, returning pointers straight from `r.ruleCache` caused concurrent `NodeReconciler` worker threads to mutate `rule.Status.NodeEvaluations` and `rule.Status.FailedNodes` on the shared in-memory pointer without synchronization, resulting in Go data races and memory corruption under concurrent node reconciliation load.
~~2. **Prometheus Metric Leak Fix**: Added `metrics.EvaluationDuration.DeleteLabelValues(rule.Name)` to `reconcileDelete` in `nodereadinessrule_controller.go` to ensure histogram metric series are unregistered when a rule CR is deleted.~~
3. **Unit Tests**: Added `TestGetApplicableRulesForNode_DeepCopy` in `helper_unit_test.go` to verify that `getApplicableRulesForNode` returns isolated, deep-copied rule instances.

## Related Issue

Fixes #331

## Type of Change

/kind bug

## Testing

- Unit testing via `go test -v ./internal/controller/...`
- Verified deep-copy isolation unit test passes without race conditions.

## Checklist

- [x] `make test` passes
- [x] `make lint` passes

## Does this PR introduce a user-facing change?

```release-note
NONE
```

PR #333, opened 2026-08-01
fix: clear stale FailedNodes entry when a node recovers

bhuvan-somisetty:
## Description

This is a follow-up to #332, extracted per @ajaysundark's request to review #209 and evaluate the `recordFailure`/stale `FailedNodes` patch separately from the ruleCache data-race fix.

`processNodeAgainstAllRules` (the `NodeReconciler` path in `node_controller.go`) records a `NodeFailure` via `recordNodeFailure` when `evaluateRuleForNode` errors, but never clears it when the node later evaluates successfully. Since `rule` here is a deep-copied snapshot carried over from `ruleCache` (per #332), a stale failure entry from a prior reconcile keeps getting copied forward into `rule.Status.FailedNodes` on every subsequent successful evaluation, so a recovered node is stuck listed as failed indefinitely.

`nodereadinessrule_controller.go`'s `processAllNodesForRule` (the `RuleReconciler` path) already clears `FailedNodes` inline on the success branch, so it isn't affected — only the `NodeReconciler` path needed the fix.

This adds a `clearNodeFailure` helper (matching the existing `recordNodeFailure` pattern) and calls it on the success branch of `processNodeAgainstAllRules`.

## Related

Follow-up to #332, evaluated from #209.

## Type of Change

/kind bug

## Testing

- `go build ./...`
- `go vet ./...`
- `go test ./internal/controller/...`
- `golangci-lint run ./internal/controller/...`
- Added `TestClearNodeFailure` and `TestRecordThenClearNodeFailure` covering the helper directly.

## Checklist

- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fixed a bug where a node that recovered from an evaluation error remained listed in a NodeReadinessRule's status.failedNodes indefinitely.
```

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -159,6 +159,8 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
 			errs = append(errs, err)
 			metrics.Failures.WithLabelValues(rule.Name, "EvaluationError").Inc()
+		} else {
+			r.clearNodeFailure(rule, node.Name)
 		}
 
 		// Persist the rule status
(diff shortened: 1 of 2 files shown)


bhuvan-somisetty:
Thanks for the ok-to-test, @ajaysundark! All the Prow jobs (test, e2e, verify-all) are green now. Whenever you get a chance to take a look, let me know if you'd like anything changed.

Priyankasaggu11929:
> /cc @vitorfloriano @Priyankasaggu11929 do either of you have time for this scenario review?

I'll take a look at it, and post some comments by tomorrow EOD.

Priyankasaggu11929:
or wait, I see PR https://github.com/kubernetes-sigs/node-readiness-controller/pull/209 landed first and 


@bhuvan-somisetty - you have already dropped some review comments around it here https://github.com/kubernetes-sigs/node-readiness-controller/pull/332#issuecomment-5150514376

IIUC (from just PR titles and from the code changes as well), both these PRs are supposed to be fixing the same problem.

Let's try to get #209 merged.

bhuvan-somisetty:
Makes sense, #209 already covers this exact fix along with the data race one. Closing this in favor of that. Thanks for checking!

ajaysundark:
Thanks for your PR. Let's track the fix in the current open PR #381

PR #334, opened 2026-08-01
cleanup(reporter): address review follow-ups on unchanged-condition skip

LightCreator1007:
## Description
Follow-up to #263, addressing the review comments left open when it merged.

- Fold the idempotency check into a single early return after the condition search, dropping the `needsUpdate` flag and the separate `transitionTime` variable, the split logic was hard to follow.
- Replace the block comment with line comments (block comments are a Go anti-pattern), and refer to `heartbeatPeriod` rather than a hardcoded "5 minutes", since the period is configurable.
- Rename test case `State change triggers immediate write` to `update condition on state change` — the immediate write is the outcome, not the cause, and is now noted in a comment.
- Return a bool instead of an update count from the test helper, collapsing the assertion to one comparison.

No behaviour change.

## Related Issue
None

## Type of Change

/kind cleanup

## Testing
This change is a refactor only.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
NONE
```

### DIFF
--- a/cmd/readiness-condition-reporter/main.go
+++ b/cmd/readiness-condition-reporter/main.go
@@ -25,6 +25,7 @@ import (
 	"net/url"
 	"os"
 	"os/signal"
+	"slices"
 	"syscall"
 	"time"
 
@@ -303,51 +304,46 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
 			status = corev1.ConditionTrue
 		}
 
-		// Find the condition we report on, carrying its transition time over for
-		// as long as the status is unchanged.
-		var transitionTime metav1.Time
-		idx := -1
-		for i, condition := range node.Status.Conditions {
-			if string(condition.Type) == conditionType {
-				idx = i
-				if condition.Status == status {
-					transitionTime = condition.LastTransitionTime
+		// Find the condition we report on, if it already exists.
+		idx := slices.IndexFunc(node.Status.Conditions, func(c corev1.NodeCondition) bool {
+			return string(c.Type) == conditionType
+		})
 
-					// Nothing has changed about the condition, so skip the update
-					// to avoid unnecessary etcd write amplification.
-					// Write still forced once LastHeartbeatTime is stale,
-					// this is to signal that the reporter is alive
-					if condition.Reason == health.Reason &&
-						condition.Message == health.Message &&
-						time.Since(condition.LastHeartbeatTime.Time) < heartbeatPeriod {
-						klog.V(4).InfoS("Condition state unchanged, skipping node status update", "node", nodeName, "condition", conditionType)
-						reporterConditionWritesTotal.WithLabelValues("skipped").Inc()
-						return nil
-					}
-				}
-				break
-			}
-		}
-
-		if transitionTime.IsZero() {
-			transitionTime = now
-		}
-
-		// Create condition
-		condition := corev1.NodeCondition{
+		// Create a new condition
+		newCondition := corev1.NodeCondition{
 			Type:               corev1.NodeConditionType(conditionType),
 			Status:             status,
 			LastHeartbeatTime:  now,
-			LastTransitionTime: transitionTime,
+			LastTransitionTime: now,
 			Reason:             health.Reason,
 			Message:            health.Message,
 		}
 
 		// Update node status
 		if idx >= 0 {
-			node.Status.Conditions[idx] = condition
+			existingCondition := node.Status.Conditions[idx]
+
+			// Nothing has changed about the condition and heartbeat is fresh
+			// skip write to avoid unnecessary etcd churn, force write if heartbeat is stale
+			if existingCondition.Status == status &&
+				existingCondition.Reason == health.Reason &&
+				existingCondition.Message == health.Message &&
+				time.Since(existingCondition.LastHeartbeatTime.Time) < heartbeatPeriod {
+				klog.V(4).InfoS("Condition state unchanged, skipping node status update", "node", nodeName, "condition", conditionType)
+				reporterConditionWritesTotal.WithLabelValues("skipped").Inc()
+				return nil
+			}
+
+			// Status is unchanged (even though something else, like Message, updated) —
+			// preserve the original transition time.
+			// A zero value means no transition was recorded, fallback to now
+			if existingCondition.Status == status && !existingCondition.LastTransitionTime.IsZero() {
+				newCondition.LastTransitionTime = existingCondition.LastTransitionTime
+			}
+
+			node.Status.Conditions[idx] = newCondition
 		} else {
-			node.Status.Conditions = append(node.Status.Conditions, condition)
+			node.Status.Conditions = append(node.Status.Conditions, newCondition)
 		}
 
 		_, err = client.CoreV1().Nodes().UpdateStatus(ctx, node, metav1.UpdateOptions{})
--- a/cmd/readiness-condition-reporter/main.go
+++ b/cmd/readiness-condition-reporter/main.go
@@ -309,7 +309,6 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
 			return string(c.Type) == conditionType
 		})
 
-		// Create a new condition
 		newCondition := corev1.NodeCondition{
 			Type:               corev1.NodeConditionType(conditionType),
 			Status:             status,
@@ -319,7 +318,6 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
 			Message:            health.Message,
 		}
 
-		// Update node status
 		if idx >= 0 {
 			existingCondition := node.Status.Conditions[idx]
 
@@ -336,8 +334,7 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
 
 			// Status is unchanged (even though something else, like Message, updated) —
 			// preserve the original transition time.
-			// A zero value means no transition was recorded, fallback to now
-			if existingCondition.Status == status && !existingCondition.LastTransitionTime.IsZero() {
+			if existingCondition.Status == status {
 				newCondition.LastTransitionTime = existingCondition.LastTransitionTime
 			}
 
--- a/cmd/readiness-condition-reporter/main.go
+++ b/cmd/readiness-condition-reporter/main.go
@@ -303,54 +303,44 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
 			status = corev1.ConditionTrue
 		}
 
-		// Find existing condition to preserve transition time if status hasn't changed
-		var transitionTime metav1.Time
+		// Find the existing condition of this type: it drives both the write
+		// skip below and the transition time carried over onto the new condition.
 		var existingCondition *corev1.NodeCondition
-
 		for _, condition := range node.Status.Conditions {
 			if string(condition.Type) == conditionType {
 				condCopy := condition
 				existingCondition = &condCopy
-				if condition.Status == status {
-					transitionTime = condition.LastTransitionTime
-				}
 				break
 			}
 		}
 
 		// If the semantic state is completely unchanged, bypass the API write
 		// to prevent etcd write amplification and control plane flooding.
-		needsUpdate := true
-		if existingCondition != nil && existingCondition.Status == status && existingCondition.Reason == health.Reason && existingCondition.Message == health.Message {
-			needsUpdate = false
-			/*
-				NOTE: Skipping the write stops refreshing the LastHeartbeatTime on every tick.
-				To mitigate this, force an update every 5 minutes even if the state is unchanged.
-			*/
-			if time.Since(existingCondition.LastHeartbeatTime.Time) >= heartbeatPeriod {
-				needsUpdate = true
-			}
-		}
-
-		if !needsUpdate {
-			// state has not changed for specified period, skip the write
+		// Skipping the write also stops refreshing LastHeartbeatTime, which is the
+		// only signal that the reporter is still alive, so a write is still forced
+		// once the condition has gone untouched for heartbeatPeriod.
+		if existingCondition != nil &&
+			existingCondition.Status == status &&
+			existingCondition.Reason == health.Reason &&
+			existingCondition.Message == health.Message &&
+			time.Since(existingCondition.LastHeartbeatTime.Time) < heartbeatPeriod {
 			klog.V(4).InfoS("Condition state unchanged, skipping node status update", "node", nodeName, "condition", conditionType)
 			reporterConditionWritesTotal.WithLabelValues("skipped").Inc()
 			return nil
 		}
 
-		if transitionTime.IsZero() {
-			transitionTime = now
-		}
-
-		// Create condition
+		// Create condition, preserving the transition time if the status itself
+		// hasn't changed; a status flip starts a new transition.
 		condition := corev1.NodeCondition{
 			Type:               corev1.NodeConditionType(conditionType),
 			Status:             status,
 			LastHeartbeatTime:  now,
-			LastTransitionTime: transitionTime,
+			LastTransitionTime: now,
 			Reason:             health.Reason,
 			Message:            health.Message,
+		}
+		if existingCondition != nil && existingCondition.Status == status && !existingCondition.LastTransitionTime.IsZero() {
+			condition.LastTransitionTime = existingCondition.LastTransitionTime
 		}
 
 		// Update node status
--- a/cmd/readiness-condition-reporter/main.go
+++ b/cmd/readiness-condition-reporter/main.go
@@ -303,57 +303,50 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
 			status = corev1.ConditionTrue
 		}
 
-		// Find the existing condition of this type: it drives both the write
-		// skip below and the transition time carried over onto the new condition.
-		var existingCondition *corev1.NodeCondition
-		for _, condition := range node.Status.Conditions {
+		// Find the condition we report on, carrying its transition time over for
+		// as long as the status is unchanged.
+		var transitionTime metav1.Time
+		idx := -1
+		for i, condition := range node.Status.Conditions {
 			if string(condition.Type) == conditionType {
-				condCopy := condition
-				existingCondition = &condCopy
+				idx = i
+				if condition.Status == status {
+					transitionTime = condition.LastTransitionTime
+
+					// Nothing has changed about the condition, so skip the update
+					// to avoid unnecessary etcd write amplification.
+					// Write still forced once LastHeartbeatTime is stale,
+					// this is to signal that the reporter is alive
+					if condition.Reason == health.Reason &&
+						condition.Message == health.Message &&
+						time.Since(condition.LastHeartbeatTime.Time) < heartbeatPeriod {
+						klog.V(4).InfoS("Condition state unchanged, skipping node status update", "node", nodeName, "condition", conditionType)
+						reporterConditionWritesTotal.WithLabelValues("skipped").Inc()
+						return nil
+					}
+				}
 				break
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


ajaysundark:
Thanks for iterating on this!

/cc @AnuragThePathak as he had the original comments

cmd/readiness-condition-reporter/main.go
```diff
@@ -255,54 +255,44 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
 			status = corev1.ConditionTrue
 		}
 
-		// Find existing condition to preserve transition time if status hasn't changed
-		var transitionTime metav1.Time
+		// Find the existing condition of this type: it drives both the write
+		// skip below and the transition time carried over onto the new condition.
```
AnuragThePathak: I am genuinely finding this comment hard to comprehend.

cmd/readiness-condition-reporter/main.go
```diff
@@ -255,54 +255,44 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
...
 		var existingCondition *corev1.NodeCondition
-
 		for _, condition := range node.Status.Conditions {
 			if string(condition.Type) == conditionType {
 				condCopy := condition
 				existingCondition = &condCopy
-				if condition.Status == status {
```
AnuragThePathak, line 314: Actually what I was suggesting was not to remove this part.

I suggested this part to merge here

```go
if existingCondition != nil &&
			existingCondition.Status == status &&
			existingCondition.Reason == health.Reason &&
			existingCondition.Message == health.Message &&
			time.Since(existingCondition.LastHeartbeatTime.Time) < heartbeatPeriod {
			klog.V(4).InfoS("Condition state unchanged, skipping node status update", "node", nodeName, "condition", conditionType)
			return nil
}
```

If we merge it here, we will already have existingCondition != nil &&
			existingCondition.Status == status, true

we just need to check for remaining 3

existingCondition local variable is not being leveraged elsewhere either way.

AnuragThePathak: My idea looks something like this for reference @LightCreator1007 

```go
// Find the existing condition of this type: it drives both the write
		// skip below and the transition time carried over onto the new condition.
		var transitionTime metav1.Time
		for _, condition := range node.Status.Conditions {
			if string(condition.Type) == conditionType {
				if condition.Status == status {
					transitionTime = condition.LastTransitionTime

					if condition.Reason == health.Reason &&
						condition.Message == health.Message &&
						time.Since(condition.LastHeartbeatTime.Time) < heartbeatPeriod {
						klog.V(4).InfoS(
							"Condition state unchanged, skipping node status update",
							 "node", nodeName, "condition", conditionType)
						return nil
					}
				}
				break
			}
		}

		if transitionTime.IsZero() {
			transitionTime = now
		}

		// Create condition, preserving the transition time if the status itself
		// hasn't changed; a status flip starts a new transition.
		condition := corev1.NodeCondition{
			Type:               corev1.NodeConditionType(conditionType),
			Status:             status,
			LastHeartbeatTime:  now,
			LastTransitionTime: transitionTime,
			Reason:             health.Reason,
			Message:            health.Message,
		}
```

Feel free to evaluate and improvise on top of my idea as well or bring something more innovative.

AnuragThePathak (commented):
Thanks for following up @LightCreator1007 glad you decided not to wrap up the chapter with merging of the original PR.

cmd/readiness-condition-reporter/main.go
LightCreator1007: I am sorry for that, it was mostly affected due to the refactors and oversight post refactor.

cmd/readiness-condition-reporter/main.go
LightCreator1007: I had misunderstood the intention here, merged the check as you specified in your example, and this seems to be the best way to me too. 
Only thing I have changed/added is that I now capture the index in the same loop so the write back can reuse it instead of scanning for conditions the second time. Please take a look at that part and let me know if that is okay.

cmd/readiness-condition-reporter/main.go
```diff
@@ -305,16 +295,9 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
...
-		found := false
-		for i, c := range node.Status.Conditions {
-			if string(c.Type) == conditionType {
-				node.Status.Conditions[i] = condition
-				found = true
-				break
-			}
-		}
-
-		if !found {
```
AnuragThePathak, line 366: I must admit, this is something I never had my eyes on, and didn't anticipate this change either until I saw this today (the major reason being it was not part of what you changed earlier and existed way before that). This slightly makes me question whether your former approach to use existingCondition was actually better, but considering that involves pointer, and considering simplicity of our current needs I'm approving this.

AnuragThePathak (approved):

cmd/readiness-condition-reporter/main.go
```diff
@@ -255,41 +255,31 @@ func updateNodeCondition(ctx context.Context, client kubernetes.Interface, nodeN
...
-		// Find existing condition to preserve transition time if status hasn't changed
+		// Find the condition we report on, carrying its transition time over for
+		// as long as the status is unchanged.
 		var transitionTime metav1.Time
-		var existingCondition *corev1.NodeCondition
-
-		for _, condition := range node.Status.Conditions {
+		idx := -1
+		for i, condition := range node.Status.Conditions {
 			if string(condition.Type) == conditionType {
```
AnuragThePathak: @LightCreator1007 I was thinking about the level of nesting we are introducing here, and I felt like replacing the if followed by branching with if not true followed by a continue will get rid of one level of nesting and make the things easier to read than the state we are in today.

If any further refactoring needed in future, we should do it separately, for now this minor change should be good enough.

AnuragThePathak: @LightCreator1007 a little change in plan. Let's standardize the process in one go as we are already touching beyond the scope of what your original PR was.

I would like to split the process into two parts.
1. Searching the existing condition
2. The aftermath

Part 2 would have two branches based on conditionFound is true or false.

Considered the idea of splitting the part 2 in two parts performing necessary processing like the early exit and mutation separately but in current situation that level of granular splitting is not as beneficial. Hence I would recommend a structure like this

```go
idx := slices.IndexFunc(node.Status.Conditions, func(c corev1.NodeCondition) bool {
    return string(c.Type) == conditionType
})

newCondition := corev1.NodeCondition{
    Type:               corev1.NodeConditionType(conditionType),
    Status:             status,
    LastHeartbeatTime:  now,
    LastTransitionTime: now,
    Reason:             health.Reason,
    Message:            health.Message,
}

if idx >= 0 {
    existingCondition := node.Status.Conditions[idx]

    // Nothing changed since the last report and the heartbeat is still fresh —
    // skip the write to avoid unnecessary etcd churn.
    if existingCondition.Status == status &&
        existingCondition.Reason == health.Reason &&
        existingCondition.Message == health.Message &&
        time.Since(existingCondition.LastHeartbeatTime.Time) < heartbeatPeriod {
        klog.V(4).InfoS("Condition state unchanged, skipping node status update", "node", nodeName, "condition", conditionType)
        return nil
    }

    // Status is unchanged (even though something else, like Message, updated) —
    // preserve the original transition time.
    if existingCondition.Status == status {
        newCondition.LastTransitionTime = existingCondition.LastTransitionTime
    }

    node.Status.Conditions[idx] = newCondition
} else {
    node.Status.Conditions = append(node.Status.Conditions, newCondition)
}
```

Treat it like a prototype even though maybe ready to use.

AnuragThePathak (changes requested):

cmd/readiness-condition-reporter/main.go
AnuragThePathak: I believe this approach is the easiest to understand for someone who is seeing this portion for the first time.

LightCreator1007: Hello,
I implemented the structure you proposed almost exactly, but the only thing I added was an `IsZero()`  check to the `existingCondition`'s `LastTransitionTime`.
I did this because without the guard, an existing condition with no recorded transition time would have `null` copied forward on every unchanged-status write. It promotes it to `now` in the current iteration of the code.

Let me know if this works.

AnuragThePathak: @LightCreator1007 I would disagree with this approach. It's our node condition reporter solely in charge of owning the lifecycle of the condition, and it never leaks a zero LastTransitionTime unless there is a bug. I would say let's avoid adding defensive checks unless it's actually needed or we are not certain of something.

LightCreator1007: I have removed the check.

AnuragThePathak: You can mark this comment as resolved, while I perform one final review. I hope you also it's ready to merge (like tests are covered properly, all tests passing, ci etc?

LightCreator1007: Yes to the best of my knowledge all tests are passing and it is ready to merge.

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -218,27 +220,23 @@ func TestUpdateNodeCondition(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			client := fake.NewSimpleClientset(tt.existingNode)
 
-			countUpdates := func() int {
-				n := 0
+			updateCalled := func() bool {
```
AnuragThePathak: About this, I gave some thought about this, I think your original approach of counting was better. So please revert this one.

But change the wantUpdateCalled to integer value (update name of that variable too as wantUpdateCalled sounds boolean to anyone).

AnuragThePathak (commented):
I have two more suggestions, I see that we are not testing two situations now.

1. Smoke test for node not found error path
2. Validation of the LastTransitionTime

Since we spent so much time refactoring, I think working on these two and sealing these two will elevate the standard of this method significantly.

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -262,6 +278,39 @@ func TestUpdateNodeCondition(t *testing.T) {
...
+					t.Errorf("LastTransitionTime = %v, want it preserved as %v", foundCondition.LastTransitionTime, previousTransition)
+				}
+			} else if foundCondition.LastTransitionTime.Before(&testStart) {
+				t.Errorf("LastTransitionTime = %v, want a new transition at or after %v", foundCondition.LastTransitionTime, testStart)
+			}
 		})
 	}
 }
+
+func TestUpdateNodeConditionNodeNotFound(t *testing.T) {
```
AnuragThePathak: You didn't have to write an extra test for this one. Simply could have passed nil in existingNode in one of the test cases above.

In fakeClientSet now you will get error if try to pass nil. So convert the existingNode to an array format, push to the array if not nil. That will work as FakeClientSet supports dynamic input.

Then after that in case of Not Found error return the method successfully, else fatal error.

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -209,34 +214,45 @@ func TestUpdateNodeCondition(t *testing.T) {
...
+			wantTransitionPreserved: true,
 		},
 	}
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			client := fake.NewSimpleClientset(tt.existingNode)
 
-			updateCalled := func() bool {
+			countUpdates := func() int {
```
AnuragThePathak: Rename this method to countUpdateCalls, and move outside t.Run

just below staleTime declaration at line 113 looks like a great place to me.

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -209,34 +214,45 @@ func TestUpdateNodeCondition(t *testing.T) {
...
 			}
 
+			var previousTransition metav1.Time
+			for _, cond := range tt.existingNode.Status.Conditions {
+				if string(cond.Type) == conditionType {
+					previousTransition = cond.LastTransitionTime
+					break
+				}
+			}
+			testStart := metav1.NewTime(time.Now())
```
AnuragThePathak: I think using 

If previousTransistion.isZero() {
previousTransistion = now
}

be more convenient here.

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -262,6 +278,39 @@ func TestUpdateNodeCondition(t *testing.T) {
 			if foundCondition.Reason != tt.wantReason {
 				t.Errorf("Condition reason = %v, want %v", foundCondition.Reason, tt.wantReason)
 			}
+
+			if tt.wantTransitionPreserved {
+				if !foundCondition.LastTransitionTime.Equal(&previousTransition) {
+					t.Errorf("LastTransitionTime = %v, want it preserved as %v", foundCondition.LastTransitionTime, previousTransition)
+				}
+			} else if foundCondition.LastTransitionTime.Before(&testStart) {
```
AnuragThePathak: About this one please use !after here. That feels natural to comprehend for a human in my perspective, as it's error catching scenario. Welcome for any critics on my suggestion though.

p.s. actually using !after will cover for != test start as well which is an unintentional perk.

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -112,13 +113,14 @@ func TestUpdateNodeCondition(t *testing.T) {
...
-		wantReason       string
-		wantUpdateCalled bool
+		name                    string
+		existingNode            *corev1.Node
+		health                  *HealthResponse
+		heartbeatPeriod         time.Duration
+		wantStatus              corev1.ConditionStatus
+		wantReason              string
+		wantUpdateCount         int
+		wantTransitionPreserved bool
```
AnuragThePathak: wantTransitionTimePreserved seems like right choice of name to me even though slightly longer. This one also works for now. Go for whatever feels natural to you as well.

AnuragThePathak (changes requested):

DsThakurRawat:
Heads up, #367 merged into `main` about an hour ago and it lands in this same file, so the branch doesn't merge cleanly any more. Checked it against `main` at `d74dabd`:

```
Auto-merging cmd/readiness-condition-reporter/main.go
CONFLICT (content): Merge conflict in cmd/readiness-condition-reporter/main_test.go
```

`main.go` is fine. The only conflict is in `main_test.go` and it's the mechanical kind: both branches appended a new top level test function at the end of the file, so git can't tell which one goes first. `main` has `TestParseDurationWithDefault` now, this branch has `TestUpdateNodeConditionNodeNotFound`.

Keeping both, in either order, clears it. I resolved it that way locally and `gofmt`, `go build ./...` and `go test ./cmd/readiness-condition-reporter/...` all came back clean, so there's nothing behavioural in it.

It might also sort itself out. The review note above suggests folding that case into the existing table rather than keeping a separate test, and if `TestUpdateNodeConditionNodeNotFound` goes away then so does the conflict.

Sorry about the collision, that was my PR that landed.

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -207,38 +225,62 @@ func TestUpdateNodeCondition(t *testing.T) {
...
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			client := fake.NewSimpleClientset(tt.existingNode)
-
-			countUpdates := func() int {
-				n := 0
-				for _, a := range client.Actions() {
-					if a.GetVerb() == "update" && a.GetSubresource() == "status" && a.GetResource().Resource == "nodes" {
-						n++
+			var objs []runtime.Object
```
AnuragThePathak:
```go
            var client *fake.Clientset
			if tt.existingNode != nil {
				client = fake.NewClientset(tt.existingNode)
			} else {
				client = fake.NewClientset()
			}


            var previousTransition metav1.Time
			if tt.existingNode != nil {
				for _, cond := range tt.existingNode.Status.Conditions {
					if string(cond.Type) == conditionType {
						previousTransition = cond.LastTransitionTime
						break
					}
				}
			}
			if previousTransition.IsZero() {
				previousTransition = metav1.NewTime(time.Now())
			}
```

Decouple client and previousTransition.

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -207,38 +225,62 @@ func TestUpdateNodeCondition(t *testing.T) {
...
+			}
+			client := fake.NewSimpleClientset(objs...)
+
+			if previousTransition.IsZero() {
+				previousTransition = metav1.NewTime(time.Now())
 			}
 
 			err := updateNodeCondition(context.Background(), client, nodeName, conditionType, tt.health, tt.heartbeatPeriod)
+			if tt.existingNode == nil {
+				if !apierrors.IsNotFound(err) {
```
AnuragThePathak: This log/condition won't tell exact failure scenario, as it doesn't differentiate between no error and error being something other than not found.

```go
             if err != nil {
				if tt.wantNotFoundErr {
					if !apierrors.IsNotFound(err) {
						t.Fatalf("updateNodeCondition() error = %v, want a NotFound error", err)
					}
					if got := countUpdateCalls(client); got != tt.wantUpdateCount {
						t.Errorf("UpdateStatus called = %v, want %v", got, tt.wantUpdateCount)
					}
					return
				}
				t.Fatalf("updateNodeCondition() error = %v", err)
			}
			if tt.wantNotFoundErr {
				t.Fatalf("updateNodeCondition() succeeded, want a NotFound error")
			}
```

Also introduced wantNotFoundErr because inputs in tests are not often expected to remain the same during the runtime most often.

cmd/readiness-condition-reporter/main_test.go
AnuragThePathak: @LightCreator1007 update this please

AnuragThePathak (changes requested):
Remaining looks good other than the addressed 3

cmd/readiness-condition-reporter/main_test.go
AnuragThePathak: @LightCreator1007 lgtm but this is still pending

AnuragThePathak (approved):
@ajaysundark this one looks good to me. Feel free to merge as per release and other convince as no functional change

cmd/readiness-condition-reporter/main_test.go
LightCreator1007: done.

cmd/readiness-condition-reporter/main_test.go
```diff
@@ -207,38 +225,68 @@ func TestUpdateNodeCondition(t *testing.T) {
...
+				for _, cond := range tt.existingNode.Status.Conditions {
+					if string(cond.Type) == conditionType {
+						previousTransition = cond.LastTransitionTime
+						break
 					}
 				}
-				return n
+			}
+			if previousTransition.IsZero() {
+				previousTransition = metav1.NewTime(time.Now())
```
ajaysundark: nit: it'd be good to base previousTransition to a fixed / deterministic timestamp instead of injecting 'now' as you're using it to comparing later.

ajaysundark (approved):
@LightCreator1007  Thanks for this nice refactor!

added a nit comment (non blocking) if you could find time to update, please consider handling this.

/hold
/lgtm
/approve

ajaysundark (approved):
/lgtm

ajaysundark:
Thanks @LightCreator1007 @AnuragThePathak !

AnuragThePathak (approved):
LGTM with the update on deterministic previousTransistion. We are not exactly testing the equality of time stamp, time.Now() put by the updateNodeCondition, but that should be fine I believe in current situation I think.

LightCreator1007:
hello again :) 
If i am not wrong ig the cleanup is mostly complete, but the PR does not merge due to a hold tag...
it might need to be updated, lest further refactors are required.

ajaysundark:
@LightCreator1007 Thanks for the callout, Sorry I missed the hold. This is ready to merge. 

/unhold

PR #343, opened 2026-08-03
fix: use optimistic locking on rule status patches

### LINKED ISSUE #341: [BUG] RuleReconciler's updateRuleStatus can silently discard concurrent NodeReconciler status writes
### What happened?

`updateRuleStatus` (`internal/controller/nodereadinessrule_controller.go:561-593`) overwrites `rule.Status.NodeEvaluations`/`AppliedNodes`/`FailedNodes` wholesale from whatever `RuleReconciler` computed at the start of its node sweep. It patches with a plain `client.MergeFrom` (no `MergeFromWithOptimisticLock`), wrapped in `retry.RetryOnConflict`. Since the patch carries no resourceVersion precondition, the API server never returns a conflict, so the retry wrapper never actually retries.

Net effect: if `NodeReconciler` (a separate controller/queue) writes a per-node status update for a node while `RuleReconciler`'s sweep is still running, `updateRuleStatus`'s later overwrite silently erases that update. No error, no log, no retry — it just disappears.

This is the same bug class `#180` fixed for `node.spec.taints` (see the `MergeFromWithOptimisticLock` comment on `addTaintBySpec`/`removeTaintBySpec`), just never applied to the rule-status patches.

### Reproduction (verified against `main` @ `43bff46`)

Drop this into `internal/controller/` as a plain Go test and run `go test ./internal/controller/... -run TestIssue341LostUpdateRepro -v`:

```go
package controller

import (
	"context"
	"testing"

	corev1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/types"
	"k8s.io/client-go/kubernetes/fake"
	"k8s.io/client-go/tools/events"
	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"

	readinessv1alpha1 "sigs.k8s.io/node-readiness-controller/api/v1alpha1"
)

func TestIssue341LostUpdateRepro(t *testing.T) {
	ctx := context.Background()
	scheme := runtime.NewScheme()
	_ = corev1.AddToScheme(scheme)
	_ = readinessv1alpha1.AddToScheme(scheme)

	nodeA := &corev1.Node{
		ObjectMeta: metav1.ObjectMeta{Name: "node-a", Labels: map[string]string{"role": "worker"}},
		Status:     corev1.NodeStatus{Conditions: []corev1.NodeCondition{{Type: "Ready", Status: corev1.ConditionTrue}}},
	}
	nodeB := &corev1.Node{
		ObjectMeta: metav1.ObjectMeta{Name: "node-b", Labels: map[string]string{"role": "worker"}},
		Status:     corev1.NodeStatus{Conditions: []corev1.NodeCondition{{Type: "Ready", Status: corev1.ConditionTrue}}},
	}
	rule := &readinessv1alpha1.NodeReadinessRule{
		ObjectMeta: metav1.ObjectMeta{Name: "repro-rule"},
		Spec: readinessv1alpha1.NodeReadinessRuleSpec{
			Conditions:      []readinessv1alpha1.ConditionRequirement{{Type: "Ready", RequiredStatus: corev1.ConditionTrue}},
			Taint:           corev1.Taint{Key: "readiness.k8s.io/repro", Effect: corev1.TaintEffectNoSchedule},
			NodeSelector:    metav1.LabelSelector{MatchLabels: map[string]string{"role": "worker"}},
			EnforcementMode: readinessv1alpha1.EnforcementModeContinuous,
		},
	}

	fc := fakeclient.NewClientBuilder().WithScheme(scheme).WithObjects(nodeA, nodeB, rule).WithStatusSubresource(rule).Build()
	c := &RuleReadinessController{
		Client: fc, Scheme: scheme, clientset: fake.NewSimpleClientset(),
		ruleCache: map[string]*readinessv1alpha1.NodeReadinessRule{rule.Name: rule}, EventRecorder: events.NewFakeRecorder(10),
	}

	// RuleReconciler starts a sweep from a nodeList snapshot that only knows about node-a.
	if err := c.processAllNodesForRule(ctx, rule, &corev1.NodeList{Items: []corev1.Node{*nodeA}}); err != nil {
		t.Fatalf("processAllNodesForRule: %v", err)
	}
	// While that sweep is "in flight", NodeReconciler independently reconciles node-b.
	if err := c.processNodeAgainstAllRules(ctx, nodeB); err != nil {
		t.Fatalf("processNodeAgainstAllRules: %v", err)
	}
	// RuleReconciler now finishes its (stale) sweep and patches the rule status.
	if err := c.updateRuleStatus(ctx, rule); err != nil {
		t.Fatalf("updateRuleStatus: %v", err)
	}

	final := &readinessv1alpha1.NodeReadinessRule{}
	_ = fc.Get(ctx, types.NamespacedName{Name: rule.Name}, final)
	for _, e := range final.Status.NodeEvaluations {
		if e.NodeName == "node-b" {
			return // node-b survived, bug not present
		}
	}
	t.Fatalf("node-b's evaluation was silently discarded. Final NodeEvaluations: %+v", final.Status.NodeEvaluations)
}
```

On `main` this fails with:

```
node-b's evaluation was silently discarded. Final NodeEvaluations: [{NodeName:node-a ...}]
```

`node-b` should be there — `NodeReconciler` wrote it — but `RuleReconciler.updateRuleStatus`'s wholesale overwrite erased it, silently, with no error anywhere.

### Manual reproduction on a real cluster

1. Create a `NodeReadinessRule` matching several nodes.
2. Trigger a `RuleReconciler` reconcile for the rule (e.g. touch an annotation) with enough nodes that the sweep takes a noticeable amount of time.
3. While that's running, flip a readiness condition on one of the rule's nodes so `NodeReconciler` reconciles it and writes its own status update for that node.
4. Once both finish, check `rule.status.nodeEvaluations` for that node — the entry written in step 3 is gone, reverted to the older value `RuleReconciler`'s sweep had computed, with no error logged.

### Impact

`status.nodeEvaluations`/`appliedNodes`/`failedNodes` can silently lose or revert entries whenever `RuleReconciler` and `NodeReconciler` touch the same rule concurrently — which is normal steady-state operation, not an edge case. Anything reading rule status (the reporter, dashboards, alerting) can see incorrect data with nothing in the logs to explain it.

### Fix

Same fix as `#180`: use `client.MergeFromWithOptimisticLock` on the rule-status patch, and merge per-node changes by node name instead of overwriting the whole slice, so `RuleReconciler`'s sweep only ever replaces the entries it actually recomputed.

bhuvan-somisetty:
## Description

`RuleReconciler` and `NodeReconciler` both patch `NodeReadinessRule.Status` concurrently, but every status/finalizer patch used a plain `client.MergeFrom` with no resourceVersion precondition, wrapped in `retry.RetryOnConflict`. A JSON merge patch only carries that precondition when `MergeFromWithOptimisticLock` is used, so without it the API server never returns a conflict and the retry wrapper never actually retries. This is the same bug `#180` fixed for node taint patches (`addTaintBySpec`/`removeTaintBySpec`), just left open on the rule-status side.

The worst instance was `updateRuleStatus`: it replaced `NodeEvaluations`/`FailedNodes` wholesale from a snapshot computed at the start of a `RuleReconciler` sweep, so it could silently discard a concurrent `NodeReconciler` per-node update for a node outside that sweep's snapshot. Fixed by having `processAllNodesForRule` return a delta of exactly the per-node changes it made, and merging that delta by node name instead of overwriting the whole slice.

Also added the missing optimistic lock to `ensureFinalizer`, the finalizer removal in `reconcileDelete`, `cleanupDeletedNodes`, and `markBootstrapCompleted`'s node annotation patch, matching the pattern already used by `addTaintBySpec`/`removeTaintBySpec`.

## Related

Fixes #341

## Type of Change

/kind bug

## Testing

- `go build ./...`
- `go vet ./...`
- `go test ./internal/controller/...` (63/63 specs pass; the only failure locally is envtest's Windows-only teardown limitation, unrelated to this change)
- Added two regression tests: one proving a `NodeReconciler`-written evaluation for a node outside the `RuleReconciler` sweep survives `updateRuleStatus`, and one proving `updateRuleStatus` actually retries (and doesn't lose data) on a genuine conflict.

## Checklist

- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fixed a bug where concurrent status writes from RuleReconciler and NodeReconciler to the same NodeReadinessRule could silently overwrite each other, since the retry-on-conflict wrapper around these patches never actually detected a conflict.
```

### DIFF
--- a/internal/controller/helper.go
+++ b/internal/controller/helper.go
@@ -136,14 +136,12 @@ func labelsEqual(a, b map[string]string) bool {
 	return maps.Equal(a, b)
 }
 
-// nodeStatusDelta captures the per-node NodeEvaluation/NodeFailure changes a single
-// processAllNodesForRule sweep actually produced, keyed by node name. It intentionally does not
-// carry AppliedNodes/ObservedGeneration/DryRunResults: those fields only ever have one writer
-// (RuleReconciler), so a plain overwrite of them is safe.
+// nodeStatusDelta captures the per-node NodeEvaluation/NodeFailure changes produced by a single
+// processAllNodesForRule sweep, keyed by node name. It excludes AppliedNodes, ObservedGeneration,
+// and DryRunResults, which have a single writer and are safe to overwrite directly.
 //
-// A nil value in failures means "clear any failure recorded for this node" (the node evaluated
-// successfully). evaluations only ever holds entries for nodes that were freshly (re-)evaluated
-// this sweep; a failed evaluation leaves the node's prior NodeEvaluation untouched.
+// A nil value in failures clears any failure recorded for that node. evaluations only holds
+// entries for nodes freshly (re-)evaluated this sweep.
 type nodeStatusDelta struct {
 	evaluations map[string]readinessv1alpha1.NodeEvaluation
 	failures    map[string]*readinessv1alpha1.NodeFailure
@@ -151,12 +149,6 @@ type nodeStatusDelta struct {
 
 // applyNodeStatusDelta merges delta into rule's NodeEvaluations/FailedNodes, replacing only the
 // entries for nodes present in delta and leaving every other node's entry untouched.
-//
-// This is the crux of fixing the lost-update bug described in #341: a naive full-slice
-// replacement of NodeEvaluations/FailedNodes (computed from a nodeList snapshot taken at the
-// start of a RuleReconciler sweep) would silently discard any per-node status update written
-// concurrently by NodeReconciler for a node this particular sweep didn't touch. Merging by node
-// name instead means each writer only ever overwrites the entries it just recomputed.
 func applyNodeStatusDelta(rule *readinessv1alpha1.NodeReadinessRule, delta nodeStatusDelta) {
 	if len(delta.evaluations) > 0 {
 		merged := make([]readinessv1alpha1.NodeEvaluation, 0, len(rule.Status.NodeEvaluations)+len(delta.evaluations))
--- a/internal/controller/helper.go
+++ b/internal/controller/helper.go
@@ -19,9 +19,11 @@ package controller
 import (
 	"encoding/json"
 	"maps"
+	"sort"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
+
 	readinessv1alpha1 "sigs.k8s.io/node-readiness-controller/api/v1alpha1"
 )
 
@@ -132,4 +134,57 @@ func filterStatusForExistingNodes(
 // labelsEqual checks if two label maps are equal.
 func labelsEqual(a, b map[string]string) bool {
 	return maps.Equal(a, b)
+}
+
+// nodeStatusDelta captures the per-node NodeEvaluation/NodeFailure changes a single
+// processAllNodesForRule sweep actually produced, keyed by node name. It intentionally does not
+// carry AppliedNodes/ObservedGeneration/DryRunResults: those fields only ever have one writer
+// (RuleReconciler), so a plain overwrite of them is safe.
+//
+// A nil value in failures means "clear any failure recorded for this node" (the node evaluated
+// successfully). evaluations only ever holds entries for nodes that were freshly (re-)evaluated
+// this sweep; a failed evaluation leaves the node's prior NodeEvaluation untouched.
+type nodeStatusDelta struct {
+	evaluations map[string]readinessv1alpha1.NodeEvaluation
+	failures    map[string]*readinessv1alpha1.NodeFailure
+}
+
+// applyNodeStatusDelta merges delta into rule's NodeEvaluations/FailedNodes, replacing only the
+// entries for nodes present in delta and leaving every other node's entry untouched.
+//
+// This is the crux of fixing the lost-update bug described in #341: a naive full-slice
+// replacement of NodeEvaluations/FailedNodes (computed from a nodeList snapshot taken at the
+// start of a RuleReconciler sweep) would silently discard any per-node status update written
+// concurrently by NodeReconciler for a node this particular sweep didn't touch. Merging by node
+// name instead means each writer only ever overwrites the entries it just recomputed.
+func applyNodeStatusDelta(rule *readinessv1alpha1.NodeReadinessRule, delta nodeStatusDelta) {
+	if len(delta.evaluations) > 0 {
+		merged := make([]readinessv1alpha1.NodeEvaluation, 0, len(rule.Status.NodeEvaluations)+len(delta.evaluations))
+		for _, eval := range rule.Status.NodeEvaluations {
+			if _, changed := delta.evaluations[eval.NodeName]; !changed {
+				merged = append(merged, eval)
+			}
+		}
+		for _, eval := range delta.evaluations {
+			merged = append(merged, eval)
+		}
+		sort.Slice(merged, func(i, j int) bool { return merged[i].NodeName < merged[j].NodeName })
+		rule.Status.NodeEvaluations = merged
+	}
+
+	if len(delta.failures) > 0 {
+		merged := make([]readinessv1alpha1.NodeFailure, 0, len(rule.Status.FailedNodes)+len(delta.failures))
+		for _, failure := range rule.Status.FailedNodes {
+			if _, changed := delta.failures[failure.NodeName]; !changed {
+				merged = append(merged, failure)
+			}
+		}
+		for _, failure := range delta.failures {
+			if failure != nil {
+				merged = append(merged, *failure)
+			}
+		}
+		sort.Slice(merged, func(i, j int) bool { return merged[i].NodeName < merged[j].NodeName })
+		rule.Status.FailedNodes = merged
+	}
 }
--- a/internal/controller/helper.go
+++ b/internal/controller/helper.go
@@ -147,6 +147,16 @@ type nodeStatusDelta struct {
 	failures    map[string]*readinessv1alpha1.NodeFailure
 }
 
+// sortStatusByNodeName sorts rule.Status.NodeEvaluations and rule.Status.FailedNodes by NodeName.
+func sortStatusByNodeName(rule *readinessv1alpha1.NodeReadinessRule) {
+	sort.Slice(rule.Status.NodeEvaluations, func(i, j int) bool {
+		return rule.Status.NodeEvaluations[i].NodeName < rule.Status.NodeEvaluations[j].NodeName
+	})
+	sort.Slice(rule.Status.FailedNodes, func(i, j int) bool {
+		return rule.Status.FailedNodes[i].NodeName < rule.Status.FailedNodes[j].NodeName
+	})
+}
+
 // applyNodeStatusDelta merges delta into rule's NodeEvaluations/FailedNodes, replacing only the
 // entries for nodes present in delta and leaving every other node's entry untouched.
 func applyNodeStatusDelta(rule *readinessv1alpha1.NodeReadinessRule, delta nodeStatusDelta) {
@@ -160,7 +170,6 @@ func applyNodeStatusDelta(rule *readinessv1alpha1.NodeReadinessRule, delta nodeS
 		for _, eval := range delta.evaluations {
 			merged = append(merged, eval)
 		}
-		sort.Slice(merged, func(i, j int) bool { return merged[i].NodeName < merged[j].NodeName })
 		rule.Status.NodeEvaluations = merged
 	}
 
@@ -176,7 +185,8 @@ func applyNodeStatusDelta(rule *readinessv1alpha1.NodeReadinessRule, delta nodeS
 				merged = append(merged, *failure)
 			}
 		}
-		sort.Slice(merged, func(i, j int) bool { return merged[i].NodeName < merged[j].NodeName })
 		rule.Status.FailedNodes = merged
 	}
+
+	sortStatusByNodeName(rule)
 }
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -169,14 +169,7 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 
 		var successfullyPatchedRule *readinessv1alpha1.NodeReadinessRule
 
-		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			latestRule := &readinessv1alpha1.NodeReadinessRule{}
-			if err := r.Get(ctx, client.ObjectKey{Name: rule.Name}, latestRule); err != nil {
-				return err
-			}
-
-			patch := client.MergeFrom(latestRule.DeepCopy())
-
+		err := r.patchRuleStatusWithOptimisticLock(ctx, rule.Name, func(latestRule *readinessv1alpha1.NodeReadinessRule) bool {
 			// update only this specific node evaluation status
 			currEval := readinessv1alpha1.NodeEvaluation{}
 			for _, eval := range rule.Status.NodeEvaluations {
@@ -215,12 +208,8 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			}
 			latestRule.Status.FailedNodes = updatedFailedNodes
 
-			if err := r.Status().Patch(ctx, latestRule, patch); err != nil {
-				return err
-			}
-
 			successfullyPatchedRule = latestRule
-			return nil
+			return true
 		})
 
 		if err != nil {
@@ -464,7 +453,7 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
 		}
 		deferred = false
 
-		patch := client.MergeFromWithOptions(node.DeepCopy(), client.MergeFromWithOptimisticLock{})
+		stored := node.DeepCopy()
 
 		// Initialize annotations map if nil.
 		if node.Annotations == nil {
@@ -472,7 +461,7 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
 		}
 
 		node.Annotations[annotationKey] = bootstrapAnnotationValue(rule.Name)
-		if err := r.Patch(ctx, node, patch); err != nil {
+		if err := r.Patch(ctx, node, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
 			return err
 		}
 
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -299,9 +298,6 @@ func (r *RuleReadinessController) addTaintBySpec(ctx context.Context, node *core
 		stored := latestNode.DeepCopy()
 		latestNode.Spec.Taints = append(latestNode.Spec.Taints, taintSpec)
 		if err := r.Patch(ctx, latestNode, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
-			if apierrors.IsConflict(err) {
-				metrics.StatusPatchConflicts.WithLabelValues("node_taint", "add").Inc()
-			}
 			return err
 		}
 
@@ -393,9 +389,6 @@ func (r *RuleReadinessController) removeTaint(ctx context.Context, node *corev1.
 			latestNode.Annotations[key] = annotations[key]
 		}
 		if err := r.Patch(ctx, latestNode, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
-			if apierrors.IsConflict(err) {
-				metrics.StatusPatchConflicts.WithLabelValues("node_taint", "remove").Inc()
-			}
 			return err
 		}
 
@@ -474,9 +467,6 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
 
 		node.Annotations[annotationKey] = bootstrapAnnotationValue(rule.Name)
 		if err := r.Patch(ctx, node, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
-			if apierrors.IsConflict(err) {
-				metrics.StatusPatchConflicts.WithLabelValues("node_bootstrap_annotation", "patch").Inc()
-			}
 			return err
 		}
 
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -298,6 +299,9 @@ func (r *RuleReadinessController) addTaintBySpec(ctx context.Context, node *core
 		stored := latestNode.DeepCopy()
 		latestNode.Spec.Taints = append(latestNode.Spec.Taints, taintSpec)
 		if err := r.Patch(ctx, latestNode, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+			if apierrors.IsConflict(err) {
+				metrics.StatusPatchConflicts.WithLabelValues("node_taint", "add").Inc()
+			}
 			return err
 		}
 
@@ -389,6 +393,9 @@ func (r *RuleReadinessController) removeTaint(ctx context.Context, node *corev1.
 			latestNode.Annotations[key] = annotations[key]
 		}
 		if err := r.Patch(ctx, latestNode, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+			if apierrors.IsConflict(err) {
+				metrics.StatusPatchConflicts.WithLabelValues("node_taint", "remove").Inc()
+			}
 			return err
 		}
 
@@ -435,7 +442,12 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
 	deferred := false
 	annotationKey := bootstrapAnnotationKey(rule.GetUID())
 
-	// retry to handle conflict with concurrent node updates
+	// The optimistic lock here isn't guarding the annotation merge itself (that's map-valued
+	// and merges cleanly against concurrent writers, e.g. Kubelet). It guards the
+	// hasTaintBySpec check above: without it, a taint added between that check and the Patch
+	// below would go undetected, and we'd mark bootstrap complete on a node that still carries
+	// the taint. See the "should not mark bootstrap completed when the rule taints concurrently"
+	// test for the regression this prevents.
 	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		node := &corev1.Node{}
 		if err := r.Get(ctx, client.ObjectKey{Name: nodeName}, node); err != nil {
@@ -462,6 +474,9 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
 
 		node.Annotations[annotationKey] = bootstrapAnnotationValue(rule.Name)
 		if err := r.Patch(ctx, node, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
+			if apierrors.IsConflict(err) {
+				metrics.StatusPatchConflicts.WithLabelValues("node_bootstrap_annotation", "patch").Inc()
+			}
 			return err
 		}
 
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -169,46 +169,30 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 
 		var successfullyPatchedRule *readinessv1alpha1.NodeReadinessRule
 
+		delta := nodeStatusDelta{
+			evaluations: make(map[string]readinessv1alpha1.NodeEvaluation),
+			failures:    make(map[string]*readinessv1alpha1.NodeFailure),
+		}
+
+		for _, eval := range rule.Status.NodeEvaluations {
+			if eval.NodeName == node.Name {
+				delta.evaluations[node.Name] = eval
+				break
+			}
+		}
+
+		var failedNode *readinessv1alpha1.NodeFailure
+		for _, failure := range rule.Status.FailedNodes {
+			if failure.NodeName == node.Name {
+				failureCopy := failure
+				failedNode = &failureCopy
+				break
+			}
+		}
+		delta.failures[node.Name] = failedNode
+
 		err := r.patchRuleStatusWithOptimisticLock(ctx, rule.Name, func(latestRule *readinessv1alpha1.NodeReadinessRule) {
-			// update only this specific node evaluation status
-			currEval := readinessv1alpha1.NodeEvaluation{}
-			for _, eval := range rule.Status.NodeEvaluations {
-				if eval.NodeName == node.Name {
-					currEval = eval
-					break
-				}
-			}
-
-			found := false
-			for i := range latestRule.Status.NodeEvaluations {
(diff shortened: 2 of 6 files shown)


bhuvan-somisetty:
@ajaysundark fixed the commit message (had a couple of bare #NNN references prow flagged as invalid). Should be clear now, ready whenever you get a chance to take a look.

internal/controller/node_controller.go
```diff
@@ -374,15 +363,15 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
...
+		stored := node.DeepCopy()
 
 		// Initialize annotations map if nil.
 		if node.Annotations == nil {
 			node.Annotations = make(map[string]string)
 		}
 
 		node.Annotations[annotationKey] = bootstrapAnnotationValue(ruleName)
-		if err := r.Patch(ctx, node, patch); err != nil {
+		if err := r.Patch(ctx, node, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{})); err != nil {
```
ajaysundark, line 465: It seems like annotations can merge cleanly (because it is a map?) so optimistic locking here is actually a bad idea, as can compete with Kubelet patches as well..

internal/controller/nodereadinessrule_controller.go
```diff
@@ -198,9 +201,19 @@ func (r *RuleReconciler) reconcileDelete(ctx context.Context, rule *readinessv1a
...
+		if err := r.Get(ctx, client.ObjectKey{Name: rule.Name}, latest); err != nil {
+			return client.IgnoreNotFound(err)
+		}
+		if !controllerutil.ContainsFinalizer(latest, finalizerName) {
+			return nil
+		}
+
+		stored := latest.DeepCopy()
+		controllerutil.RemoveFinalizer(latest, finalizerName)
+		return r.Patch(ctx, latest, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{}))
```
ajaysundark, line 217: Does this issue apply for finalizers as well? Or did you verify only for `rule.status`. confirm all the patches are required.

ajaysundark (commented):
Thanks for the PR. Did you confirm if all the patches are required, please write e2e tests if possible.

bhuvan-somisetty:
Good catch, both of you. Reverted the node annotation patch back to a plain `client.MergeFrom` — you're right that it merges cleanly as a map, so the optimistic lock there was just going to fight with kubelet's own patches for no reason.

Kept the lock on the two finalizer patches (add + remove) though, since `finalizers` is a `[]string` and a JSON merge patch replaces list fields wholesale rather than merging them — without the resourceVersion precondition a concurrent write there could get silently dropped. Left a comment at each site explaining why the two cases are handled differently so it's not ambiguous next time.

Also added two unit tests simulating concurrent RuleReconciler/NodeReconciler status writes to make sure the merge-by-node-name logic actually survives a real conflict and retry, not just the happy path.

bhuvan-somisetty:
Rebased onto latest main branch and resolved all merge conflicts. All tests are passing cleanly.

internal/controller/node_controller.go
ajaysundark: Recently, we had to add optimistic lock for annotation patch to fix a race condition with remove-taint. This PR had previously suggested that approach, but we initially thought it wasn't necessary. Thank you for pointing this out.

ajaysundark: Given it's already handled does this change here for node annotations still necessary?

internal/controller/helper.go
```diff
@@ -133,3 +135,56 @@ func filterStatusForExistingNodes(
...
+}
+
+// applyNodeStatusDelta merges delta into rule's NodeEvaluations/FailedNodes, replacing only the
+// entries for nodes present in delta and leaving every other node's entry untouched.
+//
+// This is the crux of fixing the lost-update bug described in #341: a naive full-slice
+// replacement of NodeEvaluations/FailedNodes (computed from a nodeList snapshot taken at the
+// start of a RuleReconciler sweep) would silently discard any per-node status update written
+// concurrently by NodeReconciler for a node this particular sweep didn't touch. Merging by node
+// name instead means each writer only ever overwrites the entries it just recomputed.
```
ajaysundark: Can we keep these comments concise? A function doc comment could self describe what functionality it is exposing. Context specific to this issue doesn't belong there, unless it carries long-term maintenance information.

https://go.dev/doc/comment#func is a good ref for comments.

internal/controller/node_controller.go
```diff
@@ -169,14 +169,7 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
...
 
-		err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			latestRule := &readinessv1alpha1.NodeReadinessRule{}
-			if err := r.Get(ctx, client.ObjectKey{Name: rule.Name}, latestRule); err != nil {
-				return err
-			}
-
-			patch := client.MergeFrom(latestRule.DeepCopy())
-
+		err := r.patchRuleStatusWithOptimisticLock(ctx, rule.Name, func(latestRule *readinessv1alpha1.NodeReadinessRule) bool {
```
ajaysundark, line 172: I expect this will spike 409s at API for our scale test. 

cc @vitorfloriano and I dont think we monitor API conflicts..it may reflect in reconcile latency though, not sure.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -266,41 +274,66 @@ func (r *RuleReadinessController) cleanupDeletedNodes(ctx context.Context, rule
...
+// instead of replacing NodeEvaluations/FailedNodes wholesale.
 //
 //nolint:unparam // Keep error return for future extensibility and API stability.
-func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule, nodeList *corev1.NodeList) error {
+func (r *RuleReadinessController) processAllNodesForRule(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule, nodeList *corev1.NodeList) (nodeStatusDelta, error) {
 	log := ctrl.LoggerFrom(ctx)
 
 	log.Info("Processing all nodes for rule", "rule", rule.Name, "totalNodes", len(nodeList.Items))
 
+	delta := nodeStatusDelta{
```
ajaysundark, line 299: I like this idea in theory, but want to look deeper into the details and risks of it. Will spend more time on this and add more comments.

ajaysundark (commented):
Thanks for the PR. I left some comments. Will have a second deeper look later.

ajaysundark:
xref - https://github.com/kubernetes-sigs/node-readiness-controller/pull/320#issuecomment-5382315086

@rawadhossain can we run an experiment with this patch and #320 API conflicts metric? That could be the baseline for #345, on how NRE helps with status handling.

cc @Karthik-K-N

internal/controller/node_controller.go
bhuvan-somisetty: Good push — checked, and it's still needed, but not for the reason I originally reached for. The annotation merge itself is fine (map, merges cleanly). What the lock actually guards is the `hasTaintBySpec` check right before the patch: without the resourceVersion precondition, a taint added between that check and the Patch call goes undetected, and we'd mark bootstrap complete on a node that still carries the taint. There's already a regression test for exactly this ("should not mark bootstrap completed when the rule taints concurrently"), and it predates this PR — this lock isn't new, this PR only touched variable naming here. Added a comment inline explaining this so it's not re-litigated later.

internal/controller/helper.go
bhuvan-somisetty: Fair, trimmed both comments (nodeStatusDelta and applyNodeStatusDelta) to describe what they do rather than the #341 story. Thanks for the go.dev link, wasn't following that convention closely enough here.

internal/controller/nodereadinessrule_controller.go
bhuvan-somisetty: Yes, applies to finalizers too, same root cause. Confirmed and covered: ensureFinalizer's add and reconcileDelete's finalizer removal both now go through client.MergeFromWithOptimisticLock (same pattern as addTaintBySpec/removeTaintBySpec). All four call sites (rule status, finalizer add, finalizer remove, cleanupDeletedNodes) share patchRuleStatusWithOptimisticLock or the equivalent inline pattern, so nothing is left on the plain-MergeFrom path.

internal/controller/node_controller.go
bhuvan-somisetty: Valid concern, and you're right that we weren't watching for it. Added node_readiness_status_patch_conflicts_total (labels: resource, operation), incremented on every 409 from an optimistic-locked patch — rule status, rule finalizer add/remove, node taint add/remove, and the node bootstrap annotation patch. Should give a direct signal instead of inferring it from reconcile latency. Happy to dig into the scale-test numbers with @vitorfloriano once this metric is out there.

bhuvan-somisetty:
Thanks for the thorough review. Pushed a follow-up commit addressing the open threads: trimmed the nodeStatusDelta/applyNodeStatusDelta doc comments to be self-describing, added node_readiness_status_patch_conflicts_total to make the 409 rate from optimistic locking observable (rule status, rule finalizer add/remove, node taint add/remove, bootstrap annotation), confirmed and documented why markBootstrapCompleted still needs its lock (guards the hasTaintBySpec check, not the annotation merge - existing regression test covers it), and confirmed finalizer add/remove both go through the same optimistic-lock pattern as rule status now. Left the nodeStatusDelta merge-by-name approach itself as-is since you want to look deeper there.

internal/metrics/metrics.go
```diff
@@ -145,6 +145,17 @@ var (
 		[]string{"rule", "condition"},
 	)
 
+	// StatusPatchConflicts tracks optimistic-lock conflicts on status/annotation patches.
+	// A rising rate here means concurrent writers are contending for the same object; it may
+	// show up as increased reconcile latency before it shows up as visible errors.
+	StatusPatchConflicts = prometheus.NewCounterVec(
```
ajaysundark: @bhuvan-somisetty Could we leave the new metrics from this PR scope to #320 and #288 to handle this holistically? 

cc @rawadhossain

bhuvan-somisetty: Makes sense, dropped it — pulled StatusPatchConflicts and its call sites back out. Will pick this up again in #320/#288 instead.

internal/controller/node_controller.go
```diff
@@ -446,7 +435,12 @@ func (r *RuleReadinessController) markBootstrapCompleted(ctx context.Context, no
 	deferred := false
 	annotationKey := bootstrapAnnotationKey(rule.GetUID())
 
-	// retry to handle conflict with concurrent node updates
+	// The optimistic lock here isn't guarding the annotation merge itself (that's map-valued
+	// and merges cleanly against concurrent writers, e.g. Kubelet). It guards the
+	// hasTaintBySpec check above: without it, a taint added between that check and the Patch
+	// below would go undetected, and we'd mark bootstrap complete on a node that still carries
+	// the taint. See the "should not mark bootstrap completed when the rule taints concurrently"
+	// test for the regression this prevents.
```
ajaysundark:
```suggestion
	// The optimistic lock here protects from a race-condition adding a taint between hasTaintBySpec
	// check and mark completed annotation patch from concurrent reconciliations. 
```

internal/controller/helper.go
```diff
@@ -133,3 +135,48 @@ func filterStatusForExistingNodes(
...
+// A nil value in failures clears any failure recorded for that node. evaluations only holds
+// entries for nodes freshly (re-)evaluated this sweep.
+type nodeStatusDelta struct {
+	evaluations map[string]readinessv1alpha1.NodeEvaluation
+	failures    map[string]*readinessv1alpha1.NodeFailure
+}
+
+// applyNodeStatusDelta merges delta into rule's NodeEvaluations/FailedNodes, replacing only the
+// entries for nodes present in delta and leaving every other node's entry untouched.
+func applyNodeStatusDelta(rule *readinessv1alpha1.NodeReadinessRule, delta nodeStatusDelta) {
```
ajaysundark, line 152: can we add unit-tests for these?

internal/controller/nodereadinessrule_controller.go
```diff
@@ -266,41 +274,66 @@ func (r *RuleReadinessController) cleanupDeletedNodes(ctx context.Context, rule
...
-		return r.Status().Patch(ctx, fresh, patch)
+		return true
 	})
 }
 
-// processAllNodesForRule processes all nodes when a rule changes.
+// processAllNodesForRule processes all nodes when a rule changes. It mutates rule.Status in place
+// (as before) and additionally returns a nodeStatusDelta describing exactly which nodes' status
+// this sweep changed, so updateRuleStatus can merge those changes into the latest stored status
+// instead of replacing NodeEvaluations/FailedNodes wholesale.
```
ajaysundark, line 291:
```suggestion
// processAllNodesForRule processes all nodes when a rule changes. It mutates rule.Status in place
// and additionally returns a nodeStatusDelta describing exactly which nodes' status are changed.
// so updateRuleStatus can merge those changes into the latest stored status
// instead of replacing NodeEvaluations/FailedNodes wholesale.
```

internal/controller/nodereadinessrule_controller.go
```diff
@@ -570,39 +603,63 @@ func (r *RuleReadinessController) removeRuleFromCache(ctx context.Context, ruleN
...
+// patchRuleStatusWithOptimisticLock fetches the latest NodeReadinessRule, lets mutate apply status
+// changes to it, and patches the result back with an optimistic-locked JSON merge patch. mutate
+// should return false if it made no changes, to skip an unnecessary Patch call.
+//
+// We use client.MergeFromWithOptimisticLock here for the same reason addTaintBySpec/
+// removeTaintBySpec do (see node_controller.go): a JSON merge patch replaces slice fields
+// (NodeEvaluations, AppliedNodes, FailedNodes) wholesale rather than merging them, so without a
+// resourceVersion precondition retry.RetryOnConflict can never observe a genuine conflict and a
+// concurrent status write from the other reconciler (RuleReconciler and NodeReconciler both patch
+// NodeReadinessRule.Status independently) can be silently overwritten.
```
ajaysundark:
```suggestion
// patchRuleStatusWithOptimisticLock fetches the latest NodeReadinessRule, and apply mutate status
// changes to it. It then patches the result to API with an optimistic-locked JSON merge patch. mutate
// should return false if it made no changes, to skip an unnecessary Patch call.
//
// We use client.MergeFromWithOptimisticLock here for a JSON merge patch replaces slice fields
// (NodeEvaluations, AppliedNodes, FailedNodes) wholesale rather than merging them, so without a
// resourceVersion precondition retry.RetryOnConflict can never observe a genuine conflict and a
// concurrent status write from the other reconciler (RuleReconciler and NodeReconciler both patch
// NodeReadinessRule.Status independently) can be silently overwritten.
```

internal/controller/nodereadinessrule_controller.go
```diff
@@ -570,39 +603,63 @@ func (r *RuleReadinessController) removeRuleFromCache(ctx context.Context, ruleN
...
+		return r.Status().Patch(ctx, latestRule, client.MergeFromWithOptions(stored, client.MergeFromWithOptimisticLock{}))
+	})
+}
+
+// updateRuleStatus updates the status of a NodeReadinessRule. delta carries the per-node
+// NodeEvaluations/FailedNodes changes processAllNodesForRule actually produced this reconcile;
+// it is merged into the latest stored status by node name (see applyNodeStatusDelta) rather than
+// replacing those fields wholesale, so a concurrent per-node update from NodeReconciler
+// (processNodeAgainstAllRules) for a node outside this sweep isn't silently discarded.
+func (r *RuleReadinessController) updateRuleStatus(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule, delta nodeStatusDelta) error {
```
ajaysundark, line 642: should updateRuleStatus skip patch when delta is empty?

ajaysundark (commented):
I left some comments, could you ptal?
Thanks for iterating on this.

bhuvan-somisetty:
Updated comments and docstrings per suggestions, added unit tests for applyNodeStatusDelta, and updated updateRuleStatus to return early when there are no status changes. Thanks for the review!

DsThakurRawat:
took this for a run locally since it has been open a while and main has moved under it.

the headline, because it is the thing that goes stale: i merged current `upstream/main`
(`6723e00`) into your branch head (`89bc5b4`) and it is clean. zero conflicts across the
three files that both touched, `go build ./...` and `go vet ./...` clean, and the full
suite green on the merged tree (`internal/controller` 16.3s, plus reporter, metrics and
webhook packages). so the rebase is not going to be the hard part.

on the premise, i wanted to see the two behaviours rather than trust the docs, so i put a
throwaway envtest package on the branch and made two writers race a real API server:

```
plain client.MergeFrom from a stale copy
  writer A held rv=199, object had moved to rv=203
  patch err  = <nil>
  isConflict = false
  -> patch SUCCEEDED from a stale copy

client.MergeFromWithOptimisticLock from a stale copy
  writer A held rv=207, object had moved to rv=208
  patch err  = Operation cannot be fulfilled on nodes "lockproof-opt":
               the object has been modified; please apply your changes to the latest
               version and try again
  isConflict = true
```

so the `retry.RetryOnConflict` wrappers really were unreachable, exactly as your
description says.

the slice half is the part i found more interesting, because on a map field the unlocked
patch is harmless. i seeded a node with taint `a`, let writer B append taint `b`, then had
a stale writer A write back `[a, c]`:

```
writer A snapshot taints: [a]
after writer B          : [a b]
after writer A          : [a c]
-> writer B's taint "b" was silently discarded. no error, no conflict.
```

which is why the optimistic lock on its own would not have been enough and the per node
delta is doing real work.

three things i checked because they are where this kind of change usually goes wrong, and
all three hold:

- `applyNodeStatusDelta` runs inside the `patchRuleStatusWithOptimisticLock` closure, so it
  re-applies against the freshly fetched object on every retry attempt rather than against
  the snapshot. that is the bit that would quietly undo the fix if it were outside.
- the `nodeStatusDelta` docstring says `AppliedNodes`, `ObservedGeneration` and
  `DryRunResults` have a single writer and are safe to overwrite. grepping for writes to
  those three, they only ever appear in `nodereadinessrule_controller.go`, never in
  `node_controller.go`, so excluding them is right.
- switching from wholesale replace to merge normally means entries for deleted nodes stop
  being pruned. `cleanupDeletedNodes` still re-filters against the fresh object inside its
  own closure, so that path is unaffected.

one thing i went looking for and did not find, worth recording so nobody else spends the
time: #381 landed after your merge base and adds `clearNodeFailure`, which mutates the
stale in memory `rule` rather than the fresh one. that looked like it would reintroduce
the same lost update through a new door. it does not. on the merged tree the node path
rebuilds `FailedNodes` inside the locked closure keeping other nodes' entries and
re-appending only this node's, so the clear is expressed per node and lands correctly. it
is correct but indirect, since `clearNodeFailure`'s only observable job is to make that
append loop find nothing.

which leads to the one suggestion i have, and it only becomes visible once both changes
sit on the same branch. after the merge there are two per node merge implementations: your
`applyNodeStatusDelta` in `helper.go`, and the inline evaluation and `FailedNodes` rebuild
in `processNodeAgainstAllRules`. they do the same thing with different code. folding the
node path onto `applyNodeStatusDelta` with a single node delta would leave one merge rule
to reason about instead of two, and would mean the sort order is applied consistently on
both paths. entirely your call whether that belongs here or in a follow up, since this PR
is already doing enough.

happy to share the envtest package if it is useful as a regression test, though it is
proving a client-go property rather than anything about this repo, so it probably does not
belong in the tree.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -570,39 +605,71 @@ func (r *RuleReadinessController) removeRuleFromCache(ctx context.Context, ruleN
...
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		latestRule := &readinessv1alpha1.NodeReadinessRule{}
+		if err := r.Get(ctx, client.ObjectKey{Name: ruleName}, latestRule); err != nil {
+			return err
+		}
+
+		stored := latestRule.DeepCopy()
+		if !mutate(latestRule) {
+			return nil
+		}
```
ajaysundark: Actually can we handle skip rule-patch here, instead of checking whether each of the field changed within the mutate function, so all write calls get this check?

```suggestion
		mutate(latestRule)
		if apiequality.Semantic.DeepEqual(stored.Status, latestRule.Status) {
			return nil
		}
```

internal/controller/nodereadinessrule_controller.go
```diff
@@ -570,39 +605,71 @@ func (r *RuleReadinessController) removeRuleFromCache(ctx context.Context, ruleN
...
+//
+// We use client.MergeFromWithOptimisticLock here for a JSON merge patch replaces slice fields
+// (NodeEvaluations, AppliedNodes, FailedNodes) wholesale rather than merging them, so without a
+// resourceVersion precondition retry.RetryOnConflict can never observe a genuine conflict and a
+// concurrent status write from the other reconciler (RuleReconciler and NodeReconciler both patch
+// NodeReadinessRule.Status independently) can be silently overwritten.
+func (r *RuleReadinessController) patchRuleStatusWithOptimisticLock(
+	ctx context.Context,
+	ruleName string,
+	mutate func(latest *readinessv1alpha1.NodeReadinessRule) (changed bool),
```
ajaysundark: nit: if you no longer need the return value of mutate, this can be void. This could simplify the mutate functions as they only need to transform, dont have to check each of the fields and return false/true.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -570,39 +605,71 @@ func (r *RuleReadinessController) removeRuleFromCache(ctx context.Context, ruleN
...
+	err := r.patchRuleStatusWithOptimisticLock(ctx, rule.Name, func(latestRule *readinessv1alpha1.NodeReadinessRule) bool {
+		appliedNodesChanged := !slices.Equal(latestRule.Status.AppliedNodes, rule.Status.AppliedNodes)
+		observedGenChanged := latestRule.Status.ObservedGeneration != rule.Status.ObservedGeneration
+		dryRunChanged := !reflect.DeepEqual(latestRule.Status.DryRunResults, rule.Status.DryRunResults)
+		hasDeltaChanges := len(delta.evaluations) > 0 || len(delta.failures) > 0
 
-		patch := client.MergeFrom(latestRule.DeepCopy())
+		if !hasDeltaChanges && !appliedNodesChanged && !observedGenChanged && !dryRunChanged {
+			return false
+		}
```
ajaysundark: this can be reverted to previous commit state, by checking the semantic comparison in the patch helper instead.

internal/controller/node_controller.go
```diff
@@ -215,12 +208,8 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			}
 			latestRule.Status.FailedNodes = updatedFailedNodes
```
ajaysundark, line 207: nit: you are sorting node order at rule-patch but appending here at the end. do you want to sortByNodeName(latestRule) here?

ajaysundark (commented):
Left some comments, ptal!

This is close, thanks for iterating on this PR.

ajaysundark:
@bhuvan-somisetty  Could you also check the failing lint?

ajaysundark:
>folding the node path onto applyNodeStatusDelta with a single node delta would leave one merge rule
> the sort order is applied consistently on both paths

@DsThakurRawat Thanks for taking a review pass on this. I saw the sort-order. But dont follow your single node-delta suggestion. Could you clarify?

> happy to share the envtest package if it is useful as a regression test

are you suggesting an e2e test for concurrent rule update scenario? Feel free send a PR if you want us to take a look.

bhuvan-somisetty:
Addressed all suggestions: updated patchRuleStatusWithOptimisticLock to check apiequality.Semantic.DeepEqual to skip unchanged status patches across all callers, simplified the mutate functions to void return, and added node name sorting in NodeReconciler. Thanks!

DsThakurRawat:
taking both questions in order.

1. the single node-delta suggestion, and why it is not just tidying.

`applyNodeStatusDelta` already takes a map keyed by node name, so one node is just a
one-entry map. the suggestion is that the inline block in `processNodeAgainstAllRules`
builds a `nodeStatusDelta` for this one node and calls it, rather than open coding
find-or-append a second time.

what makes it worth doing is that the two implementations do not agree today. the inline
block starts from a zero value:

```go
currEval := readinessv1alpha1.NodeEvaluation{}
for _, eval := range rule.Status.NodeEvaluations {
    if eval.NodeName == node.Name { currEval = eval; break }
}
```

then writes `currEval` into `latestRule` whether or not the loop found anything.
`evaluateRuleForNode` returns early on any taint add or remove failure, before
`updateNodeEvaluationStatus` runs, so in that case the cached rule holds no evaluation for
the node and the zero value is what gets persisted.

on your head `eb5a35e` merged with current main `4c6b9d2`: seeded status with a healthy
evaluation for `probe-node` and one for `other-node`, made the node patch fail, called
`processNodeAgainstAllRules`.

```
BEFORE: probe-node has a real evaluation, other-node has one too
AFTER : NodeEvaluations count = 2
  [0] nodeName=""           taintStatus=""        conditionResults=0
  [1] nodeName="other-node" taintStatus="Present" conditionResults=0
```

`probe-node`'s evaluation is gone, replaced by an entry with an empty `nodeName`.
`other-node` survives, so merge-by-name is doing its job. it is only the node being
reconciled that gets clobbered, and with the sort you just added the empty name sorts to
index 0.

2. the delta path does not have this, which is the actual argument.

same starting status, same failure expressed the way `processAllNodesForRule` expresses it,
a failure entry and no evaluation entry:

```
AFTER delta path: NodeEvaluations count = 2
  [0] nodeName="other-node" taintStatus="Present" conditionResults=0
  [1] nodeName="probe-node" taintStatus="Absent"  conditionResults=1
FailedNodes count = 1
  [0] nodeName="probe-node" reason="EvaluationError"
```

the evaluation survives and the failure is recorded next to it. `applyNodeStatusDelta` is
guarded on `len(delta.evaluations) > 0`, and `processAllNodesForRule` only puts an entry in
the map when it found one, so "nothing evaluated this sweep" means "leave what is there".
folding the node path onto it fixes the above as a side effect. the full suite is green on
that merged tree, `internal/controller` included, so nothing currently catches it.

3. on the envtest package, no, i was not proposing an e2e test.

what i had proves a client-go property, merge patch versus optimistic lock. that is true of
any controller and does not belong in this repo, which is why i hedged it rather than
offering it outright.

4. there is a real gap next to it though.

every concurrency test in `internal/controller` today is either the controller-runtime fake
client with an interceptor scripting a single conflict, or single threaded. none run two
writers concurrently against a real API server. the lost update this PR fixes is a genuine
race and a scripted conflict does not reproduce it. the test worth having is two goroutines
on envtest driving a rule-level and a node-level status write at the same rule, asserting
neither side's entry is lost, plus the empty `nodeName` case above pinned as a regression.

happy to send that as a PR, either onto this branch or as a follow up once this merges.
tell me which shape you want and i will put it up.

5. unrelated, the lint @bhuvan-somisetty is chasing.

`make lint` reproduces here as a single gofmt issue at
`internal/controller/helper_unit_test.go:238`: the map value comments in the new
`applyNodeStatusDelta` test need realigning, and there is a trailing blank line at end of
file. `gofmt -w internal/controller/helper_unit_test.go` clears it.

bhuvan-somisetty:
Good call on using applyNodeStatusDelta in processNodeAgainstAllRules! Refactored processNodeAgainstAllRules to construct a nodeStatusDelta for the target node and pass it to applyNodeStatusDelta, avoiding duplicate logic and preventing zero-value evaluation entries when evaluation fails. Thanks!

internal/controller/node_controller.go
```diff
@@ -172,58 +169,31 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
...
-				return err
+		var failedNode *readinessv1alpha1.NodeFailure
+		for _, failure := range rule.Status.FailedNodes {
+			if failure.NodeName == node.Name {
+				failureCopy := failure
+				failedNode = &failureCopy
+				break
 			}
+		}
+		delta.failures[node.Name] = failedNode
```
ajaysundark: This logic doesnt seem to be correct. Could you please check if it handles success and failure cases correctly? it doesnt seem to be clearing the stale status.

also please add unit-tests?

bhuvan-somisetty:
Updated NodeReconciler (processNodeAgainstAllRules) to handle evaluation success and failure distinctly when building nodeStatusDelta:
- On evaluation success: delta.failures[node.Name] = nil to clear any stale failure from latestRule.Status.FailedNodes, and delta.evaluations[node.Name] is set to the fresh evaluation.
- On evaluation failure: delta.failures[node.Name] is set to the recorded failure, and delta.evaluations[node.Name] is omitted so stale evaluations aren't persisted.

Also added unit tests for these cases in helper_unit_test.go. Thanks for catching this!

ajaysundark (approved):
/lgtm

PR #344, opened 2026-08-04
docs: add observability design

ajaysundark:
## Description
Add Observability Design Document for Node Readiness Controller

Assisted by: AGY / Gemini

## Related Issue
None

## Type of Change

/kind documentation
/kind feature
/kind design

## Testing
NA

## Checklist
- [ ] `make test` passes
- [ ] `make lint` passes

```release-note
NONE
```

### DIFF
--- a/docs/observability-design.md
+++ b/docs/observability-design.md
@@ -1,0 +1,184 @@
+# Observability Design Document: Node Readiness Controller
+
+### LFX Project: Granular Metrics and Improving SLIs from NRC
+
+This project brings better visibility into the Node Readiness Controller.
+
+The project answers two main questions:
+
+1. **Readiness as an SLO across owners:** Different teams own different readiness rules across the stack. Networking owns CNI readiness, storage owns CSI, and a hardware team owns GPU.
+When a node takes too long to become available, operators need to see which stage was on the critical path and how long it took, both on average and at the tail. They need to know whether the node was held by the built-in `Ready` condition from kubelet or by a specific readiness rule. The goal is to attribute delays to a stage and owner without reading controller logs, node logs, or raw YAML.
+
+2. **Fleet-wide availability:** Across the fleet, operators need to know which nodes or pools are not available for a given requirement. For example, they need to see what percentage of nodes are not GPU-ready. This is a roll-up over nodes, sliced by rule.
+
+### Goals
+
+* Add granular per-rule metrics.
+* Build a Grafana dashboard for fleet-wide SLOs.
+
+### Non-Goals
+
+* Build a Headlamp plugin. This was originally a goal, but we rescoped it as a follow-up project for LFX Term 3.
+
+## Design
+
+### 1. Personas
+
+We design observability around three distinct personas. Each persona asks different questions and looks at different interfaces.
+
+1. **Infrastructure Owners (Cluster Operators)**: These operators run the cluster and manage the node lifecycle. They need to know if the fleet is healthy and if the controller is working.
+2. **Component / Rule Owners**: These are teams that own the infrastructure components that gate node readiness, like CNI plugins or GPU drivers. They need to know whether / which specific check is failing.
+3. **Workload Owners (Application developers)**: These users run application pods on the cluster. They care about scheduling delays and node availability.
+
+---
+
+### 2. Signals
+
+We identify four telemetry surface in NRC: metric, status, event, or log and will follow below principles to assign each observability need to the right channel.
+
+#### Metric vs status vs event vs log decisions
+* Metrics to track aggregate cluster rates, durations, and counts over time. NRC will never use node names as metric labels for scale concerns.
+* Custom Resource status (`NodeReadinessRule.status`) will hold current rule health and evaluation summaries. New `NodeReadinessEvaluation` api will also be implemented for the individual node readiness status.
+* Kubernetes events will record single-node lifecycle actions like `TaintAdded`, `TaintRemoved`, `TaintAdopted`, and `BootstrapCompleted`.
+* Logs capture detailed controller traces. In managed Kubernetes environments, customers cannot read controller logs, so metrics and status must answer the operational questions for other personas.
+
+#### Core signals
+| Signal | Instrument | Decision |
+|---|---|---|
+| **Rule discovery (count, mode)** | Metric (`node_readiness_rules`) + `kubectl get nrr` | While `kubectl get` provides an API-level list, this gauge gives operators a controller-side view. The existing `node_readiness_rules_total` gauge violates naming conventions and ignores rule modes. We deprecate it in favor of `node_readiness_rules` without the `_total` suffix, adding labels to slice by `enforcement_mode` and `dry_run`. |
+| **Selector match (nodes grouped per rule)** | Metric (`node_readiness_rule_matched_nodes`) | Tracks how many nodes match each rule selector. Because we are removing per-node lists from `NodeReadinessRule.status` for scalability, this scrape-time gauge collector preserves visibility into selector reach. A zero value signals a misconfigured selector. |
+| **Nodes held or released per rule** | Metric (`node_readiness_rule_nodes`) | Shows the direct capacity impact of readiness gating on the cluster. The existing `node_readiness_nodes_by_state` gauge requires scale testing and tuning. We replace it with an always-on scrape-time collector that reads ground-truth `spec.taints` directly from the node cache. |
+| **Blocking conditions per rule** | Metric (`node_readiness_blocked_nodes`) | Assesses which infrastructure component is blocking nodes across the fleet. Today, `node_readiness_condition_failures_total` counts an unsatisfied condition during normal bootstrapping as a failure, which inflates rates with every reconcile. We replace it with a `node_readiness_blocked_nodes{rule,condition}` gauge. Per-node drill-downs will live in the upcoming `NodeReadinessEvaluation` CRD. |
+| **Per-node evaluation outcome** | CR status (`NodeReadinessEvaluationStatus`) + Node Events | Detailed per-node evaluation state moves to the new `NodeReadinessEvaluation` CRD. In addition, `markBootstrapCompleted` will emit a Kubernetes Node Event similar to taint operations so operators can inspect bootstrap completion in `kubectl describe node`. |
+| **Per-node readiness condition state** | `kube_node_status_condition` (kube-state-metrics) + Doc update | No controller code changes needed. `kube-state-metrics` already exports custom node conditions. Because we do not enforce a fixed prefix like `readiness.k8s.io` for conditions, we will update our documentation to recommend domain-scoped condition types (such as `network.example.io/CustomConditionReady`) so teams can group and monitor them in Grafana. |
+| **Taint add and remove activity** | Counter (`node_readiness_taint_operations_total`) + Node Events | Retain both channels. The counter tracks rate, churn, and flapping across the fleet, while Kubernetes Events tell the lifecycle story on individual nodes. |
+| **Operational failures** | Counter (`node_readiness_failures_total`) | Consolidates all operational errors across evaluations, taint operations, and status patches into a single counter. It uses a fixed vocabulary of snake_case reason strings for clear alerting. |
+| **Rule staleness and aliveness** | CR status conditions + workqueue depth | Remove `node_readiness_rule_last_reconciliation_timestamp_seconds`. Under `GenerationChangedPredicate` without resyncs, an unchanged rule is still actively enforced, making old timestamps false alarms. We track controller activity using evaluation rate and workqueue depth. |
+| **Bootstrap hold duration** | Histogram (`node_readiness_bootstrap_hold_duration_seconds`) | Measures total unschedulable time caused by readiness gating. A `taint_origin` label separates taints applied by the controller (`controller`) from those adopted at boot (`adopted`). |
+
+---
+
+### 3. Proposed Metric Surface
+
+The target metric surface freezes at `v1beta1`. Controller metrics use the `node_readiness_` prefix, and reporter metrics use the `node_readiness_reporter_` prefix.
+
+#### Controller metrics
+
+| Metric name | Type | Labels | Description | Stability tier |
+|---|---|---|---|---|
+| `node_readiness_build_info` | gauge | `version` | Controller binary version for managed-export monitoring. | ALPHA |
+| `node_readiness_rules` | gauge | `enforcement_mode`, `dry_run` | Number of rules by enforcement mode and dry-run state. | BETA |
+| `node_readiness_rule_matched_nodes` | gauge | `rule` | Number of nodes matching the rule selector. | BETA |
+| `node_readiness_rule_nodes` | gauge | `rule`, `state` (`held` or `released`) | Number of nodes currently gated or released by the rule. | BETA |
+| `node_readiness_blocked_nodes` | gauge | `rule`, `condition` | Number of nodes blocked by each required condition. | ALPHA |
+| `node_readiness_taint_operations_total` | counter | `rule`, `operation` (`add` or `remove`) | Total taint additions and removals performed by the controller. | BETA (STABLE candidate) |
+| `node_readiness_failures_total` | counter | `rule`, `reason` | Total operational failures across evaluation, taint operations, and status patches. | BETA |
+| `node_readiness_api_conflicts_total` | counter | `rule`, `operation` | Number of API write conflicts per retry attempt. | ALPHA |
+| `node_readiness_evaluation_duration_seconds` | histogram | `rule` | Duration of rule evaluation and enforcement per node. | ALPHA |
+| `node_readiness_enforcement_latency_seconds` | histogram | `rule`, `operation` (`add` or `remove`) | Latency from condition transition to taint addition or removal. | BETA |
+| `node_readiness_bootstrap_hold_duration_seconds` | histogram | `rule`, `taint_origin` (`controller` or `adopted`) | Total duration a node remained unschedulable due to readiness gating. | BETA |
+| `node_readiness_bootstrap_completed_total` | counter | `rule` | Number of nodes that completed bootstrap taint removal. | BETA |
+
+#### Reporter metrics (new proposed for daemonset)
+We expose these metrics on `/metrics` and `/healthz` for each reporter pod.
+
+| Metric name | Type | Labels | Description | Stability tier |
+|---|---|---|---|---|
+| `node_readiness_reporter_build_info` | gauge | `version` | Reporter binary version to track fleet version skew. | ALPHA |
+| `node_readiness_reporter_check_duration_seconds` | histogram | none | Duration of health probe checks. | ALPHA |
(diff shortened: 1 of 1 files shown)


ajaysundark:
cc @AnuragThePathak

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+* Add granular per-rule metrics.
+* Build a Grafana dashboard for fleet-wide SLOs.
+
+### Non-Goals
+
+* Build a Headlamp plugin. This was originally a goal, but we rescoped it as a follow-up project for LFX Term 3.
+
+## Design
+
+### 1. Personas
```
Karthik-K-N, line 25: Should we add user stories like personas?

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+1. **Readiness as an SLO across owners:** Different teams own different readiness rules across the stack. Networking owns CNI readiness, storage owns CSI, and a hardware team owns GPU.
+When a node takes too long to become available, operators need to see which stage was on the critical path and how long it took, both on average and at the tail. They need to know whether the node was held by the built-in `Ready` condition from kubelet or by a specific readiness rule. The goal is to attribute delays to a stage and owner without reading controller logs, node logs, or raw YAML.
+
+2. **Fleet-wide availability:** Across the fleet, operators need to know which nodes or pools are not available for a given requirement. For example, they need to see what percentage of nodes are not GPU-ready. This is a roll-up over nodes, sliced by rule.
+
+### Goals
+
+* Add granular per-rule metrics.
+* Build a Grafana dashboard for fleet-wide SLOs.
+
```
Karthik-K-N, line 18: we are also building Scrape-Time Collector, isn't a goal of this exercise?

Karthik-K-N (commented):
can we move this /docs/proposals
so in future we can keep all proposals at one place?

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+The `node_readiness_failures_total` counter uses a fixed set of snake_case reason strings:
+* `evaluation_error`
+* `add_taint_error` and `add_taint_conflict_exhausted`
+* `remove_taint_error` and `remove_taint_conflict_exhausted`
+* `status_patch_error` and `status_patch_conflict_exhausted`
+* `bootstrap_completion_mark_error`
+
+---
+
+### 4. New Scrape-Time Collector Surface
```
Karthik-K-N, line 123: is it mandatory requirement or performance improvement?

docs/observability-design.md
ajaysundark: yes, that's a good point. it could be a goal.

docs/observability-design.md
ajaysundark: User-stories felt bit redundant to me. I captured as 'questions' in SLIs/SLOs table below. Do you think I could may be extract that out at the top?

docs/observability-design.md
ajaysundark: Not measured individually. But I had few other optimizations like batched status updates (something similar to `endpointslice-updates-batch-period` [in kcm](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/), and it showed significant perf difference in my 5k+ node testing.

I'll add it in my list to present it to one of our discussions

docs/observability-design.md
Karthik-K-N: Yeah, What I thought is if any reader goes through this doc they can relate themself to any user story so the expectations may be clear.

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+* Build a Headlamp plugin. This was originally a goal, but we rescoped it as a follow-up project for LFX Term 3.
+
+## Design
+
+### 1. Personas
+
+We design observability around three distinct personas. Each persona asks different questions and looks at different interfaces.
+
+1. **Infrastructure Owners (Cluster Operators)**: These operators run the cluster and manage the node lifecycle. They need to know if the fleet is healthy and if the controller is working.
+2. **Component / Rule Owners**: These are teams that own the infrastructure components that gate node readiness, like CNI plugins or GPU drivers. They need to know whether / which specific check is failing.
```
AvineshTripathi, line 30:
```suggestion
2. **Component / Rule Owners**: These are teams that own the infrastructure components that gate node readiness, like CNI plugins or GPU drivers. They need to know whether / which specific component is delaying the readiness.
```

A check could be a generic check. I am fine with either but feel above suggestion fits more. Thoughts?

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+
+---
+
+### 2. Signals
+
+We identify four telemetry surface in NRC: metric, status, event, or log and will follow below principles to assign each observability need to the right channel.
+
+#### Metric vs status vs event vs log decisions
+* Metrics to track aggregate cluster rates, durations, and counts over time. NRC will never use node names as metric labels for scale concerns.
+* Custom Resource status (`NodeReadinessRule.status`) will hold current rule health and evaluation summaries. New `NodeReadinessEvaluation` api will also be implemented for the individual node readiness status.
```
AvineshTripathi, line 41: are we referring `NodeReadinessRule.status` to [this](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/api/v1alpha1/nodereadinessrule_types.go#L350) or the new api?

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+
+1. **Infrastructure Owners (Cluster Operators)**: These operators run the cluster and manage the node lifecycle. They need to know if the fleet is healthy and if the controller is working.
+2. **Component / Rule Owners**: These are teams that own the infrastructure components that gate node readiness, like CNI plugins or GPU drivers. They need to know whether / which specific check is failing.
+3. **Workload Owners (Application developers)**: These users run application pods on the cluster. They care about scheduling delays and node availability.
+
+---
+
+### 2. Signals
+
+We identify four telemetry surface in NRC: metric, status, event, or log and will follow below principles to assign each observability need to the right channel.
```
AvineshTripathi, line 37: do we consider annotations as a signal for bootstraping currently?

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+| **Selector match (nodes grouped per rule)** | Metric (`node_readiness_rule_matched_nodes`) | Tracks how many nodes match each rule selector. Because we are removing per-node lists from `NodeReadinessRule.status` for scalability, this scrape-time gauge collector preserves visibility into selector reach. A zero value signals a misconfigured selector. |
+| **Nodes held or released per rule** | Metric (`node_readiness_rule_nodes`) | Shows the direct capacity impact of readiness gating on the cluster. The existing `node_readiness_nodes_by_state` gauge requires scale testing and tuning. We replace it with an always-on scrape-time collector that reads ground-truth `spec.taints` directly from the node cache. |
+| **Blocking conditions per rule** | Metric (`node_readiness_blocked_nodes`) | Assesses which infrastructure component is blocking nodes across the fleet. Today, `node_readiness_condition_failures_total` counts an unsatisfied condition during normal bootstrapping as a failure, which inflates rates with every reconcile. We replace it with a `node_readiness_blocked_nodes{rule,condition}` gauge. Per-node drill-downs will live in the upcoming `NodeReadinessEvaluation` CRD. |
```
AvineshTripathi, line 51: how is it different from [`NodesByState`](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/metrics/metrics.go#L128)

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+|---|---|---|---|---|
+| `node_readiness_build_info` | gauge | `version` | Controller binary version for managed-export monitoring. | ALPHA |
+| `node_readiness_rules` | gauge | `enforcement_mode`, `dry_run` | Number of rules by enforcement mode and dry-run state. | BETA |
+| `node_readiness_rule_matched_nodes` | gauge | `rule` | Number of nodes matching the rule selector. | BETA |
+| `node_readiness_rule_nodes` | gauge | `rule`, `state` (`held` or `released`) | Number of nodes currently gated or released by the rule. | BETA |
+| `node_readiness_blocked_nodes` | gauge | `rule`, `condition` | Number of nodes blocked by each required condition. | ALPHA |
+| `node_readiness_taint_operations_total` | counter | `rule`, `operation` (`add` or `remove`) | Total taint additions and removals performed by the controller. | BETA (STABLE candidate) |
+| `node_readiness_failures_total` | counter | `rule`, `reason` | Total operational failures across evaluation, taint operations, and status patches. | BETA |
+| `node_readiness_api_conflicts_total` | counter | `rule`, `operation` | Number of API write conflicts per retry attempt. | ALPHA |
```
AvineshTripathi, line 76: is this needed? acc to the discussion [here](https://github.com/kubernetes-sigs/node-readiness-controller/pull/320) i think this can go as logs instead. wdyt?

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+| `node_readiness_bootstrap_completed_total` | counter | `rule` | Number of nodes that completed bootstrap taint removal. | BETA |
+
+#### Reporter metrics (new proposed for daemonset)
+We expose these metrics on `/metrics` and `/healthz` for each reporter pod.
+
+| Metric name | Type | Labels | Description | Stability tier |
+|---|---|---|---|---|
+| `node_readiness_reporter_build_info` | gauge | `version` | Reporter binary version to track fleet version skew. | ALPHA |
+| `node_readiness_reporter_check_duration_seconds` | histogram | none | Duration of health probe checks. | ALPHA |
+| `node_readiness_reporter_checks_total` | counter | `result` (`healthy`, `unhealthy`, or `error`) | Total probe check results over time. | ALPHA |
```
AvineshTripathi, line 89: will have labels ` (healthy, unhealthy, or error)`?

AvineshTripathi (commented):
added some questions, rest lgtm

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+| `node_readiness_rules_total` | Rename and reshape | Migrate to `node_readiness_rules{enforcement_mode, dry_run}` gauge. Drop `_total` suffix from gauge. |
+| `node_readiness_nodes_by_state` | Rename and reshape | Replace status-derived gauge with an always-on scrape-time collector `node_readiness_rule_nodes{rule, state="held"|"released"}`. The metrics need to be scale-tested and flag enabled by default |
+| `node_readiness_bootstrap_duration_seconds` | Rename and reshape | Migrate into `node_readiness_bootstrap_hold_duration_seconds{rule, taint_origin="adopted"}` histogram. |
+| `node_readiness_reconciliation_latency_seconds` | Rename and reshape | Rename to `node_readiness_enforcement_latency_seconds` to prevent collision with `controller-runtime` reconcile metrics. Normalize operation values to `add` and `remove`. |
+| `node_readiness_condition_failures_total` | Delete | Replace state-as-counter metric with the `node_readiness_blocked_nodes` gauge. |
+| `node_readiness_rule_last_reconciliation_timestamp_seconds` | Delete | Remove timestamp gauge. Unchanged rules are still actively enforced under `GenerationChangedPredicate` without resyncs. |
```
rawadhossain, line 166: right now this is metric (node_readiness_rule_last_reconciliation_timestamp_seconds) exists for every rule, even if a rule has zero matching nodes, and some of our tooling relies on that. If we remove it, is there another metric that provides the same guarantee, or is that something we need to figure out?

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+
+The target metric surface freezes at `v1beta1`. Controller metrics use the `node_readiness_` prefix, and reporter metrics use the `node_readiness_reporter_` prefix.
+
+#### Controller metrics
+
+| Metric name | Type | Labels | Description | Stability tier |
+|---|---|---|---|---|
+| `node_readiness_build_info` | gauge | `version` | Controller binary version for managed-export monitoring. | ALPHA |
+| `node_readiness_rules` | gauge | `enforcement_mode`, `dry_run` | Number of rules by enforcement mode and dry-run state. | BETA |
+| `node_readiness_rule_matched_nodes` | gauge | `rule` | Number of nodes matching the rule selector. | BETA |
```
rawadhossain, line 71: Since this metric (node_readiness_rule_matched_nodes) is tied to the new collector, should [this ](https://github.com/kubernetes-sigs/node-readiness-controller/pull/286) wait until the collector is added, or is the idea to implement it using the current approach first and migrate it later?

docs/observability-design.md
```diff
@@ -0,0 +1,184 @@
...
+Note 1: Some metrics are intentionally started as BETA as they feed to alerting rules and SLO dashboards. Marking them BETA protects from sudden breaking changes.
+Note 2: Additional metrics for Node-Readiness-Daemon (NRD) will be followed-up as a separate design update to this document.
+
+#### SLIs / SLOs
+
+Below table captures how NRC can be used to build alerts for above personas.
+
+| Persona | Question | SLI / SLO | Alerts and health checks |
+|---|---|---|---|
+| NRC cluster operators | Are there nodes stuck in a bootstrapping state? | **Nodes held:** `sum by (rule) (node_readiness_rule_nodes{state="held"})` | **Node Capacity held:** `sum(node_readiness_rule_nodes{state="held"}) > 0` and zero bootstrap completions for 30m.<br><br>**Error budget:** `rate(node_readiness_failures_total[10m]) > 0` for 15m.<br><br>**Controller health:** `up == 0`, `workqueue_depth > 1000`. |
```
rawadhossain, line 101: Would it make more sense for this to scale with fleet size or MaxConcurrentReconciles, rather than a fixed value? Queue depth can vary quite a bit depending on how many nodes enter the queue at once, so 1000 might mean different things for different cluster sizes. wdyt?

rawadhossain (commented):
Thanks for putting this together! left a few questions. Overall looks good to me.

docs/observability-design.md
ajaysundark: both. This is recommending us to follow best-practices[1] for including a `metav1.Condition[]` in `rule.status`; I'm imagining something like "Progressing" to show rule reconciliation in-progress, "Ready" when complete.. This is helpful to tell controller's status on whether there are still some nodes in the process queue for handling a rule.

@Karthik-K-N also brought a point on the need to align this convention, to help operators to use standard workflow like:
```
 kubectl wait --for=condition=ready node-readiness-rule ..
```

I agree it would be necessary for us to also define these here in a document. But I wanted to get the initial thoughts out and add more details to agree on the 'states'

[1] ref-
* https://book.kubebuilder.io/reference/good-practices.html#why-you-should-adopt-status-conditions
* https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

docs/observability-design.md
ajaysundark: That's a valid question maybe worth even clarifying at the doc. Annotation is 'metadata' and not observability signal. This is suppose to be treated only as internal state of controller. ie: no user can design a system on top based on the annotation we maintain as that could change.

docs/observability-design.md
ajaysundark: yes, I'm thinking they are good representative state generalizable across other reporters (eg: NRD) in the future as well.  I am not sure whether more granular states / categorical counts would be needed for the http health-checks, but these three seemed good to be a starting point. This is currently proposed ALPHA so we could come back to analyze this again in the future when we decide promote to beta.

docs/observability-design.md
ajaysundark: yeah, that reads a little better, but also underwhelms the need for same rule granularity signals.

For instance, I was thinking to capture that when multiple conditions in a rule map to a network readiness-rule, the rule owner would also want to distinguish whether the latency is from IPAMReady or CNIReady condition --  note, `kube_node_status_condition` (KSM) already gives this per-condition observability,. 

^^May be identifying 'rule owner' identical to 'component owner' is not the best fit. As a rule may include multiple components as we support multiple conditions.

docs/observability-design.md
ajaysundark: This proposal aims to further break down the earlier `node_readiness_nodes_by_state` to three more explicit per-node interested metrics: 
1. node_readiness_rule_matched_nodes (how nodes are mapped against our rules)
2. node_readiness_rule_nodes (what are their current status as seen by controller - held vs released as opposed to ‘ready, not_ready, bootstrapping’ as it felt we are conflating with kubelet’s Ready. Also, bootstrapping doesn’t apply for ‘continuous’)
3. node_readiness_blocked_nodes (what is pending on “held”)

docs/observability-design.md
AvineshTripathi: That's a valid point. How about `component(s)` or `one or more components` maybe

DsThakurRawat:
two things from the code side that might be worth folding into the doc.

the scrape-time collector reading `spec.taints` fixes a blind spot that exists on main right now, not just in principle. in the #382 shape (bootstrap-only rule, completion annotation present, taint orphaned by the startup race) the node controller skips the node before any evaluation, so nothing ever updates that node's entry in the status that `node_readiness_nodes_by_state` is computed from. the orphaned taint is structurally invisible to the current gauge, while the proposed collector reads the taint that is actually on the node and reports it as held. i reproduced the absorbing state locally earlier today (details in #382), so this would make a concrete example for the "gauges freeze and report stale numbers" limitation in section 4.

second, `node_readiness_api_conflicts_total` and the `*_conflict_exhausted` reason strings can't fire on today's code: `updateRuleStatus` re-gets the rule and patches with `client.MergeFrom` inside `retry.RetryOnConflict`, and a merge patch without an optimistic lock never gets a 409, so the retry path is dead and a lost concurrent write is silent, which is #341's finding. those metrics become meaningful once #343 switches the status patches to `MergeFromWithOptimisticLock`, so #343 may be worth listing as an explicit dependency in the rollout plan.

docs/observability-design.md
ajaysundark: sorry I am just getting to this. I missed the discussion on this. Could you add capture the concern again?

API conflicts could give us the signal of how API is affected when multiple node updates are aggregated into single rule.status. This could capture the improvements from migration to NRE design, where there should be less conflicts.

docs/observability-design.md
ajaysundark: >  some of our tooling relies on that.

sorry I'm just getting to this. It should be my oversight. What tooling rely on this? Could you clarify what's the downstream dependency and guarantee we may miss with this removal?

docs/observability-design.md
ajaysundark: yes, I think we clarified this on the PR> resolving

docs/observability-design.md
ajaysundark: Thanks for the callout. These are only examples and largely targeting our scale-test assumptions only.
I dont expect the values or alerts we provide to be used as-is at production clusters. They are examples / templates and are to be treated as such. The operators know their clusters and should update their dashboards accordingly.

ajaysundark:
> and a merge patch without an optimistic lock never gets a 409
@DsThakurRawat Thanks, this is correct, none of the api_conflicts from status updates are visible today until #343.

Issue #327, opened 2026-07-28
[FEATURE] [LFX] Headlamp plugin for node-readiness-controller


ajaysundark:
### Is your feature request related to a problem or existing issue? Please describe.

NRC state today is only visible via kubectl, logs, and the Prometheus metrics surface. It may be desirable for the Headlamp users to get answers to "why is this node not accepting workloads?" directly from the UX.

This issue tracks building a plugin for Headlamp to visualize NRC state directly, proposed as a CNCF LFX mentorship project (2026 Term 3). This is a follow-up deliverable from the Term 2 observability project.

### Describe the solution you'd like

High-level scope:

- Cluster-level view of NodeReadinessRule objects with live status (nodes matched / held / released per rule, derived from node taints).
- Node-level drill-down: applicable rules, per-condition pass/fail with failure reasons, NRC-managed taints, and related Events.
- Watch-based, API-state only: This doesnt replace the metrics / Grafana. Headlamp will give visibility into per-object questions and capture relationships on the map.
- Align with other standardized node conditions to explore as a generic lifecycle-conditions panel for Node Availability.
- Design doc reviewed by NRC and Headlamp maintainers before implementation.
- Validate and document the UX and scale tests / limits for the plugin.
- Published to Headlamp Artifact Hub with docs.

ajaysundark:
This for LFX Term3 mentorship project, please refrain from self-assigning the feature.

cc @AvineshTripathi @AnuragThePathak @illume

soumadipcodeEnthusiast:
Thanks for opening this — not self-assigning, just a couple of scoping questions ahead of Term 3 applications:

On generalizing to other node conditions beyond NRC — should the initial build stay NRC-specific (NodeReadinessRule + managed taints), with the generic lifecycle-conditions panel as a stretch goal? Or generic from day one?
Since this follows the Term 2 observability project — is there an existing design doc or plugin skeleton from that term worth reading first, or is this a clean start?

LarytheLord:
I have been reading toward this and two open PRs change what the plugin should be built on, so sharing before I write anything longer.

#344 says per-node lists are being removed from `NodeReadinessRule.status` for scalability, and that per-node drill-downs move to the new `NodeReadinessEvaluation` CRD in #345. That matters here because the natural way to build the node view, listing rules and reading `appliedNodes` and `failedNodes`, would be building on something you are deleting. So the data sources look like they split by scope rather than by screen:

[...]

PR #345, opened 2026-08-04
✨ feat: introduce NodeReadinessEvaluation (NRE) CRD and controller

Karthik-K-N:
## Description

Introduces the `NodeReadinessEvaluation` (NRE) custom resource and its dedicated reconciler. NRE provides a **per-node, read-only mirror** of the evaluated state of every applicable `NodeReadinessRule`, giving operators a single object to `kubectl get nre <node-name>` to see the full readiness picture of any node without having to cross-reference multiple rule statuses.

Discussion document: https://docs.google.com/document/d/1DOP1G6i__nQN8qbhlSvdswkUquMeSKQY4LU554AbcgM/edit?usp=sharing

### Feature Flag

The controller is opt-in via: `--enable-node-readiness-evaluation
`
## Checklist
- [ ] `make test` passes
- [ ] `make lint` passes

```release-note

```
Doc #(issue)


internal/controller/nodereadinessevaluation_controller.go
```diff
@@ -0,0 +1,395 @@
...
+	client.Client
+	Scheme     *runtime.Scheme
+	Controller *RuleReadinessController
+}
+
+// SetupWithManager wires the reconciler to watch:
+//   - Node objects (conditions, taints, labels)
+//   - NodeReadinessRule objects (enqueues all nodes matching the changed rule)
+func (r *NodeReadinessEvaluationReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
```
ajaysundark, line 58: @Karthik-K-N IIUC, this would trigger two parallel reconciliations for node updates (NodeReconciler and NREReconciler) and one is handling the output of the other.  I wonder if a separate controller for NRE is a better fitting pattern for this or should be bridged into NodeReconciler's watch?

Karthik-K-N: yes I thought about that  and I missed to point out why I chose this way, here are my thoughts
1. Initially tried combining both into the NodeReconciler, but then when I checked with best practices of writing controller, if we do so we will be merging two different operations into one, Node controller adding/removing taints,NRE just storing the result, One failure should not cause requeue for other and I think NodeReconcile should be as quick as possible as it affects the workloads.
2. Followed the existing pattern of having separate controllers like NRR, Node and NRE
3. Even if there is race and NRE reconciles first and later the Node, Since the NRE also watches for condition, taint and label, if something updated by the Node Rec then eventually it triggers the NRE. We will be eventually consistent.

These are the thoughts and I lean towards keeping them separate, but let me know what do you think.

DsThakurRawat:
i tested this branch and found two behaviours worth knowing about, plus one interaction with #315. all three are reproduced with runnable tests, happy to share them.

first, a rule-creation window that never heals. the NRE fan-out fires on the rule Create event (GenerationChangedPredicate only filters updates), but the shared ruleCache is only populated on the rule controller's second reconcile pass, because the first one adds the finalizer and returns RequeueAfter one second. so the NRE reconciles for every matching node run against a cache that cannot contain the new rule yet. for rules whose evaluation then changes a node, the taint write re-triggers the node watch and everything heals. but for a rule whose conditions are already satisfied everywhere, nothing mutates any node, the rule's status patch doesn't bump generation, heartbeats don't pass the node predicate (it compares only condition type to status), and informer resyncs are filtered as no-ops. the NRE just permanently misses the rule until some unrelated node change. deletion is fine, reconcileDelete empties the cache before dropping the finalizer. i think this is the concrete version of the question ajaysundark raised above: the reconciler's correctness depends on another controller's queue having run first. listing rules from the informer inside Reconcile instead of reading the private cache would remove the ordering dependency entirely.

second, this branch merges cleanly with #315, and the two disagree once combined. evaluateRuleForNode there honours conditionPolicy anyOf, while buildRuleEvaluation here ANDs every condition unconditionally. an anyOf rule with one of two conditions satisfied ends up enforced as satisfied (no taint) while the NRE for the same node reports it Unmatched. a shared evaluation helper honouring GetConditionPolicy() in both paths would keep them from drifting.

small one: nothing currently produces RuleStatusError, so the Errors count, State Pending, and the Evaluated False condition are unreachable in this iteration. fine if that's intentional groundwork, might deserve a TODO.

ajaysundark:
@Karthik-K-N is this ready for review?

Karthik-K-N:
> @Karthik-K-N is this ready for review?

yes , its ready for review

DsThakurRawat:
The rule-creation window I flagged here on Aug 11 has a second victim now, and this one is worse because it's the reporting surface. Walked the sequence on your head `7c0ecc42`: a rule Create passes GenerationChangedPredicate into this controller too, `ruleToNodeRequests` enqueues the matching nodes immediately, but the NRE reconciler reads `getApplicableRulesForNode`, which serves only the private `ruleCache`, and the cache isn't populated until the rule controller's *second* reconcile, because the first one just adds the finalizer and returns RequeueAfter one second (nodereadinessrule_controller.go:118-125, cache write at :139). So for a rule whose conditions are already satisfied on every matching node and whose taint is absent everywhere, the NRE usually reconciles against the empty cache and writes `state: Ready` with zero rule entries. Nothing corrects it after that: evaluation mutates no node, so no node event fires, and resync updates die on your equality predicates in the `For()`. Until something else happens to touch that node, the NRE keeps telling any reader (#327's history story, dashboards) that no rule applies to it while one actively does.

The Aug 11 fix suggestion covers both consumers, and it got cheaper to justify: have this controller read rules through the informer instead of the private cache, and the ordering dependency disappears rather than getting patched around.

Separately, I reproduced prow's lint failure locally so it can be named precisely: gocritic `ifElseChain` on the four-way taint transition at nodereadinessevaluation_controller.go:290, and unparam's unused `ctx` in the test's `sharedSetup` at :91. Both trivial.

What I checked that holds: rule deletions do reach this controller (in controller-runtime v0.24.1 a nil DeleteFunc defaults to passing, so GenerationChangedPredicate only filters updates), and your D1 shows the prune works once a reconcile runs; the CEL immutability on nodeSelector makes mapping by selector safe across rule updates; node deletion is covered by the ownerReference; and the single-writer separation is real, this reconciler writes nothing but its own status. One RBAC nit: the marker asks `delete` on nodereadinessevaluations but nothing in the code ever calls it, GC does that work.

internal/controller/nodereadinessevaluation_controller.go
ajaysundark: My concern with another reconciler is that we are reintroducing a shared (Node) state again between two controllers. looking at buildRuleEvaluation it is very similar to evaluateRuleForNode. reevaluating the node status between two controllers with caches seem like a bad idea (and that seems like repeating what we did and trying to change with rule.status). :/

I saw https://github.com/kubernetes-sigs/node-readiness-controller/pull/343 emitting individual nodeStatusDelta in rule-reconciliation, which could make adding NRE even cheaper than before. Happy to discuss more / hear your thoughts further on this.

internal/controller/nodereadinessevaluation_controller.go
```diff
@@ -0,0 +1,395 @@
...
+
+// recomputeNREStatus derives status.State, status.Summary, and status.Conditions
+// from the current rules slice. Called after every full rebuild so all top-level
+// fields stay consistent.
+func recomputeNREStatus(status *readinessv1alpha1.NodeReadinessEvaluationStatus) {
+	var matched, unmatched, activeTaints, errors int32
+
+	for _, r := range status.Rules {
+		switch r.RuleStatus {
+		case readinessv1alpha1.RuleStatusMatched:
```
ajaysundark: The Matched / Unmatched rules on status is bit unclear to me. Sorry I missed to get this at the doc, will look into it again.

internal/controller/nodereadinessevaluation_controller.go
```diff
@@ -0,0 +1,395 @@
...
+			errors++
+		}
+		if r.TaintStatus == readinessv1alpha1.TaintStatusPresent {
+			activeTaints++
+		}
+	}
+
+	status.Summary = readinessv1alpha1.EvaluationSummary{
+		MatchedRules:   &matched,
+		UnmatchedRules: &unmatched,
```
ajaysundark: does 'unmatched' (node-selector escaped?) add value to the per-node status? As a node-owner I would less-likely be interested in rules that are not affecting my node. 

Imagine a large cluster with many different node-pools. Setting the status of a node in pool-A for other rules associated with every other pools may be rather confusing to the user.

internal/controller/nodereadinessevaluation_controller.go
```diff
@@ -0,0 +1,395 @@
...
+type NodeReadinessEvaluationReconciler struct {
+	client.Client
+	Scheme     *runtime.Scheme
+	Controller *RuleReadinessController
+}
+
+// SetupWithManager wires the reconciler to watch:
+//   - Node objects (conditions, taints, labels)
+//   - NodeReadinessRule objects (enqueues all nodes matching the changed rule)
+func (r *NodeReadinessEvaluationReconciler) SetupWithManager(mgr ctrl.Manager) error {
```
ajaysundark, line 57: I expect, adding another controller will also increase the API interaction volume to a larger extent at scale.

ajaysundark (commented):
Thanks for looking into this @Karthik-K-N!

Left some comments, mostly on the need for separate controller loop for NRE. We could find sometime to align on the implementation plan to avoid adding to your work and delaying this.

ajaysundark:
> NRE reconciler reads getApplicableRulesForNode, which serves only the private ruleCache, and the cache isn't populated until the rule controller's second reconcile

Thanks for the callout @DsThakurRawat. agree, there are some risks with a managed cache and two controllers reconciling at two different snapshots of the cached data. we need to carefully handle the race-conditions.

internal/controller/nodereadinessevaluation_controller.go
Karthik-K-N: I think if we can finalize on this aspect, then other things will be easy to tackle
Since we use controller-runtime  and it uses a Shared Informer Cache, additional controller does not increase API server watch traffic or our memory footprint. but It does duplicate the reconcile queue (may be CPU usage) and small window of race, but it provides less blast radius and separate of concern, we have two controller doing different things.
Also I think we should consdier at what scale level we can expect this latency of having new controller

internal/controller/nodereadinessevaluation_controller.go
Karthik-K-N: True, I think we can drop the `UnmatchedRules` field.

internal/controller/nodereadinessevaluation_controller.go
Karthik-K-N: So the RuleStatus field helps to understand "**whether node satisfy every condition listed in the rule**"
Matched - All rule.Spec.Conditions are satisfied - ideally no taint
Unmatched - Any condition is not satisfied - there is taint on node

Do you think we should use better field name to convey this? 

Instead of Matched should I change that to RuleStatusSatisfied means the rule is satisfied by the node?

internal/controller/nodereadinessevaluation_controller.go
```diff
@@ -0,0 +1,449 @@
...
+		).
+		Complete(r)
+}
+
+// ruleToNodeRequests maps a NodeReadinessRule event to reconcile requests for
+// only the Nodes that match the rule's nodeSelector. Filtering here avoids
+// enqueueing the full cluster on every rule change — on a 5k-node cluster a
+// rule targeting 200 nodes produces 200 requests, not 5,000.
+//
+// nodeSelector is immutable enforeced via CEL validateion on the spec, so a changing
```
yindia:
```suggestion
// nodeSelector is immutable, enforced via CEL validation on the spec, so a changing
```

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	//
+	// +optional
+	TaintRemovedAt *metav1.Time `json:"taintRemovedAt,omitempty"`
+}
+
+// EvaluationSummary aggregates the results to provide a high-level overview.
+// +kubebuilder:validation:MinProperties=1
+type EvaluationSummary struct {
+	// matchedRules is the total number of rules currently matching this node.
+	//
```
yindia: Nit: Doc says "matching," but code counts satisfied rules (`nodereadinessevaluation_controller.go:384`). All entries in status.Rules already matched the selector, so this is ambiguous. 

Suggest either renaming to `satisfiedRules`, or keeping matchedRules but updating the description to match what the code counts

internal/controller/nodereadinessevaluation_controller.go
```diff
@@ -0,0 +1,449 @@
...
+	prevRules := make([]readinessv1alpha1.RuleEvaluation, len(nre.Status.Rules))
+	copy(prevRules, nre.Status.Rules)
+
+	patch := client.MergeFrom(nre.DeepCopy())
+
+	// Rebuild the full rules slice from scratch on every reconcile.
+	// listType=atomic means the controller owns the whole slice.
+	newRules := make([]readinessv1alpha1.RuleEvaluation, 0, len(applicableRules))
+	for _, rule := range applicableRules {
+		if !rule.DeletionTimestamp.IsZero() || rule.Spec.DryRun {
```
yindia: Dry-run rules apply to the node but are skipped here, so they never appear in status.Rules. The type doc promises "all applicable rules" (`types.go:64, :104, :239`), so the list is incomplete. Either reword the doc to "non-dry-run rules," or include them with a dryRun marker on the entry.

internal/controller/nodereadinessevaluation_controller.go
yindia: One scaling thing I noticed: the status patch is never a no-op. We rebuild the whole Rules slice and stamp LastEvaluationTime: now on each entry, then patch without checking whether the evaluation actually changed.

Where that bites is rule fanout. `ruleToNodeRequests` turns one rule change into a reconcile per matching node, and because every reconcile writes, a single rule edit on a large cluster becomes a status write per matching node, even where that node's result is unchanged. Scales with matching-nodes × rule-changes.

Could we skip the patch when only the timestamps would differ, so unaffected nodes don't get rewritten?

yindia:
@ajaysundark On the cache-ordering window: reading rules via an informer List in `Reconcile`, instead of the private `ruleCache`, closes it. The fan-out fires on the rule watch event, so the informer already has the new rule, whereas ruleCache isn't written until RuleReconciler's second pass. It also drops the ordering dependency, with no need to remove `ruleCache` itself. wdyt ?

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="nodeName is immutable and cannot be changed once set"
+	NodeName string `json:"nodeName,omitempty"`
+}
+
+// NodeReadinessEvaluationStatus defines the observed state of NodeReadinessEvaluation.
+// +kubebuilder:validation:MinProperties=1
+type NodeReadinessEvaluationStatus struct {
+	// conditions represent the latest available observations of the node's readiness evaluation state.
+	// Known condition types are:
+	// - "Evaluated": indicates whether the controller successfully evaluated all rules without errors.
```
ajaysundark, line 79: Is it possible to get Ready=False AND Evaluated=True on a NREStatus? I'm trying to see whether Evaluated is functionally identical signal as Ready or how it should be interpreted.

ajaysundark: Do I understand your comments correctly: you think Evaluated=False suggests 'errors' processing this node with NRC, whereas Ready=False means Taint is present (pending readiness on this node). So a normal user would subscribe to "kubectl wait --for=condition=Ready".

who / how do you see Evaluated should be consumed? and could you also clarify what 'errors' do you see this to be. is it like permission (for eg: to remove taint) or something else?

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// NodeEvaluationState indicates the overall readiness state of the node based on all rules.
```
ajaysundark: is this needed in addition to Ready=True/False/Unknown to show combinational status (compose all rule outcomes)? your comments seem to map to True, False & Unknown respectively.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	// taintRemovedAt is the time the controller successfully removed the taint
+	// after the node satisfied all conditions.
+	//
+	// +optional
+	TaintRemovedAt *metav1.Time `json:"taintRemovedAt,omitempty"`
+}
+
+// EvaluationSummary aggregates the results to provide a high-level overview.
+// +kubebuilder:validation:MinProperties=1
+type EvaluationSummary struct {
```
ajaysundark: I understand why you want this top level summary object here.

Maybe I am being extra careful since we are now moving out of a collapsing entity to isolated objects -- do we want this summary at the top, or is this something we could plan later? since we anticipate a handful or <10 rule matchings per node, I wonder if we could get away with having only the Conditions and RuleEvaluation[] array here?

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+// RuleEvaluation defines the outcome of evaluating a single NodeReadinessRule against this Node.
+type RuleEvaluation struct {
+	// ruleRef identifies the NodeReadinessRule this evaluation applies to.
+	// It bundles the rule's name (human-readable) and UID (tamper-proof identity)
+	// so that delete-and-recreate of a same-named rule is always detectable.
+	//
+	// +required
+	RuleRef RuleRef `json:"ruleRef,omitempty,omitzero"`
```
ajaysundark: super nit: could we keep RuleName and RuleUID directly here? Since these are keys to these objects, avoiding an extra deref would be more convenient for the consumers.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	//
+	// +optional
+	Summary EvaluationSummary `json:"summary,omitempty,omitzero"`
+
+	// rules contains the evaluation outcomes for all rules applicable to this node.
+	// The slice is owned and fully replaced by the controller on each reconcile (listType=atomic).
+	//
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=100
```
ajaysundark, line 101: NA - I'm not suggesting to change this. but I'll be worried about a node having to deal with 100 taints!~ .

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	// summary provides a quick, aggregated overview of the rules applied to this node.
+	//
+	// +optional
+	Summary EvaluationSummary `json:"summary,omitempty,omitzero"`
+
+	// rules contains the evaluation outcomes for all rules applicable to this node.
+	// The slice is owned and fully replaced by the controller on each reconcile (listType=atomic).
+	//
+	// +optional
+	// +listType=atomic
```
ajaysundark: listType=map with Key=ruleName?

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+
+// RuleStatus defines the result of evaluating a NodeReadinessRule's criteria against a Node.
+// +kubebuilder:validation:Enum=Satisfied;Unsatisfied;Error
+type RuleStatus string
+
+const (
+	// RuleStatusSatisfied indicates that the Node successfully met all conditions
+	// defined in the NodeReadinessRule. The controller will ensure the corresponding
+	// taint is removed so the node is unblocked.
+	RuleStatusSatisfied RuleStatus = "Satisfied"
```
ajaysundark, line 47: Trying to compare with the states we introduced here: https://github.com/kubernetes-sigs/node-readiness-controller/pull/344.

We added new state conventions for nodes: 'held' vs 'released' - but I think we saw them as "node's" state per rule. these are rule's state so "satisfied" makes sense. is that how you're thinking?

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+
+	// RuleStatusUnsatisfied indicates that one or more conditions defined in the
+	// NodeReadinessRule were not met. The controller will ensure the corresponding
+	// taint remains present to block scheduling.
+	RuleStatusUnsatisfied RuleStatus = "Unsatisfied"
+
+	// RuleStatusError indicates that a programmatic or configuration error occurred
+	// during the evaluation process (e.g., an invalid or unparseable NodeSelector).
```
ajaysundark: how will unparseable NodeSelector error is attributable to any specific node? I think rule-configuration error on the node's evaluation error is confusing.

I could imagine a more generic 'Blocked' for RuleStatus on the node - but then the question about - when do we decide unsatisfied vs blocked comes up as we do not have any 'time' budget on the rule-reconciliation.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	// so that delete-and-recreate of a same-named rule is always detectable.
+	//
+	// +required
+	RuleRef RuleRef `json:"ruleRef,omitempty,omitzero"`
+
+	// ruleStatus indicates the overall outcome of the rule's criteria against the Node.
+	//
+	// +required
+	RuleStatus RuleStatus `json:"ruleStatus,omitempty"`
+
```
ajaysundark, line 126: should we add the taintKey/effect here to help for self-reporting (instead of the active taints 'count' at the summary)?

I dunno, it would be redundant to rule.spec.taint, but for the headlamp kind of clients we are building, the CR may give a singular view of the readiness state on a 'node' level, without have to lookup again on the rule).

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=256
+	Reason string `json:"reason,omitempty"`
+
+	// message is a comprehensive, human-readable explanation providing further context.
+	//
+	// +optional
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=10240
```
ajaysundark, line 160: is it intentionally 10K?

ajaysundark: I was surprised at this thinking why should they be this long, but looks like k8s api gives 32k for Message and 1024! for Reason as well: https://github.com/kubernetes/apimachinery/blob/50d9b4a672b474db2e0bf61c968d87465b55eb56/pkg/apis/meta/v1/types.go#L1702. so ours is fine!

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	// +optional
+	TaintObservedAt *metav1.Time `json:"taintObservedAt,omitempty"`
+
+	// taintAddedAt is the time NRC itself applied the taint to the node.
+	// This field is nil when the taint was pre-existing and NRC adopted it rather than creating it.
+	// Use taintObservedAt for "how long has the node been blocked"; use this field to measure
+	// NRC's own apply latency (taintAddedAt - firstEvaluatedAt).
+	//
+	// +optional
+	TaintAddedAt *metav1.Time `json:"taintAddedAt,omitempty"`
```
ajaysundark, line 199: do we update this in continuous when Taint is reapplied?

ajaysundark: or should we split into FirstAppliedAt and CurrentTaintAppliedAt?

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+
+	// lastEvaluationTime records the exact moment the controller most recently assessed this rule.
+	//
+	// +required
+	LastEvaluationTime metav1.Time `json:"lastEvaluationTime,omitempty,omitzero"`
+
+	// firstEvaluatedAt is the time the rule was first assessed against this node.
+	//
+	// +optional
+	FirstEvaluatedAt *metav1.Time `json:"firstEvaluatedAt,omitempty"`
```
ajaysundark, line 180: nit: is this for observing possible rule-reconciliation delay on the node? I think the other TaintAt times  are also set at the first rule-reconciliation times. but I think I'm okay to keep this as it may bring some additional insights on rule's actions later.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	// +kubebuilder:validation:MaxItems=32
+	ReadinessConditions []ConditionEvaluationResult `json:"readinessConditions,omitempty"`
+
+	// lastEvaluationTime records the exact moment the controller most recently assessed this rule.
+	//
+	// +required
+	LastEvaluationTime metav1.Time `json:"lastEvaluationTime,omitempty,omitzero"`
```
ajaysundark, line 175: This means we are updating all the matching nodes' RuleEvals, irrespective of whether they are satisfied. 

Rule-reconciliations should be less frequent - only on rule modifications - so this seems fine, and 'freshness' details of evaluations maybe helpful.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	// +kubebuilder:validation:Minimum=0
+	Errors *int32 `json:"errors,omitempty"`
+}
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:resource:scope=Cluster,shortName=nre
+// +kubebuilder:printcolumn:name="Node",type=string,JSONPath=`.spec.nodeName`,description="The name of the target Node."
+// +kubebuilder:selectablefield:JSONPath=`.spec.nodeName`
+// +kubebuilder:printcolumn:name="State",type=string,JSONPath=`.status.state`,description="The overall readiness evaluation state of the node."
```
ajaysundark, line 213: ah, now I see where do you see these enums for state.

could we use Status=(?Condition==Ready).status?

api/v1alpha1/nodereadinessevaluation_types.go
ajaysundark: we would need this for multiple rule-workers working in parallel. and ref is not compatible with map, so it'd be good to consider flattening the rule references here.

internal/controller/nodereadinessevaluation_controller.go
ajaysundark: hmm. I maybe wrong about the API scale. but my primary concern is that NRE suppose to be a 'record' of what the controller enforced at node but a second reconciler recomputes the decision by further observations. This could bring other race and complexity of state handling.

For example, how do we distinguish even between kubelet added it vs another reconciler added it, when NRE is computed separately in its reconciler?  

> blast radius and separate of concern

I understand your concern on the implementation risks injecting this detail throughout the controller code. I was imagining #343 could simplify it for the rule, and maybe eventually #389 when we would have to handle the node updates all at just one place.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+	// The validation constraints enforce standard Kubernetes resource naming
+	// (RFC 1123 DNS Subdomain format), as defined in upstream apimachinery:
+	// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L198
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="nodeName is immutable and cannot be changed once set"
+	NodeName string `json:"nodeName,omitempty"`
```
AnuragThePathak: Considering the fact that spec is usually meant for defining the desired state of a CRD, may I know if you considered the trade-offs while putting NodeName as a spec?

My first thought was that we will be better off using 1:1 metadata.name for node and nre.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+
+const (
+	// NodeEvaluationStateReady indicates the node has successfully satisfied all applicable rules.
+	NodeEvaluationStateReady NodeEvaluationState = "Ready"
+
+	// NodeEvaluationStateNotReady indicates one or more applicable rules are currently not satisfied.
+	NodeEvaluationStateNotReady NodeEvaluationState = "NotReady"
+
+	// NodeEvaluationStatePending indicates the node is currently being evaluated or state is unknown.
+	NodeEvaluationStatePending NodeEvaluationState = "Pending"
```
AnuragThePathak: I have thought a lot about introducing a NodeEvaluationStateError as well here. But when I searched the bit, I found new convention is to minimize the cardinality of states.

So probably wanted to raise the point about have you thought if evaluation error state on top level can be worth adding?

Also my suggestion about pending state would be to get rid of it. I thought about it, I think it is useful for reconciliation process that takes time to complete, such as those requiring io operations or something cpu consuming, not necessary in our case. Feel free to correct me if needed, I didn't find any spec mandating need for pending.

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: Yes Correct, What I thought is, when the Ready/Available is False, User can check the status of Evaluated 
1. when Its True that means the Rule is successfully evaluated but its not satisfied on the Node.
2. when its False it can be Inprogress means the Rule is currently being reconciled
3. when its False it can be error means, there can be any intermittent error like api error or selector error

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: It feels redundant but having it will enhance the user experience
1. Helps in printing the result when queried the NRE using kubectl as we can't easily print the conditions
```
kubectl get nre
NAME             NODE             STATE       AGE
nre-node-worker1 node-worker1     Available   5m
nre-node-worker2 node-worker2     NotAvailable 2m
```
2. Currently its being markes as selectable field so we can do something like this
```
kubectl get nre --field-selector status.state=NotAvailable
```


Apart from this there is no other intention we should be good to drop as well.

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: Sure, I will update, Even this was added to provide the better UX for user, may be we can consider it later if needed, I will remove it for now.

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: True that, do you think as a controller we should provide some recommendation? or controller optimal performance when the Node has X taints?

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: Yes, Correct, Its the Rule's State

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: The Error is not attributed to Node rather the particular Rule of the Node. But yes having unparseble Nodeselector means it belong to all Nodes and which does not makes sense, I just removed this state. I think with this we no more need NodeEvaluationStatePending Node state as well, I just removed that.

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: We can add, its provides better clarity and better UX, avoid looks up

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: so currently we support only one taint lifecycle Absent->Present cycle, do you expect to support Absent→Present→Absent→Present cycle as well, does it have any impact on telemetry?

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: Yes, This is the time the NRC saw the rule. helps to see how long it took to evaluate the rule on tainted node

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: Array filtering was not supported when I check last time, but I can see if there are any alternatives if we really want to drop the state field.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+// EvaluationSummary aggregates the results to provide a high-level overview.
+// +kubebuilder:validation:MinProperties=1
+type EvaluationSummary struct {
+	// matchedRules is the total number of rules currently matching this node.
+	//
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	MatchedRules *int32 `json:"matchedRules,omitempty"`
```
AnuragThePathak: About this, I briefly saw the controller code, seemed like we were assigning it to satisfied rules. Let's name it clearly if that's the intention.

If matches rules mean no of rules that target the node, that we will have to update the implementation.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,260 @@
...
+	// The validation constraints enforce standard Kubernetes resource naming
+	// (RFC 1123 DNS Subdomain format), as defined in upstream apimachinery:
+	// https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L198
+	//
+	// +required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=253
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$`
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="nodeName is immutable and cannot be changed once set"
+	NodeName string `json:"nodeName,omitempty"`
```
AnuragThePathak, line 71: Considering the fact that spec is usually meant for defining the desired state of a CRD, may I know if you considered the trade-offs while putting NodeName as a spec?

My first thought was that we will be better off using 1:1 metadata.name for node and nre.

With on-going updates, the line maybe slightly misplaced, putting the context in the comment only

```go
NodeName string `json:"nodeName,omitempty"`
```

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,260 @@
...
+	Message string `json:"message,omitempty"`
+
+	// readinessConditions provides a detailed breakdown of each condition evaluation
+	// for this Node. This allows for granular debugging of which specific criteria passed/failed.
+	//
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	// +kubebuilder:validation:MaxItems=32
+	ReadinessConditions []ConditionEvaluationResult `json:"readinessConditions,omitempty"`
```
AnuragThePathak, line 170: I would be against reusing the old ConditionEvaluationResult. When adding default status, I ran into an issue of not being able to properly provide visibility into what's the actual status and what's the status we are calculating after applying default status, because of the name 'CurrentStatus'.

Now that we have a chance to introduce new API, I think replacing 'CurrentStatus' with something like 'EffectiveStatus' and 'ObservedStatus' can improve the observability.

AnuragThePathak (commented):
Haven't got a chance to look into the controller code, which I will get back soon.

About the API, loved the recent update, to drop error and loading states (loading was unnecessary anyway for our controller as it's not time taking by nature). Error being displayed at the rule level solved a lot of problems.

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: Yeah, The idea was to keep 1:1 mapping only and currently even the metadata.name is also set to nodename, We should be good to drop this field, but only thing is then the spec will become empty so I thought to keep the nodename as immutable field for better clarity.

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: I tend to agree with you, I think instead of currentState we should rename it to ObservedState and we can possibly add another field as EffectiveState which can help the user to understand the effective state which will be ObservedState if present or Default state if CurrentState not present
but
I think if we do this we should do it at both NRE and NRR level to avoid further confusion and keep them simple, Given that both are bit dependent or it may give two different results to user. I can take this up as a follow up task.

api/v1alpha1/nodereadinessevaluation_types.go
AnuragThePathak: Yes that's also right, spec will be empty. But unless there's some guideline or practice in common projects not to do so, we should be good I guess.

api/v1alpha1/nodereadinessevaluation_types.go
AnuragThePathak: Ah `state` suffix would be confusing I guess, `status` would be more consistent.

About the NRE and NRR, not quite sure about that. My personal preference would have been to do directly in NRR but obviously not possible due to schema constraints breaking existing schemas, so thought while introducing a new one, should be better doing it from day 1.

Anyway that's your call.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// lastEvaluationTime records the exact moment the controller most recently assessed this rule.
+	//
+	// +required
+	LastEvaluationTime metav1.Time `json:"lastEvaluationTime,omitempty,omitzero"`
```
AnuragThePathak, line 175: Randomly realized that we are putting suffix of "-At" everywhere except this one.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,260 @@
...
+	//
+	// +optional
+	// +listType=map
+	// +listMapKey=ruleName
+	// +kubebuilder:validation:MaxItems=100
+	Rules []RuleEvaluation `json:"rules,omitempty"`
+}
+
+// RuleEvaluation defines the outcome of evaluating a single NodeReadinessRule against this Node.
+type RuleEvaluation struct {
```
AnuragThePathak, line 106: Personally I think ConditionPolicy is a value that may be worth adding here.

api/v1alpha1/nodereadinessevaluation_types.go
```diff
@@ -0,0 +1,260 @@
...
+	// +kubebuilder:validation:MaxLength=253
+	TaintKey string `json:"taintKey,omitempty"`
+
+	// taintEffect is the effect of the taint managed by this rule, stamped at
+	// evaluation time so this entry is self-contained without requiring a lookup
+	// of the rule. Matches rule.spec.taint.effect.
+	//
+	// +required
+	// +kubebuilder:validation:Enum=NoSchedule;PreferNoSchedule;NoExecute
+	TaintEffect corev1.TaintEffect `json:"taintEffect,omitempty"`
```
AnuragThePathak, line 147: Also do we keep Taint key and effect separate for cases where people may be interested in filtering by the effect?

internal/controller/nodereadinessevaluation_controller.go
```diff
@@ -0,0 +1,331 @@
...
+		conditionResults = append(conditionResults, readinessv1alpha1.ConditionEvaluationResult{
+			Type:           condReq.Type,
+			CurrentStatus:  observedStatus,
+			RequiredStatus: condReq.RequiredStatus,
+			DefaultStatus:  condReq.GetDefaultStatus(),
+		})
+	}
+
+	ruleStatus := readinessv1alpha1.RuleStatusSatisfied
+	if !allConditionsSatisfied {
```
AnuragThePathak: You are completely ignoring conditionPolicy. Have to update that.

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: Nice catch, I updated it to `LastEvaluatedAt`

api/v1alpha1/nodereadinessevaluation_types.go
Karthik-K-N: Yes, Thats the idea.

internal/controller/nodereadinessevaluation_controller.go
Karthik-K-N: I have rebased and updated the code. Thank you

PR #348, opened 2026-08-04
fix: compare node labels in both directions in update predicate

### LINKED ISSUE #346: [BUG] Node label changes are silently dropped by the update predicate when a removed label had an empty value
### What happened?

`NodeReconciler`'s update predicate decides whether a Node event is worth reconciling by comparing conditions, taints and labels. The label comparison in `labelsEqual` ([internal/controller/helper.go#L108-L120](https://github.com/kubernetes-sigs/node-readiness-controller/blob/43bff46/internal/controller/helper.go#L108-L120)) reports two different label maps as equal when a key that was removed had an empty value:

```go
func labelsEqual(a, b map[string]string) bool {
	if len(a) != len(b) {
		return false
	}

	for k, v := range a {
		if b[k] != v {
			return false
		}
	}

	return true
}
```

`b[k]` returns the zero value `""` for a missing key, so when `v` is also `""` the mismatch is not detected. Combined with the length check, this means any label update that swaps one empty valued label for another empty valued label compares equal, the predicate returns `false`, and the Node event is dropped before `Reconcile` ever runs.

This matters because empty valued labels are the norm on Nodes, not an edge case. The `node-role.kubernetes.io/<role>` convention always uses an empty value, and this repository's own tests and examples select on `node-role.kubernetes.io/worker: ""`.

Note that the two sibling helpers in the same file, `conditionsEqual` and `taintsEqual`, both check key existence properly with `value, exists := aMap[key]`. `labelsEqual` is the only one of the three that omits the existence check, which is why this looks like an oversight rather than an intentional shortcut.

The consequence is that a Node that starts matching a rule's `nodeSelector` because of such a relabel is never evaluated against it, so the readiness taint is not applied and the gate the rule was created to enforce is silently bypassed. The reverse case leaves a stale taint in place. This does not self correct quickly: `conditionsEqual` deliberately ignores `lastHeartbeatTime`, so kubelet heartbeats do not generate reconcile events, and a Node in steady state can go a long time before any condition or taint change reopens the window. For `bootstrap-only` rules the bootstrap gate is missed entirely.

Realistic triggers are label updates that rewrite the whole label map in a single API call rather than adding and removing keys in separate calls: a `kubectl apply` against the Node, Cluster API or Karpenter Node label reconciliation, or a labeller DaemonSet writing its full desired label set.

### Steps to Reproduce

The behaviour is visible directly at the helper level. Adding this to `internal/controller/helper_unit_test.go` on `43bff46` fails:

```go
func TestLabelsEqualEmptyValueSwap(t *testing.T) {
	g := NewWithT(t)

	old := map[string]string{"node-role.kubernetes.io/worker": ""}
	new := map[string]string{"node-role.kubernetes.io/infra": ""}

	g.Expect(labelsEqual(old, new)).To(BeFalse())
}
```

```
--- FAIL: TestLabelsEqualEmptyValueSwap (0.00s)
    Expected
        <bool>: true
    to be false
```

End to end:

1. Create a rule whose `nodeSelector` is `matchLabels: {node-role.kubernetes.io/infra: ""}`, with taint `readiness.k8s.io/infra-ready:NoSchedule` and a condition the Node does not currently satisfy.
2. Take a Node whose only role label is `node-role.kubernetes.io/worker: ""`.
3. Replace the label map in one update so the Node carries `node-role.kubernetes.io/infra: ""` instead, for example `kubectl apply` with the new label set.
4. The Node now matches the rule, but no reconcile is triggered and `readiness.k8s.io/infra-ready` is never applied. `kubectl get nodereadinessrule infra-ready -o yaml` shows no `nodeEvaluations` entry for that Node.

Changing a label whose value is non empty, or adding a label without removing one, both work correctly, which is what makes this easy to miss.

### Expected Behavior

`labelsEqual` should report two label maps as equal only when they contain exactly the same keys with the same values, so that any real label change enqueues the Node for reconciliation.

`main` (commit `43bff46`)

### Kubernetes Version

Not version specific. Reproduced against envtest with Kubernetes 1.36.2 binaries.

I checked the open issues and PRs before filing. #235 and #342 also concern label changes, but they are about the controller not reacting correctly *after* it has seen a label change and decided the rule no longer applies. This issue is one level earlier: the event never reaches `Reconcile` at all because the predicate filters it out. No open PR touches `labelsEqual`. #339 adds new helpers to `helper.go` but leaves the function body unchanged.

I would like to work on this and will open a PR shortly.

tejassinghbhati:
## Description

`labelsEqual` combined a length check with a one-directional comparison:

```go
for k, v := range a {
	if b[k] != v {
		return false
	}
}
```

A key that is missing from `b` reads back as the zero value `""`, so when `v` is also `""` the difference is not detected. Because the lengths still match, removing an empty valued label and adding a different empty valued one in the same update compares equal. The `NodeReconciler` update predicate then computes `labelsChanged == false` and, if conditions and taints are also unchanged, drops the event before `Reconcile` runs.

Empty valued labels are the norm on Nodes rather than an edge case. The `node-role.kubernetes.io/<role>` convention always uses an empty value, and this repository's own tests and examples select on `node-role.kubernetes.io/worker: ""`. A Node that starts matching a rule's `nodeSelector` through such a relabel is therefore never evaluated against that rule, so its readiness taint is never applied. The node does not recover quickly either, because `conditionsEqual` deliberately ignores heartbeat updates, so nothing re-enqueues the Node until a real condition or taint change happens.

The fix uses `maps.Equal`, which compares keys and values in both directions and cannot regress in the same way. I kept the `labelsEqual` helper rather than inlining the call so the predicate still reads consistently next to `conditionsEqual` and `taintsEqual`, and so the regression test has a stable target. Happy to inline it if reviewers prefer.

Worth noting that `conditionsEqual` and `taintsEqual`, in the same file, both already check key existence with `value, exists := aMap[key]`. `labelsEqual` was the only one of the three missing that check.

## Related Issue

Fixes #346

## Type of Change

/kind bug

## Testing

Added `TestLabelsEqual` to `internal/controller/helper_unit_test.go`, a table driven test that also asserts the comparison is symmetric. The two empty value cases fail on `main` and pass with this change:

```
--- FAIL: TestLabelsEqualEmptyValueSwap (0.00s)
    Expected
        <bool>: true
    to be false
```

Full run after the change, using envtest with Kubernetes 1.36.2 binaries:

```
--- PASS: TestLabelsEqual (0.00s)
    --- PASS: TestLabelsEqual/identical_labels
    --- PASS: TestLabelsEqual/different_value_for_the_same_key
    --- PASS: TestLabelsEqual/extra_key
    --- PASS: TestLabelsEqual/empty_valued_role_label_swapped_for_another
    --- PASS: TestLabelsEqual/disjoint_empty_valued_keys_alongside_a_shared_label
    --- PASS: TestLabelsEqual/empty_valued_label_kept_unchanged
    --- PASS: TestLabelsEqual/both_nil
    --- PASS: TestLabelsEqual/nil_compared_to_empty_valued_label
ok  	sigs.k8s.io/node-readiness-controller/internal/controller
```

The existing controller suite is unaffected: 62 of 62 specs pass, and `internal/webhook` plus `cmd/readiness-condition-reporter` pass as well.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fixed the node update predicate dropping label changes when a removed label had an empty value, which could leave a node without the readiness taint of a rule it had just started matching.
```

### DIFF
--- a/internal/controller/helper.go
+++ b/internal/controller/helper.go
@@ -105,13 +105,7 @@ func taintsEqual(a, b []corev1.Taint) bool {
 	return true
 }
 
-// labelsEqual checks if two label maps are equal.
-//
-// maps.Equal compares both keys and values in both directions. A hand-rolled
-// length check plus a one-directional `b[k] != v` comparison is not sufficient:
-// a missing key yields the zero value "", so removing an empty valued label and
-// adding a different empty valued one in the same update would compare equal and
-// the Node event would be dropped by the update predicate.
+// labelsEqual checks if two label maps hold the same keys with the same values.
 func labelsEqual(a, b map[string]string) bool {
 	return maps.Equal(a, b)
 }
--- a/internal/controller/helper.go
+++ b/internal/controller/helper.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"encoding/json"
+	"maps"
 
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/types"
@@ -105,16 +106,12 @@ func taintsEqual(a, b []corev1.Taint) bool {
 }
 
 // labelsEqual checks if two label maps are equal.
+//
+// maps.Equal compares both keys and values in both directions. A hand-rolled
+// length check plus a one-directional `b[k] != v` comparison is not sufficient:
+// a missing key yields the zero value "", so removing an empty valued label and
+// adding a different empty valued one in the same update would compare equal and
+// the Node event would be dropped by the update predicate.
 func labelsEqual(a, b map[string]string) bool {
-	if len(a) != len(b) {
-		return false
-	}
-
-	for k, v := range a {
-		if b[k] != v {
-			return false
-		}
-	}
-
-	return true
+	return maps.Equal(a, b)
 }
--- a/internal/controller/helper_unit_test.go
+++ b/internal/controller/helper_unit_test.go
@@ -57,6 +57,76 @@ func TestLegacyBootstrapAnnotationKey(t *testing.T) {
 	g.Expect(key).To(Equal("readiness.k8s.io/bootstrap-completed-my-rule"))
 }
 
+func TestLabelsEqual(t *testing.T) {
+	tests := []struct {
+		name     string
+		a        map[string]string
+		b        map[string]string
+		expected bool
+	}{
+		{
+			name:     "identical labels",
+			a:        map[string]string{"env": "prod"},
+			b:        map[string]string{"env": "prod"},
+			expected: true,
+		},
+		{
+			name:     "different value for the same key",
+			a:        map[string]string{"env": "prod"},
+			b:        map[string]string{"env": "staging"},
+			expected: false,
+		},
+		{
+			name:     "extra key",
+			a:        map[string]string{"env": "prod"},
+			b:        map[string]string{"env": "prod", "tier": "frontend"},
+			expected: false,
+		},
+		{
+			// A missing key reads back as the zero value "", so an empty valued
+			// label that is swapped for a different empty valued one must still
+			// be reported as a change.
+			name:     "empty valued role label swapped for another",
+			a:        map[string]string{"node-role.kubernetes.io/worker": ""},
+			b:        map[string]string{"node-role.kubernetes.io/infra": ""},
+			expected: false,
+		},
+		{
+			name:     "disjoint empty valued keys alongside a shared label",
+			a:        map[string]string{"a": "", "shared": "x"},
+			b:        map[string]string{"b": "", "shared": "x"},
+			expected: false,
+		},
+		{
+			name:     "empty valued label kept unchanged",
+			a:        map[string]string{"node-role.kubernetes.io/worker": ""},
+			b:        map[string]string{"node-role.kubernetes.io/worker": ""},
+			expected: true,
+		},
+		{
+			name:     "both nil",
+			a:        nil,
+			b:        nil,
+			expected: true,
+		},
+		{
+			name:     "nil compared to empty valued label",
+			a:        nil,
+			b:        map[string]string{"node-role.kubernetes.io/worker": ""},
+			expected: false,
+		},
+	}
+
+	for _, tt := range tests {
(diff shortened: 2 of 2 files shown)


internal/controller/helper.go
```diff
@@ -105,16 +106,12 @@ func taintsEqual(a, b []corev1.Taint) bool {
 }
 
 // labelsEqual checks if two label maps are equal.
+//
+// maps.Equal compares both keys and values in both directions. A hand-rolled
+// length check plus a one-directional `b[k] != v` comparison is not sufficient:
+// a missing key yields the zero value "", so removing an empty valued label and
+// adding a different empty valued one in the same update would compare equal and
+// the Node event would be dropped by the update predicate.
```
Karthik-K-N: I think its better to keep the function comments minimal.
Lets capture only about the current code in the comments, 
A future reader might not know what is b[k].

Karthik-K-N (changes requested):

internal/controller/helper.go
tejassinghbhati: Good point, that comment was describing the old implementation rather than the code as it stands, and `b[k]` no longer appears anywhere in the function. Trimmed it to a single line in b4faf56 so it matches the style of `conditionsEqual` and `taintsEqual` above it:

```go
// labelsEqual checks if two label maps hold the same keys with the same values.
func labelsEqual(a, b map[string]string) bool {
	return maps.Equal(a, b)
}
```

I reworded the comment on the table entry the same way, so it describes the shape of the input instead of commenting on removed code. The reasoning for the change is kept in the commit message and in #346 where it stays discoverable without cluttering the function.

tejassinghbhati:
Thanks for the review. Pushed b4faf56 which cuts the comment down to a single line describing what the function does now, so there is no longer any reference to the old implementation or to identifiers that are not in the code any more. I gave the comment on the test table entry the same treatment.

Ready for another look whenever you have a moment.

Karthik-K-N:
/ok-to-test

LGTM
/assign @ajaysundark for final approval

ajaysundark:
/approve


Thanks for the patch!

PR #339, opened 2026-08-02
fix(controller): handle node deletion, label selector unmatching, and status drift in NodeReconciler

bhuvan-somisetty:
## Description
This PR resolves issue #338 by fixing systemic node lifecycle state management, orphaned taint cleanup, and status synchronization bugs across `NodeReconciler` and `RuleReconciler`:

1. **Node Deletion Cleanup**: When a node is deleted from the cluster, `NodeReconciler.Reconcile` handles the `apierrors.IsNotFound(err)` condition via `handleNodeDeletion(ctx, nodeName)`. It purges the deleted node from `rule.Status.NodeEvaluations`, `rule.Status.AppliedNodes`, and `rule.Status.FailedNodes` across all cached rules and updates Prometheus `node_readiness_nodes_by_state` gauge metrics.
2. **Comprehensive `cleanupDeletedNodes`**: Updated `RuleReconciler.cleanupDeletedNodes` to filter `AppliedNodes` and `FailedNodes` alongside `NodeEvaluations`.
3. **Orphaned Taint Removal on Selector Unmatch**: In `processNodeAgainstAllRules`, when a node's labels change such that it no longer matches a rule's `spec.NodeSelector`, the controller checks if the node holds `rule.Spec.Taint` or has status entries. If so, it invokes `removeTaintBySpec` to remove the orphaned taint and strips the node from rule status.
4. **Real-time `AppliedNodes` Synchronization**: Updated `NodeReconciler` status patch logic to append matching nodes to `rule.Status.AppliedNodes` upon successful evaluation or remove them if evaluation fails/unmatches.

## Related Issue
Fixes #338

## Type of Change
/kind bug

## Testing
- Unit and integration tests added in `internal/controller/node_controller_test.go` and `internal/controller/nodereadinessrule_controller_test.go` covering node deletion handling, label selector unmatch taint removal, and `AppliedNodes` status synchronization.
- Verified test suite passes locally.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

## Does this PR introduce a user-facing change?

```release-note
Fix state inconsistency, orphaned readiness taints, and status drift in NodeReconciler on node deletion and label selector changes.
```

Signed-off-by: bhuvan-somisetty <somisettybhuvan5@gmail.com>

Issue #235, opened 2026-05-10
Node label changes can leave stale rule-managed taints after nodeSelector no longer matches


sahitya-chandra:
## What happened

When a NodeReadinessRule has already managed a node, and that node later changes labels so it no longer matches the rule nodeSelector, the node reconciler does not remove the taint previously managed by that rule.

The node update predicate does enqueue reconciliation on label changes, but the reconciliation path only loads rules that currently match the node. After the label change, the rule is no longer considered, so the stale taint is never cleaned up

## Why this matters

A node can remain incorrectly tainted after it leaves the rule target set. For NoSchedule taints, this can prevent workloads from scheduling on the node. For NoExecute taints, this can be more disruptive because pods without tolerations may be evicted or kept away even though the rule no longer applies to that node

## Reproduction

1. Create a node with label `env=test`
2. Create a continuous NodeReadinessRule with `nodeSelector.matchLabels.env=test` and a taint such as `readiness.k8s.io/test-taint:NoSchedule`
3. Reconcile the node while the rule applies, so the rule records a NodeEvaluation for that node and manages/adopts the taint
4. Change the node label to `env=other`
5. Reconcile the node again

## Actual behavior

The taint remains on the node

The relevant log line from a focused regression test is:

```text
Processing node against rules {"node": "node-controller-test-node", "ruleCount": 0}
```

Because the rule no longer matches the node, `getApplicableRulesForNode` returns no rules, and the cleanup path never runs

## Expected behavior

If a rule previously managed a node, and the node stops matching that rule due to a label change, the controller should remove the rule-managed taint and clean up the node-specific rule status for that node

The cleanup should be scoped to nodes that have evidence of prior rule management, such as existing NodeEvaluation or appliedNodes status, so the controller does not remove unrelated pre-existing taints from nodes that never matched the rule

## Notes

Rule deletion and rule nodeSelector changes already have cleanup paths. The missing case is node label changes that make an already-managed node stop matching the selector

tejassinghbhati:
Small correction to the notes here, carried over from #347 which I am closing as a duplicate of this.

The notes say rule deletion already has a cleanup path. It does, but `cleanupTaintsForRule` filters by `ruleAppliesTo` in the same way `getApplicableRulesForNode` does, so a node that was relabelled out of the selector is skipped there too:

```go
for _, node := range nodeList.Items {
	if !r.ruleAppliesTo(ctx, rule, &node) {
		continue
	}
```

[...]

PR #349, opened 2026-08-04
fix: clean up taints on rule deletion for nodes that stopped matching the selector

### LINKED ISSUE #342: [BUG] Orphaned readiness taint remains when a node's labels no longer match a rule's selector
### What happened?

Split out of #338 per @ajaysundark's [review comment](https://github.com/kubernetes-sigs/node-readiness-controller/issues/338#issuecomment-5163027297): when a node's labels change so it no longer matches a `NodeReadinessRule`'s `spec.nodeSelector`, `NodeReconciler` simply stops seeing the rule as applicable via `getApplicableRulesForNode` and moves on. Any taint that rule previously applied to the node is never removed, and the node's stale entries in `rule.Status.NodeEvaluations`/`AppliedNodes`/`FailedNodes` are never cleaned up.

### Steps to Reproduce

1. Create a rule with `nodeSelector: {tier: frontend}` and taint `readiness.k8s.io/frontend-ready:NoSchedule`.
2. Label a node `tier=frontend`; once a required condition is `False`, the controller applies the taint.
3. Relabel the node: `kubectl label node worker-1 tier=backend --overwrite`.
4. Inspect `worker-1.spec.taints` — `readiness.k8s.io/frontend-ready` is still present even though the node no longer matches `tier: frontend`.

### Expected Behavior

Ideally the taint gets removed and the node drops out of the rule's status once it stops matching. In practice this needs a reliable way for `NodeReconciler` to detect "used to match, no longer does" without depending on `rule.Status` as the signal for what to act on, since the controller's own previously-written status isn't a trustworthy basis for deciding what action to take next (a first attempt at this in #339 did exactly that, and @ajaysundark flagged it as fragile in review).

`main` (commit `43bff46`)

### Kubernetes Version

Not version-specific.

Carried over from review discussion on #338/#339:

- The straightforward alternative — have `NodeReconciler` evaluate *all* rules (not just currently-matching ones) on every node update, so it can catch "still has a managed taint but no longer matches" — works, but is expensive, especially with the extra status writes it implies for every node event.
- If a manual relabel (e.g. via script) causes this, removing the taint could reasonably be handled by the same script/automation rather than the controller. `dry-run` mode also exists to catch this kind of misconfiguration before it happens.
- Given that, this is a real gap but arguably not high-impact, and needs a proper design discussion on how to detect the "used to match" transition reliably before implementing anything — rather than being bundled into the node-deletion cleanup fix in #338/#339.

### LINKED ISSUE #347: [BUG] Deleting a NodeReadinessRule leaves a permanently orphaned taint on nodes that stopped matching its selector
### What happened?

Deleting a `NodeReadinessRule` can leave its taint on a Node permanently, with no controller left that is able to remove it.

`reconcileDelete` calls `cleanupTaintsForRule` and then removes the finalizer. `cleanupTaintsForRule` ([internal/controller/nodereadinessrule_controller.go#L670-L697](https://github.com/kubernetes-sigs/node-readiness-controller/blob/43bff46/internal/controller/nodereadinessrule_controller.go#L670-L697)) only considers Nodes that currently match the rule's `nodeSelector`:

```go
	if !r.ruleAppliesTo(ctx, rule, &node) {
		continue
	}

	if r.hasTaintBySpec(&node, rule.Spec.Taint) {
		...
		if err := r.removeTaintBySpec(ctx, &node, rule.Spec.Taint, rule.Name); err != nil {
```

If a Node was tainted by the rule and its labels later changed so it no longer matches the selector, that Node is skipped during cleanup. The finalizer is then removed and the rule object is deleted, so the taint is stranded. Nothing in the system references it any more, and no future reconcile can remove it. The only remedy is for an operator to notice and delete the taint by hand.

This is a strictly worse variant of #235 and #342. In those, the rule still exists, so a fix has something to act on later and the taint is recoverable. On the deletion path the owner disappears, which makes the taint permanent. For a `NoSchedule` taint the Node stops accepting new workloads indefinitely. For `NoExecute` it also keeps evicting Pods that lack a toleration, on a Node that no rule governs any more.

I want to flag one thing about scope, because it is what makes this fixable independently of the #342 discussion. The reason #342 is hard is that `NodeReconciler` needs a trustworthy way to detect "this Node used to match and no longer does" on every Node event, and the review discussion on #339 concluded that leaning on `rule.Status` for that is fragile. The deletion path does not have that problem. It runs once, it is terminal, and the rule is being torn down, so the set of Nodes the rule ever managed is exactly what needs cleaning. `rule.Status` is a reasonable signal there in a way it is not for steady state reconciliation.

A blanket "remove this taint key from every Node" would not be safe, because the validating webhook only rejects rules that share a taint key and effect when their selectors *overlap*. Two rules can legitimately share `readiness.k8s.io/x:NoSchedule` with disjoint selectors, and deleting one must not strip the taint from Nodes owned by the other. Restricting cleanup to Nodes that either currently match the selector or appear in this rule's own `status.nodeEvaluations` / `status.appliedNodes` avoids that, since a Node managed only by the sibling rule never appears in this rule's status.

### Steps to Reproduce

1. Create a Node labelled `tier=frontend`.
2. Create a continuous rule selecting `matchLabels: {tier: frontend}` with taint `readiness.k8s.io/frontend-ready:NoSchedule` and a condition the Node does not satisfy. The controller applies the taint.
3. Relabel the Node so it no longer matches: `kubectl label node worker-1 tier=backend --overwrite`.
4. Delete the rule: `kubectl delete nodereadinessrule frontend-ready`.
5. The rule is gone: `kubectl get nodereadinessrule frontend-ready` returns `NotFound`.
6. The taint is still there: `kubectl get node worker-1 -o jsonpath='{.spec.taints}'` still lists `readiness.k8s.io/frontend-ready`.

I confirmed this with an envtest based regression test. The controller logs `Cleaning up taints for deleted rule`, removes the finalizer, and the rule object is deleted, while the taint stays on the Node:

```
INFO  Cleaning up taints for deleted rule  {"rule": "scratch-orphan-rule"}
[FAILED] rule deletion must not leave a permanently orphaned taint
```

Note there is no error and no warning event, so this is silent from an operator's point of view.

### Expected Behavior

Deleting a `NodeReadinessRule` should remove that rule's taint from every Node the rule managed, including Nodes that stopped matching its `nodeSelector` while the rule was alive. After the rule object is gone, no taint owned by it should remain in the cluster.

`main` (commit `43bff46`)

### Kubernetes Version

Not version specific. Reproduced against envtest with Kubernetes 1.36.2 binaries.

I searched the existing issues and PRs before filing and did not find this case covered. #235 and #342 are about the `NodeReconciler` path while the rule still exists, and neither mentions rule deletion. #247 concerns bootstrap annotations left behind on deletion rather than taints, and was resolved by moving to UID keyed annotations. None of the open PRs (#236, #237, #209, #339) touch `cleanupTaintsForRule` or `reconcileDelete`.

I would like to work on this and will open a PR shortly.

tejassinghbhati:
## Description

`cleanupTaintsForRule` only considered Nodes that currently match the rule's `spec.nodeSelector`:

```go
	if !r.ruleAppliesTo(ctx, rule, &node) {
		continue
	}
```

A Node that the rule had tainted, and that was later relabelled out of the selector, was skipped during cleanup. `reconcileDelete` then removed the finalizer and the rule object was deleted, leaving the taint behind with no owner. Nothing in the system references it any more, so no future reconcile can remove it and an operator has to clear it by hand. For a `NoSchedule` taint the Node stops accepting new workloads indefinitely; for `NoExecute` it also keeps evicting Pods without a toleration, on a Node that no rule governs.

Cleanup now also covers Nodes recorded in the rule's own status, through `status.nodeEvaluations` and `status.appliedNodes`.

Two design points worth calling out, since both came up while looking at the related issues:

**Why using status is acceptable here.** The review discussion on #339 concluded that leaning on `rule.Status` to decide what to act on is fragile, which is a large part of why #342 is still open. That objection is about steady state reconciliation, where the controller repeatedly consults status it wrote itself. The deletion path is different: it runs once, it is terminal, and the rule is being torn down, so the set of Nodes the rule ever managed is exactly the set that needs cleaning. There is no ongoing loop for a stale entry to feed back into.

**Why not just remove the taint everywhere.** The validating webhook rejects rules sharing a taint key and effect only when their selectors *overlap*, so two rules may legitimately share `readiness.k8s.io/x:NoSchedule` with disjoint selectors. Unconditionally stripping the key on deletion would pull the taint off Nodes owned by the surviving sibling rule. Scoping to this rule's status avoids that, because a Node managed solely by the sibling never appears in this rule's status. The second test added here locks that behaviour in.

This is deliberately scoped to the deletion path and does not attempt to resolve #342, which needs the harder "used to match, no longer does" detection during normal reconciliation. I filed this separately because the deletion case is strictly worse: while the rule still exists the taint is at least recoverable by a future fix, whereas after deletion it is permanent.

## Related Issue

Fixes #347

## Type of Change

/kind bug

## Testing

Two specs added to the existing `when a rule is deleted` context in `internal/controller/nodereadinessrule_controller_test.go`:

1. `should remove taints from nodes that no longer match the selector` reproduces the bug. The Node is managed by the rule, relabelled out of the selector, and the rule is deleted. The spec asserts the rule object is fully gone (so nothing could clean up later) and that the taint is not left behind.
2. `should not remove a taint from a node the rule never managed` guards the fix against over-removal, using a Node outside the selector that carries the same taint key and effect.

Verified the first test genuinely catches the regression by stashing the source change and re-running against unmodified `main`:

```
[FAILED] Timed out after 10.000s.
Rule deletion should not leave an orphaned taint on a node that stopped matching the selector
```

With the fix applied, the full controller suite passes: 64 of 64 specs, up from 62 before these two were added. `internal/webhook` and `cmd/readiness-condition-reporter` pass as well. Run against envtest with Kubernetes 1.36.2 binaries. `golangci-lint` v2.12.1 reports 0 issues.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fixed rule deletion leaving a permanently orphaned taint on nodes that had stopped matching the rule's nodeSelector while the rule was still active.
```

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -667,12 +667,34 @@ func (r *RuleReadinessController) processDryRun(ctx context.Context, rule *readi
 }
 
 // cleanupTaintsForRule removes taints managed by this rule from all applicable nodes.
+//
+// Nodes the rule previously managed are cleaned up as well, even if they no longer
+// match spec.nodeSelector. This runs on the terminal deletion path: once the
+// finalizer is removed the rule object is gone, so a taint left behind here can
+// never be reconciled away again and would strand the node.
+//
+// The previously managed set is taken from the rule's own status. That is sound
+// here precisely because this is a one shot cleanup of a rule that is going away,
+// unlike steady state reconciliation where status is not a trustworthy signal for
+// deciding the next action. It also keeps the cleanup correctly scoped: the
+// validating webhook only rejects rules sharing a taint key and effect when their
+// selectors overlap, so two rules may legitimately share a taint with disjoint
+// selectors. A node managed solely by such a sibling rule never appears in this
+// rule's status and is therefore left untouched.
 func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule *readinessv1alpha1.NodeReadinessRule, nodeList *corev1.NodeList) error {
 	log := ctrl.LoggerFrom(ctx)
 
+	managedNodes := make(map[string]bool, len(rule.Status.NodeEvaluations)+len(rule.Status.AppliedNodes))
+	for _, evaluation := range rule.Status.NodeEvaluations {
+		managedNodes[evaluation.NodeName] = true
+	}
+	for _, nodeName := range rule.Status.AppliedNodes {
+		managedNodes[nodeName] = true
+	}
+
 	var errors []string
 	for _, node := range nodeList.Items {
-		if !r.ruleAppliesTo(ctx, rule, &node) {
+		if !r.ruleAppliesTo(ctx, rule, &node) && !managedNodes[node.Name] {
 			continue
 		}
 
--- a/internal/controller/nodereadinessrule_controller_test.go
+++ b/internal/controller/nodereadinessrule_controller_test.go
@@ -1617,6 +1617,89 @@ var _ = Describe("NodeReadinessRule Controller", func() {
 				return err != nil && client.IgnoreNotFound(err) == nil
 			}, time.Second*10).Should(BeTrue(), "Rule should be fully deleted")
 		})
+
+		It("should remove taints from nodes that no longer match the selector", func() {
+			// Establish the node as managed by the rule so it is recorded in the
+			// rule status, which is what deletion cleanup uses to find nodes that
+			// have drifted out of the selector.
+			_, err := ruleReconciler.Reconcile(ctx, reconcile.Request{NamespacedName: types.NamespacedName{Name: "cleanup-rule"}})
+			Expect(err).NotTo(HaveOccurred())
+
+			Eventually(func() []nodereadinessiov1alpha1.NodeEvaluation {
+				managedRule := &nodereadinessiov1alpha1.NodeReadinessRule{}
+				_ = k8sClient.Get(ctx, types.NamespacedName{Name: "cleanup-rule"}, managedRule)
+				return managedRule.Status.NodeEvaluations
+			}, time.Second*10).Should(ContainElement(HaveField("NodeName", "cleanup-test-node")),
+				"Rule status should record the node it manages")
+
+			// The node is relabelled and stops matching spec.nodeSelector.
+			relabelled := &corev1.Node{}
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: "cleanup-test-node"}, relabelled)).To(Succeed())
+			relabelled.Labels["kubernetes.io/hostname"] = "some-other-hostname"
+			Expect(k8sClient.Update(ctx, relabelled)).To(Succeed())
+
+			// The taint applied by the rule is still on the node at this point.
+			Expect(k8sClient.Get(ctx, types.NamespacedName{Name: "cleanup-test-node"}, relabelled)).To(Succeed())
+			Expect(relabelled.Spec.Taints).To(ContainElement(HaveField("Key", "readiness.k8s.io/cleanup-taint")),
+				"Node should still carry the rule taint before deletion")
+
+			Expect(k8sClient.Delete(ctx, rule)).To(Succeed())
+
+			_, err = ruleReconciler.Reconcile(ctx, reconcile.Request{NamespacedName: types.NamespacedName{Name: "cleanup-rule"}})
+			Expect(err).NotTo(HaveOccurred())
+
+			// Once the finalizer is gone the rule cannot be reconciled again, so
+			// any taint still present at this point is stranded permanently.
+			Eventually(func() bool {
+				deletedRule := &nodereadinessiov1alpha1.NodeReadinessRule{}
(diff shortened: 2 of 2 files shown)


Karthik-K-N:
Lets discuss about the scope of the issue in corresponding issue.

tejassinghbhati:
Sure, that makes sense. I have written up the options and the trade offs over on #347 so the discussion stays in one place.

Happy to hold this until there is a direction, and equally happy to close it if we land on leaving the behaviour as it is and documenting it instead. Either way I will follow up with the docs change so the outcome is written down somewhere an operator will actually find it.

tejassinghbhati:
Closing this. Agreed on #347 that it is a duplicate of #235, so this should be redone against that issue once there is agreement on the approach.

The branch is still on my fork if it is worth reviving. The change scopes deletion cleanup to nodes recorded in the rule status, plus two envtest specs, one reproducing the stranded taint and one checking a node the rule never managed keeps its taint.

PR #351, opened 2026-08-04
fix: add events.k8s.io to the Helm chart manager ClusterRole

### LINKED ISSUE #350: [BUG] Helm chart RBAC is missing the events.k8s.io group, so taint events are never recorded
### What happened?

The Helm chart's manager ClusterRole does not grant permission on the `events.k8s.io` API group, so the controller cannot create the Node events it emits for taint operations. Every `TaintAdded`, `TaintRemoved` and `TaintAdopted` event is rejected by the API server when the controller is installed with Helm.

The kubebuilder markers on the controller declare both API groups ([nodereadinessrule_controller.go#L97-L98](https://github.com/kubernetes-sigs/node-readiness-controller/blob/021cd1d/internal/controller/nodereadinessrule_controller.go#L97-L98)):

```go
// +kubebuilder:rbac:groups="",resources=events,verbs=create;patch
// +kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;patch
```

and the generated `config/rbac/role.yaml` reflects that:

```yaml
- apiGroups:
  - ""
  - events.k8s.io
  resources:
  - events
  verbs:
  - create
  - patch
```

The chart's ClusterRole is maintained by hand and only carries the core group. Rendering it shows the gap:

```console
$ helm template nrr-controller charts/nrr-controller --show-only templates/rbac.yaml
...
  name: nrr-controller-manager-role
rules:
  - apiGroups: [""]
    resources: ["events"]
    verbs: ["create", "patch"]
  - apiGroups: [""]
    resources: ["nodes"]
    verbs: ["get", "list", "patch", "update", "watch"]
...
```

The core group entry is not sufficient, because the controller writes through the newer events API rather than the legacy one. `RuleReadinessController.EventRecorder` is populated from `mgr.GetEventRecorder(...)`, whose type is `k8s.io/client-go/tools/events.EventRecorder`. In controller-runtime v0.24.1 that recorder is backed by a broadcaster built in `pkg/manager/manager.go`:

```go
evtCl, err := eventsv1client.NewForConfigAndClient(config, httpClient)
...
return record.NewBroadcaster(), events.NewBroadcaster(&events.EventSinkImpl{Interface: evtCl}), true
```

`eventsv1client` is `k8s.io/client-go/kubernetes/typed/events/v1`, so the writes land on `events.k8s.io/v1` and are authorised against the `events.k8s.io` group. With the chart's role the API server returns a forbidden error instead, and the events never appear.

This only affects the Helm installation path. Installing with kustomize applies the generated `config/rbac/role.yaml`, which has the correct permissions, so the behaviour differs between the two documented install methods.

Worth noting that node events for taint operations were added deliberately in #134 and PR #158 and are described in the docs as the way to observe what the controller did to a node. On a Helm install that observability is silently missing, and the only hint is repeated forbidden errors in the controller log.

I think the underlying reason this slipped through is that nothing verifies the chart RBAC against the generated role. `hack/verify-chart-drift.sh` runs in the Helm workflow but only diffs the CRD:

```bash
diff -u \
  config/crd/bases/readiness.node.x-k8s.io_nodereadinessrules.yaml \
  charts/nrr-controller/crds/nodereadinessrules.readiness.node.x-k8s.io.yaml
```

so any divergence in the ClusterRole goes unnoticed. Comparing the two by hand, the events rule is currently the only difference; nodes, nodes/status and the three nodereadinessrules rules all match.

### Steps to Reproduce

1. Install the controller with Helm, with `rbac.create` left at its default of `true`.
2. Create a rule whose condition a node does not satisfy, so the controller applies a taint.
3. Confirm the taint was applied: `kubectl get node <node> -o jsonpath='{.spec.taints}'`.
4. Look for the corresponding event: `kubectl get events --field-selector involvedObject.name=<node>`. No `TaintAdded` event is recorded.
5. The controller log shows the write being refused, along the lines of `events.events.k8s.io is forbidden: User "system:serviceaccount:<ns>:nrr-controller" cannot create resource "events" in API group "events.k8s.io"`.

You can see the missing grant without a cluster:

```console
$ helm template nrr-controller charts/nrr-controller --show-only templates/rbac.yaml | grep -A2 'resources: \["events"\]'
  - apiGroups: [""]
    resources: ["events"]
    verbs: ["create", "patch"]
```

compared with `config/rbac/role.yaml`, which lists `""` and `events.k8s.io` together.

### Expected Behavior

The chart's manager ClusterRole should grant the same permissions the controller actually needs, matching the generated `config/rbac/role.yaml`, so taint events are recorded on a Helm install exactly as they are on a kustomize install.

`main` (commit `021cd1d`), chart `nrr-controller-0.1.0`, appVersion `v0.4.1`

### Kubernetes Version

Not version specific. The chart was rendered with Helm v3.15.1, the version pinned in the Helm workflow.

I searched the existing issues and PRs first and did not find this reported. Of the open PRs that touch the chart, #340 changes `templates/deployment.yaml` for KubeLinter and #315 changes the bundled CRD, so neither overlaps with the ClusterRole.

Happy to send a PR adding the missing group, along with a chart unit test so it cannot regress. Extending the drift check to cover RBAC as well as CRDs seems worthwhile too, but the generated role and the templated one differ in name and labels so a plain diff will not work, and that feels like a separate piece of work rather than something to fold into the fix.

tejassinghbhati:
## Description

The chart's manager ClusterRole granted events on the core API group only:

```yaml
  - apiGroups: [""]
    resources: ["events"]
    verbs: ["create", "patch"]
```

That is not enough for what the controller actually does. `RuleReadinessController.EventRecorder` comes from `mgr.GetEventRecorder(...)`, which is a `k8s.io/client-go/tools/events.EventRecorder`. In controller-runtime v0.24.1 that recorder is backed by a broadcaster constructed in `pkg/manager/manager.go`:

```go
evtCl, err := eventsv1client.NewForConfigAndClient(config, httpClient)
...
return record.NewBroadcaster(), events.NewBroadcaster(&events.EventSinkImpl{Interface: evtCl}), true
```

`eventsv1client` is `k8s.io/client-go/kubernetes/typed/events/v1`, so the writes go to `events.k8s.io/v1` and are authorised against the `events.k8s.io` group. Without that grant every `TaintAdded`, `TaintRemoved` and `TaintAdopted` write comes back forbidden, and since the recorder does not surface an error to the caller the only sign is a message in the controller log.

The kubebuilder markers on the controller already declare both groups, and the generated `config/rbac/role.yaml` carries both, so the kustomize install path was never affected. This was Helm only, which also means the two documented install methods behaved differently.

This change brings the chart in line with the generated role and adds a chart unit test suite for the manager ClusterRole, pinning the events grant along with the node and rule permissions.

One thing worth flagging for a follow up rather than fixing here: nothing currently guards this. `hack/verify-chart-drift.sh` runs in the Helm workflow but only diffs the CRD, so any divergence in the ClusterRole goes unnoticed. Extending it to cover RBAC seems worthwhile, but the generated role and the templated one differ in name and labels so a plain `diff` will not do, and it felt like the wrong thing to bundle into a fix. Comparing the two by hand, the events rule was the only difference; nodes, nodes/status and the three nodereadinessrules rules all matched already, and the new test now covers those too.

## Related Issue

Fixes #350

## Type of Change

/kind bug

## Testing

Added `charts/nrr-controller/tests/rbac_test.yaml`, which asserts the events rule lists both groups, asserts the remaining manager permissions match the generated role, and checks that nothing is rendered when `rbac.create` is disabled.

Confirmed the new test genuinely catches the regression by stashing the template change and re-running against the unmodified chart:

```
 FAIL  Test Node Readiness Controller Manager RBAC	charts/nrr-controller/tests/rbac_test.yaml
	- grants events on both the core and events.k8s.io groups
Tests:       1 failed, 18 passed, 19 total
```

With the fix applied, the whole suite passes, using Helm v3.15.1 and helm-unittest 1.0.3 to match the Helm workflow:

```
$ helm unittest charts/nrr-controller --strict
 PASS  Test Node Readiness Controller Pod Annotations and Labels
 PASS  Test Node Readiness Controller Deployment
 PASS  Test Node Readiness Controller NodeReadinessRules
 PASS  Test Node Readiness Controller Manager RBAC
 PASS  Test Node Readiness Controller Webhook and Metrics Config

Charts:      1 passed, 1 total
Test Suites: 5 passed, 5 total
Tests:       19 passed, 19 total
```

`helm lint charts/nrr-controller` passes, and the rendered role now reads:

```yaml
  - apiGroups: ["", "events.k8s.io"]
    resources: ["events"]
    verbs: ["create", "patch"]
```

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fixed the Helm chart's manager ClusterRole missing the events.k8s.io API group, which prevented the controller from recording TaintAdded, TaintRemoved and TaintAdopted events on nodes when installed with Helm.
```

### DIFF
--- a/charts/nrr-controller/templates/rbac.yaml
+++ b/charts/nrr-controller/templates/rbac.yaml
@@ -37,7 +37,10 @@ metadata:
   labels:
     {{- include "nrr-controller.labels" . | nindent 4 }}
 rules:
-  - apiGroups: [""]
+  # The controller records taint events through k8s.io/client-go/tools/events,
+  # which writes to events.k8s.io/v1, so both groups are required here. Keep this
+  # in sync with the generated config/rbac/role.yaml.
+  - apiGroups: ["", "events.k8s.io"]
     resources: ["events"]
     verbs: ["create", "patch"]
   - apiGroups: [""]
--- a/charts/nrr-controller/templates/rbac.yaml
+++ b/charts/nrr-controller/templates/rbac.yaml
@@ -37,9 +37,6 @@ metadata:
   labels:
     {{- include "nrr-controller.labels" . | nindent 4 }}
 rules:
-  # The controller records taint events through k8s.io/client-go/tools/events,
-  # which writes to events.k8s.io/v1, so both groups are required here. Keep this
-  # in sync with the generated config/rbac/role.yaml.
   - apiGroups: ["", "events.k8s.io"]
     resources: ["events"]
     verbs: ["create", "patch"]
--- a/charts/nrr-controller/tests/rbac_test.yaml
+++ b/charts/nrr-controller/tests/rbac_test.yaml
@@ -1,0 +1,72 @@
+suite: Test Node Readiness Controller Manager RBAC
+
+templates:
+  - "*.yaml"
+
+release:
+  name: nrr-controller
+
+tests:
+  - it: grants events on both the core and events.k8s.io groups
+    template: templates/rbac.yaml
+    documentSelector:
+      path: metadata.name
+      value: nrr-controller-manager-role
+    asserts:
+      - isKind:
+          of: ClusterRole
+      # The controller records taint events via k8s.io/client-go/tools/events,
+      # which writes to events.k8s.io/v1. Granting only the core group makes
+      # every TaintAdded, TaintRemoved and TaintAdopted write forbidden.
+      - contains:
+          path: rules
+          content:
+            apiGroups: ["", "events.k8s.io"]
+            resources: ["events"]
+            verbs: ["create", "patch"]
+
+  - it: grants the node and rule permissions the controller needs
+    template: templates/rbac.yaml
+    documentSelector:
+      path: metadata.name
+      value: nrr-controller-manager-role
+    asserts:
+      - contains:
+          path: rules
+          content:
+            apiGroups: [""]
+            resources: ["nodes"]
+            verbs: ["get", "list", "patch", "update", "watch"]
+      - contains:
+          path: rules
+          content:
+            apiGroups: [""]
+            resources: ["nodes/status"]
+            verbs: ["get"]
+      - contains:
+          path: rules
+          content:
+            apiGroups: ["readiness.node.x-k8s.io"]
+            resources: ["nodereadinessrules"]
+            verbs: ["get", "list", "patch", "update", "watch"]
+      - contains:
+          path: rules
+          content:
+            apiGroups: ["readiness.node.x-k8s.io"]
+            resources: ["nodereadinessrules/finalizers"]
+            verbs: ["update"]
+      - contains:
+          path: rules
+          content:
+            apiGroups: ["readiness.node.x-k8s.io"]
+            resources: ["nodereadinessrules/status"]
+            verbs: ["get", "patch", "update"]
+
+  - it: omits rbac resources when rbac.create is disabled
+    template: templates/rbac.yaml
+    set:
+      rbac:
+        create: false
+    asserts:
+      - hasDocuments:
+          count: 0


charts/nrr-controller/templates/rbac.yaml
```diff
@@ -37,7 +37,10 @@ metadata:
   labels:
     {{- include "nrr-controller.labels" . | nindent 4 }}
 rules:
-  - apiGroups: [""]
+  # The controller records taint events through k8s.io/client-go/tools/events,
+  # which writes to events.k8s.io/v1, so both groups are required here. Keep this
+  # in sync with the generated config/rbac/role.yaml.
+  - apiGroups: ["", "events.k8s.io"]
```
ajaysundark, line 40: @tejassinghbhati Thanks for this patch!

I recently migrated NRC to the new events group in controller-runtime 0.24 and handled this for our rbac:
https://github.com/kubernetes-sigs/node-readiness-controller/blob/021cd1d4cb4f571d7917a19228ece5394153738e/config/rbac/role.yaml#L25-L26

Looks like the charts (https://github.com/kubernetes-sigs/node-readiness-controller/pull/163) that was merged after this missed this update.

@tejassinghbhati Could you help review whether there are any other mismatch in resources / RBACs between charts/* and config/*, so we could handle it together?

cc @honghainguyen777 
We have https://github.com/kubernetes-sigs/node-readiness-controller/blob/021cd1d4cb4f571d7917a19228ece5394153738e/hack/verify-chart-drift.sh#L1 that handles the diff for crds, is there room to expand this for other resources such as RBAC?

charts/nrr-controller/templates/rbac.yaml
```diff
@@ -37,7 +37,10 @@ metadata:
   labels:
     {{- include "nrr-controller.labels" . | nindent 4 }}
 rules:
-  - apiGroups: [""]
+  # The controller records taint events through k8s.io/client-go/tools/events,
```
ajaysundark: nit: Thanks for adding context, but I think it's safe to omit this comment here as it is self-explanatory.

tejassinghbhati:
Comment dropped in 88f2ea1, agreed it was noise.

I went through the rest of it. Short answer is that the events grant was the only functional mismatch, everything else in RBAC lines up.

I compared each role in `config/rbac/` against its counterpart in the chart. `leader_election_role` (coordination.k8s.io/leases), `metrics_auth_role` (tokenreviews and subjectaccessreviews), `metrics_reader` (the `/metrics` nonResourceURL), and the admin, editor and viewer roles all match verb for verb. In `role.yaml` itself, nodes, nodes/status and the three nodereadinessrules rules were already correct, so events was the one that drifted. The new test in this PR pins all of them, not just events, so the rest cannot quietly drift either.

Outside RBAC I checked the webhook and the manager. The ValidatingWebhookConfiguration matches on rules, operations, failurePolicy, sideEffects, admissionReviewVersions and path. The webhook service is 443 to 9443 in both. The webhook manager patch and the metrics TLS patch match the chart on cert mount paths, secret names and the webhook containerPort. Nothing broken there.

Two differences that look deliberate rather than wrong, flagging them so you can confirm. The chart has no NetworkPolicy equivalent for `config/network-policy/allow-metrics-traffic.yaml`, though that is opt-in on the kustomize side too since `config/default` only pulls in `../rbac` and `../manager`. And the metrics service defaults differ, 8080 plain with a separate TLS overlay in kustomize versus 8443 with a `metrics.secure` toggle in the chart.

On expanding the drift check, yes, I think there is room. A plain `diff` will not work because the generated role and the templated one differ in name and labels, so it needs to render both sides and compare the rules semantically rather than textually. I will open an issue with a concrete proposal and a PR so there is something real to look at rather than just an idea.

ajaysundark:
Thanks for taking care of this!

/lgtm
/approve

PR #352, opened 2026-08-04
feat: add suspend field to NodeReadinessRuleSpec to pause rule evalua…

### LINKED ISSUE #335: [FEATURE] Add a suspend field to `NodeReadinessRuleSpec` to pause rule evaluation

Currently, there is no way to temporarily disable or pause the evaluation of a `NodeReadinessRule`. If an administrator needs to perform maintenance on a cluster, test a change, or debug an issue without the controller constantly evaluating and applying/removing taints, their only option is to delete the rule entirely or switch it to dry-run mode (which still runs the evaluations and updates the status).

I would like a `suspend: true` boolean field added to the `NodeReadinessRuleSpec`. 
When `suspend` is true, the controller should immediately skip the reconciliation loop for that rule, making no changes to the cluster or the rule's status. The rule essentially becomes frozen in its current state.

YAML example:
```yaml
kind: NodeReadinessRule
metadata:
  name: example-rule
spec:
  suspend: true # <--- New field
  enforcementMode: continuous
  taint:
    key: readiness.k8s.io/not-ready
    effect: NoSchedule
  nodeSelector:
    matchLabels:
      kubernetes.io/os: linux
  conditions:
    - type: Ready
      requiredStatus: "True"


1. **Using `dryRun: true`:** This doesn't actually stop the controller from evaluating nodes, it just prevents applying the taints. It still consumes controller resources to compute the dry-run results.
2. **Deleting the rule:** Deleting the rule works, but it causes the controller to remove all taints it applied, which alters the cluster state during maintenance. It also requires the user to back up and restore the YAML to pause operations.

vishnukothakapu:
## Description
This PR introduces a `suspend` boolean field to the `NodeReadinessRuleSpec`. When a rule is suspended (`suspend: true`), the controller skips the reconciliation loop for that specific rule, leaving any existing taints and the cluster state exactly as they are. This provides a safe, standard mechanism to pause rule evaluation during maintenance windows or debugging, without needing to delete the rule.

## Related Issue
Fixes #335

## Type of Change

/kind feature
/kind api-change

## Testing
- Added a unit test in `nodereadinessrule_controller_test.go` to explicitly verify that when `suspend: true` is set, the controller skips evaluation and does not append a finalizer.
- Regenerated CRD manifests and deepcopy functions.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Added `suspend` field to `NodeReadinessRuleSpec`. When `suspend: true` is set, the controller skips evaluation of the rule, effectively pausing taint management while preserving the current cluster state.

### DIFF
--- a/api/v1alpha1/nodereadinessrule_types.go
+++ b/api/v1alpha1/nodereadinessrule_types.go
@@ -104,6 +104,11 @@ type NodeReadinessRuleSpec struct {
 	//
 	// +optional
 	DryRun bool `json:"dryRun,omitempty"` //nolint:kubeapilinter
+
+	// suspend tells the controller to suspend operations for this NodeReadinessRule.
+	//
+	// +optional
+	Suspend bool `json:"suspend,omitempty"`
 }
 
 // ConditionRequirement defines a specific Node condition and the status value
@@ -327,6 +332,7 @@ type DryRunResults struct {
 // +kubebuilder:printcolumn:name="Effect",type=string,JSONPath=`.spec.taint.effect`,description="The taint effect: NoSchedule, PreferNoSchedule or NoExecute."
 // +kubebuilder:printcolumn:name="DryRun",type=boolean,JSONPath=`.spec.dryRun`,description="Whether the rule is in dry-run mode and only previews taint changes."
 // +kubebuilder:selectablefield:JSONPath=`.spec.dryRun`
+// +kubebuilder:printcolumn:name="Suspend",type=boolean,JSONPath=`.spec.suspend`,description="Whether operations for this rule are suspended."
 // +kubebuilder:printcolumn:name="Age",type=date,JSONPath=`.metadata.creationTimestamp`,description="The age of this resource"
 
 // NodeReadinessRule is the Schema for the NodeReadinessRules API.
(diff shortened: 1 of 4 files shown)


Karthik-K-N:
Thanks for the PR,  Given its a feature request and API change,I would recommend its always better to get consciousness about the proposed feature on the parent issue before proceeding with the PR, it generally saves duplicate work or time.

vishnukothakapu:
> Contributor

Thanks for the tip, @Karthik-K-N! You're right. I had some free time, so I thought I'd quickly prototype it to get a feel for how it would look. Going forward, I'll wait for consensus on the parent issue before spending time on implementation, so I don't end up working on something that might not be needed.

Once the parent issue is reviewed, I'd really appreciate any feedback on the implementation as well. Thanks!

Karthik-K-N:
> > Contributor
> 
> Thanks for the tip, @Karthik-K-N! You're right. I had some free time, so I thought I'd quickly prototype it to get a feel for how it would look. Going forward, I'll wait for consensus on the parent issue before spending time on implementation, so I don't end up working on something that might not be needed.
> 
> Once the parent issue is reviewed, I'd really appreciate any feedback on the implementation as well. Thanks!

That is completely fair. Apologies for that, I thought the PR is ready for review, May be if you think its under experiment, feel free to mark it as Draft or WIP.

PR #353, opened 2026-08-04
Replace metric label literals with constants

rawadhossain:
## Summary

This PR replaces metric label string literals with typed constants to keep metric label values consistent and avoid accidental changes.

### Changes
- Replace hardcoded metric label values with typed constants.
- Update controllers and tests to use the new constants.
- Standardize failure reasons, operations, and node state labels.

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -21,6 +21,40 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
 )
 
+// FailureReason represents the reason for a failed operation.
+type FailureReason string
+
+const (
+	FailureReasonEvaluationError  FailureReason = "EvaluationError"
+	FailureReasonAddTaintError    FailureReason = "AddTaintError"
+	FailureReasonRemoveTaintError FailureReason = "RemoveTaintError"
+)
+
+// TaintOperation represents a taint operation.
+type TaintOperation string
+
+const (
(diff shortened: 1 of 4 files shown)


internal/metrics/metrics.go
```diff
@@ -21,6 +21,36 @@ import (
...
+
+const (
+	OperationRemoveTaint ReconciliationOperation = "remove_taint"
+	OperationAddTaint    ReconciliationOperation = "add_taint"
+)
+
+type NodeState string
+
+const (
+	StateReady         NodeState = "ready"
```
Karthik-K-N: when comes to defining custom type and constant, we usually follow convention of constant names prefixed with their Type name, for example

```
type NodeState string

const (
    NodeStateReady         NodeState = "ready"
    NodeStateNotReady      NodeState = "not_ready"
    NodeStateBootstrapping NodeState = "bootstrapping"
)
```

can we follow the same approach?

rawadhossain: yes, we can, thanks. Updating now to prefix all the constants with their type name.

Karthik-K-N (commented):
LGTM

Given its constant, comments are generally encouraged, ideally linter might flag this.

/assign @ajaysundark

rawadhossain:
> LGTM
> 
> Given its constant, comments are generally encouraged, ideally linter might flag this.
> 
> /assign @ajaysundark

Thanks. Added the comments.

PR #364, opened 2026-08-06
docs: add Cilium CNI integration example and documentation (#121)

### LINKED ISSUE #121: Interaction between Cilium bootstrap taint and Node Readiness Controller
While integrating Node Readiness Controller (NRC) with Cilium-managed clusters, we observed potential overlap in bootstrap taint management.

**Background:**

When running Node Readiness Controller (NRC) in clusters that use Cilium as the CNI, Cilium currently manages a bootstrap [taint](https://docs.cilium.io/en/stable/installation/taints/) via a [flag](https://docs.cilium.io/en/stable/cmdref/cilium-operator/):
`node.cilium.io/agent-not-ready:NoSchedule`

**Behavior:**
- Cilium adds this taint during node initialization.
- It removes the taint once the Cilium agent is ready.
- It also sets the built-in Kubernetes node condition:
`NetworkUnavailable=False`

**Configuration Option:**

The taint key is [configurable](https://docs.cilium.io/en/stable/cmdref/cilium-agent/) via: `--agent-not-ready-taint-key` 
This allows overriding the default key (e.g. to readiness.k8s.io/network-not-ready).

Image

**Potential Overlap:**

If users configure NRC to enforce network readiness via taints and Cilium’s bootstrap taint management remains enabled, then both components may attempt to manage taints for the same network readiness domain.

One possible approach to resolve this would be:
If users choose to manage this taint with NRC, they could disable Cilium’s taint management component so that NRC becomes the sole manager of the network readiness taint.

**Motivation:**

We are working toward standardizing readiness taints under a fixed prefix (e.g. readiness.k8s.io/*) so that:
- Readiness taints can be categorized consistently.
- Autoscaling components (e.g. Cluster Autoscaler) can treat readiness taints differently from permanent taints.
- We avoid per-vendor allow-listing in autoscaling configurations.
- UX improves when integrating readiness enforcement across components.

Since Cilium already allows overriding the taint key via `--agent-not-ready-taint-key` it appears technically possible to align the bootstrap taint with a standardized readiness prefix without requiring changes to Cilium itself.

Dasmat13:
Fixes #121

### Summary
Adds documentation and an example manifest for integrating **Cilium CNI** with the Node Readiness Controller.

### Key Details
- **Taint Key Alignment:** Details configuring Cilium's `--agent-not-ready-taint-key` to match `readiness.k8s.io/network-not-ready`.
- **NodeReadinessRule:** Demonstrates using `NetworkUnavailable=False` for bootstrap network readiness.
- **Continuous Health:** Explains differences between bootstrap `NetworkUnavailable` status and continuous runtime health probes.

### Changes
- `examples/cni-readiness/cilium-network-readiness-rule.yaml`: Manifest for Cilium network readiness rule.
- `docs/book/src/examples/cni-readiness.md`: Added Cilium CNI integration guide.
- `examples/cni-readiness/README.md`: Added Cilium note.

### DIFF
--- a/docs/book/src/examples/cni-readiness.md
+++ b/docs/book/src/examples/cni-readiness.md
@@ -144,3 +144,35 @@ To test this, add a new node to the cluster.
 
 3.  **Check Taint Removal**:
     As soon as the condition becomes `True`, the Node Readiness Controller will remove the taint, and workloads will be scheduled.
+
+## Cilium CNI Integration & Bootstrap Taints
+
+When using **Cilium** as your CNI plugin:
+
+1. **Align Bootstrap Taints**: Configure Cilium's agent not-ready taint flag to use the standardized `readiness.k8s.io/*` prefix:
+   ```yaml
+   # Helm values for Cilium
+   extraArgs:
+     - --agent-not-ready-taint-key=readiness.k8s.io/network-not-ready
+   ```
+
+2. **Bootstrap NodeReadinessRule**: Once Cilium initializes networking, it sets the built-in Kubernetes node condition `NetworkUnavailable=False`. You can define a rule to untaint nodes based on `NetworkUnavailable`:
+   ```yaml
+   # cilium-network-readiness-rule.yaml
+   apiVersion: readiness.node.x-k8s.io/v1alpha1
+   kind: NodeReadinessRule
+   metadata:
+     name: cilium-network-readiness-rule
+   spec:
+     conditions:
+       - type: "NetworkUnavailable"
+         requiredStatus: "False"
+     taint:
+       key: "readiness.k8s.io/network-not-ready"
+       effect: "NoSchedule"
+       value: "true"
+     enforcementMode: "bootstrap-only"
+   ```
+
+> [!NOTE]
+> Cilium updates `NetworkUnavailable=False` during node bootstrap. For continuous post-startup health monitoring, pair this with a custom DaemonSet probe and `enforcementMode: "continuous"`.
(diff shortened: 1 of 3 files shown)


Dasmat13:
Hi @haircommander @tallclair, quick ping on this PR. Could you take a look and `/ok-to-test` when you get a chance? Thanks

ajaysundark:
cc @arnab-logs

 could you take a look?

arnab-logs:
Thanks for picking this up @Dasmat13 I went through yor PR thoroughly had a few thoughts 

The issue's suggested resolution was to pick a single owner for the readiness taint either NRC or Cilium, not both, since two independent controllers managing the same taint risks an overlap

This PR renames Cilium's taint key to the prefix `--agent-not-ready-taint-key=readiness.k8s.io/network-not-ready` & adds an NRC `NodeReadinessRule` targeting that same key via `NetworkUnavailable`. But asper Cilium's own docs, `--agent-not-ready-taint-key` only changes the name of the taint Cilium manages; Cilium's agent still adds and removes it itself independent of NRC

So in the common case Cilium will remove the taint on its own before NRC's rule could do anything i.e. nrc rule ends up watching a taint someone else already cleared. 

Also, I found 2 documented cases where trusting its own internal readiness signal alone caused problems (and where NRC could be of help)
- https://github.com/cilium/cilium/issues/10762
- https://github.com/cilium/cilium/issues/29405

I think a way to work with it would be either:

1. Make cilium as authoritative i.e. keep the renamed flag (with this we get the standardized prefix for autoscaler tooling which was also the core motivation in #121) and drop the `NodeReadinessRule` for this specific taint since it is nit doing anything Cilium isn't already doing by itself
2. Give NRC the authority to remove taints i.e. we don't touch Cilium's flag at all, leave it at its default `node.cilium.io/agent-not-ready`. Instead the docs needs to explain how to get the node to start already tainted with `readiness.k8s.io/network-not-ready` (e.g. via `--register-with-taints` at node creation) and then let the `NodeReadinessRule` from this PR remove that taint based on `NetworkUnavailable`.

cc: @ajaysundark  since this touches a bit of a design decision, wanted to get your read on which direction makes sense here

PR #373, opened 2026-08-08
docs: remove undocumented gracePeriod, fix taint comment

Daniel-1600:
## Description

`Graceperiod` is documented in the documentation but never used in any controller logic using it, so it is currently documentation-only and misleading for users.
## Checklist
- [ ] `make test` passes
- [ ] `make lint` passes

```release-note

```
Doc #(issue)

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -127,9 +127,6 @@ type NodeReadinessRuleSpec struct {
     // Node selector for targeting specific nodes
     NodeSelector *metav1.LabelSelector `json:"nodeSelector,omitempty"`
     
-    // Grace period before applying taint changes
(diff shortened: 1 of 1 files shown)


AvineshTripathi:
@ajaysundark do you think we need GracePeriod feature?

Also @Daniel-1600 I see more reference in the same file

Daniel-1600:
Ooh yeah I saw the other reference thanks , @AvineshTripathi should i proceed to remove all the other references or wait for @ajaysundark to confirm if we still really need the feature ?

ajaysundark:
> do you think we need GracePeriod feature?

These can be cleaned up; @Daniel-1600  thanks for taking care of this!

PR #379, opened 2026-08-09
Add Prometheus metrics to readiness condition reporter

### LINKED ISSUE #380: [Feature] Reporter metrics
Emit metrics from readiness reporter

rawadhossain:
### Description
Implements the Reporter metrics specified in the [observability design](https://github.com/ajaysundark/node-readiness-controller/blob/9244f4eef28234321fecc988435b846fd3c12507/docs/observability-design.md)

- `node_readiness_reporter_build_info`
- `node_readiness_reporter_check_duration_seconds`
- `node_readiness_reporter_checks_total`
- `node_readiness_reporter_condition_writes_total`

Also adds the Prometheus `/metrics` and `/healthz` endpoints to the readiness condition reporter.

### Implementation choices
- **Single HTTP server** for both `/metrics` and `/healthz` to keep the Reporter to one listener and port.
- **Dedicated Prometheus registry** to keep the Reporter's metrics isolated and avoid registration conflicts in tests, following the same general pattern used by the controller metrics.
- Port **`:9445` as the default metrics port** because the Reporter runs with `hostNetwork: true`. The address is configurable through `METRICS_BIND_ADDRESS`.

### Build fix
Updated `Dockerfile.reporter` to build the Reporter package instead of single file `main.go`, since the new `metrics.go` also needs to be included. 

Fixes #380

## Type of Change
/kind feature

## Testing
- Verified metrics scraping with a live `kind` cluster and Prometheus.
- Verified graceful SIGTERM shutdown.
- Verified the Reporter continues its normal check/condition-update loop when the metrics port is unavailable.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes


cmd/readiness-condition-reporter/main.go
```diff
@@ -161,8 +202,18 @@ func runCheck(ctx context.Context, httpClient *http.Client, clientset kubernetes
 		}
 	}
 
+	switch health.Reason {
+	case "EndpointOK":
```
AvineshTripathi: lets make these const

cmd/readiness-condition-reporter/main.go
```diff
@@ -142,6 +176,11 @@ func main() {
 		select {
 		case <-ctx.Done():
 			klog.InfoS("Shutting down readiness condition reporter", "reason", ctx.Err())
+			shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), metricsShutdownTimeout)
+			if err := metricsServer.Shutdown(shutdownCtx); err != nil {
+				klog.ErrorS(err, "Failed to gracefully shut down metrics server")
+			}
+			shutdownCancel()
```
AvineshTripathi: maybe this in defer before Shutdown func call

cmd/readiness-condition-reporter/metrics.go
```diff
@@ -0,0 +1,74 @@
...
+)
+
+func init() {
+	// Register custom metrics with the dedicated reporter registry.
+	registry.MustRegister(reporterBuildInfo)
+	registry.MustRegister(reporterCheckDuration)
+	registry.MustRegister(reporterChecksTotal)
+	registry.MustRegister(reporterConditionWritesTotal)
+
+	reporterBuildInfo.WithLabelValues(info.GetVersionParts()[0]).Set(1)
```
AvineshTripathi: while `GetVersionParts` func currently returns a non empty array. I think we still should add a length check for `info.GetVersionParts()[0]`

cmd/readiness-condition-reporter/metrics.go
```diff
@@ -0,0 +1,74 @@
...
+			Name:    "node_readiness_reporter_check_duration_seconds",
+			Help:    "Duration of health probe checks.",
+			Buckets: prometheus.DefBuckets,
+		},
+	)
+
+	// reporterChecksTotal tracks total probe check results over time.
+	reporterChecksTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "node_readiness_reporter_checks_total",
```
AvineshTripathi, line 53: `node_readiness_reporter_checks` instead

cmd/readiness-condition-reporter/main.go
```diff
@@ -127,11 +145,27 @@ func main() {
 		Timeout: defaultHTTPTimeout,
 	}
 
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.HandlerFor(registry, promhttp.HandlerOpts{}))
+	mux.HandleFunc("/healthz", func(w http.ResponseWriter, _ *http.Request) {
+		w.WriteHeader(http.StatusOK)
+	})
+	metricsServer := &http.Server{
```
AvineshTripathi, line 155: why not using "sigs.k8s.io/controller-runtime/pkg/metrics/server"? We already have it as part of go.mod

AvineshTripathi (commented):
Thank you for starting this PR! I have left some comments

cmd/readiness-condition-reporter/metrics.go
AvineshTripathi: but I see in #344 it says `node_readiness_reporter_checks_total `

cmd/readiness-condition-reporter/main.go
rawadhossain: Yeah, I looked into this one. It can run standalone without a full Manager setup, but I kept the current approach for a few reasons:
- We use a dedicated registry for the Reporter, so the `/metrics` endpoint only exposes the metrics we explicitly register.
- The current implementation gives us explicit control over the metrics server shutdown and keeps the existing klog setup.
- Using `metrics/server` doesn't really make the implementation simpler once the registry and logging setup are included.

I also tested the current setup with a real SIGTERM against a live pod, and it shuts down cleanly in ~0.3s. So I kept the current implementation since it gives us the behavior we wanted and is already validated. Im okay to switch to `metrics/server` if you think that's preferable here.

cmd/readiness-condition-reporter/main.go
rawadhossain: Done. Moved the cancellation to defer

cmd/readiness-condition-reporter/main.go
rawadhossain: yes, added constants now

cmd/readiness-condition-reporter/metrics.go
rawadhossain: yeah, that'd be better. Added a length check with unknown as the fallback when no version is available.

cmd/readiness-condition-reporter/metrics.go
rawadhossain: Yeah, was confused about this too, so kept it as in the design doc.

ajaysundark: I followed https://prometheus.io/docs/practices/naming/#metric-names. The metrics convention suggests that `an accumulating count has total as a suffix`. Also on a similar note, https://www.robustperception.io/on-the-naming-of-things/ was a good read on why our `node_readiness_rules_total` (a gauge) metric could drop the suffix as can cause to confusions like [this](https://github.com/canonical/grafana-k8s-operator/issues/316#issuecomment-2038765819).

cmd/readiness-condition-reporter/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+)
+
+// registry is a dedicated Prometheus registry for the reporter.
+var registry = prometheus.NewRegistry()
+
+var (
+	// reporterBuildInfo tracks the reporter binary version to track fleet version skew.
+	reporterBuildInfo = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "node_readiness_reporter_build_info",
```
ajaysundark, line 32: @rawadhossain just reviewed #406. Don't we need the ldFlags on this one?

rawadhossain: Yeah, we need it here too. Was planning to get #406 merged first and then follow up here. I’ll push the changes once #406 lands.

rawadhossain: @ajaysundark done with the idFlags changes now that #406 is merged. PTAL.

AvineshTripathi:
/lgtm

I think this is all good now. @ajaysundark can you take a final look

cmd/readiness-condition-reporter/metrics.go
```diff
@@ -0,0 +1,75 @@
...
+		},
+		func() float64 { return 1 },
+	)
+
+	// reporterCheckDuration tracks the duration of health probe checks.
+	reporterCheckDuration = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "node_readiness_reporter_check_duration_seconds",
+			Help:    "Duration of health probe checks.",
+			Buckets: prometheus.DefBuckets,
```
ajaysundark: An AI review caught this as excessive, so I looked it up - https://pkg.go.dev/github.com/prometheus/client_golang/prometheus#pkg-variables. The default bucket seems to be 11 buckets from 0.005 granularity to 10s.

I wonder whether these many milliseconds accuracy is needed. May be we could keep it fewer to something like {0.1, 0.5, 1, 5, 10 }?

cmd/readiness-condition-reporter/metrics.go
```diff
@@ -0,0 +1,75 @@
...
+			Help: "Total probe check results over time.",
+		},
+		[]string{"result"}, // result: healthy, unhealthy, error
+	)
+
+	// reporterConditionWritesTotal tracks how the reporter handles Node condition updates.
+	reporterConditionWritesTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "node_readiness_reporter_condition_writes_total",
+			Help: "Total Node condition updates, including skipped writes when the condition is unchanged.",
```
ajaysundark:
```suggestion
			Help: "Total Node condition updates handled by the reporter.",
```

cmd/readiness-condition-reporter/metrics.go
```diff
@@ -0,0 +1,75 @@
...
+
+// registry is a dedicated Prometheus registry for the reporter.
+var registry = prometheus.NewRegistry()
+
+var (
+	// reporterBuildInfo tracks the reporter binary version to track fleet version skew.
+	reporterBuildInfo = prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Name: "node_readiness_reporter_build_info",
+			Help: "Reporter binary version to track fleet version skew.",
```
ajaysundark:
```suggestion
			Help: "Reporter binary version.",
```

ajaysundark:
/lgtm

please address the comments when you get time before we could merge this.

cmd/readiness-condition-reporter/metrics.go
rawadhossain: Looked into this and I agree that the default buckets are more than we need here. Tested it by running the reporter against real traffic + some deliberate slow responses to see which buckets actually mattered.

So I ended up with `{0.005, 0.1, 0.25, 0.5, 1, 2.5, 5, 10}`. Kept a few more buckets since they were catching useful delays in testing.

Committed the changes. PTAL.

AvineshTripathi:
/lgtm

Adding it again! Changes looks good

PR #381, opened 2026-08-09
Clear FailedNodes on successful node reconciliation

### LINKED ISSUE #376: [BUG] NodeReconciler permanently leaks transient errors into Status.FailedNodes
### What happened?

When `NodeReconciler` encounters an error evaluating a node, it records it in `rule.Status.FailedNodes`. If the error resolves on the next reconcile and evaluation succeeds, the controller fails to clear the old failure because it lacks an `else` block to handle success. 

The stale failure is read from the local cache and actively merged back into the API server, permanently trapping the healthy node in the `FailedNodes` array. It only clears if a manual rule edit forces the `RuleReconciler` to run.

### Steps to Reproduce

1. Trigger a transient error during node evaluation
2. Observe the node is added to `rule.Status.FailedNodes`.
3. Resolve the error. The controller retries and evaluation succeeds.
4. Check the rule status. The node operates normally but is permanently stuck in `FailedNodes`.

### Expected Behavior

`processNodeAgainstAllRules` must include an `else` block to explicitly clear the node from `rule.Status.FailedNodes` upon successful evaluation, mirroring the correct logic found in `processAllNodesForRule`.

main

### Kubernetes Version

Client Version v1.30.0

### Controller Logs

_No response_

### Additional Details

This issue is entirely distinct from PR #222 and PR #237.
Those PRs fixed the retry mechanism by ensuring that transient errors are properly returned to the reconciler so controller runtime can requeue them.

While those PRs ensured the controller retries on failure, they did not fix the controller forgetting to clean up Status.FailedNodes upon success.

Ayush4958:
## Description
This fixes a state leak in the NodeReconciler where transient evaluation errors permanently leaked into `rule.Status.FailedNodes`. 

When a transient error resolves and the node successfully evaluates on the next retry, the controller was forgetting to clear the old failure from the cached rule object. As a result, the status patch actively re-merged the stale error back into the API server

This fix adds the missing `else` block to `processNodeAgainstAllRules` to clear the failure upon success, matching the correct logic already present in the RuleReconciler's `processAllNodesForRule`. It also adds a unit test to prevent future regressions.

## Related Issue
Fixes #376 

## Type of Change
/kind bug

## Testing
Added a regression test in `node_controller_test.go` (`"should clear FailedNodes entry when node evaluation succeeds after a prior transient failure"`) which simulates a transient failure followed by a successful evaluation, asserting that `FailedNodes` drops to empty.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fixed a bug where transient errors during node reconciliation caused nodes to become permanently stuck in the `FailedNodes` status of a `NodeReadinessRule`.

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -161,16 +161,7 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			metrics.Failures.WithLabelValues(rule.Name, string(metrics.FailureReasonEvaluationError)).Inc()
 		} else {
 			// Clear any stale failures from previous reconciliation attempts.
-			// Without this, transient errors permanently leak into Status.FailedNodes
-			// because the cached rule object still carries the old failure entry,
-			// which gets re merged into the API server on every subsequent patch.
-			var updatedFailedNodes []readinessv1alpha1.NodeFailure
-			for _, failure := range rule.Status.FailedNodes {
-				if failure.NodeName != node.Name {
-					updatedFailedNodes = append(updatedFailedNodes, failure)
-				}
-			}
-			rule.Status.FailedNodes = updatedFailedNodes
+			r.clearNodeFailure(rule, node.Name)
 		}
 
 		// Persist the rule status
@@ -434,6 +425,17 @@ func (r *RuleReadinessController) recordNodeFailure(
 		LastEvaluationTime: metav1.Now(),
 	})
 
+	rule.Status.FailedNodes = failedNodes
+}
+
+// clearNodeFailure removes any failure record for a specific node from the rule status.
+func (r *RuleReadinessController) clearNodeFailure(rule *readinessv1alpha1.NodeReadinessRule, nodeName string) {
+	var failedNodes []readinessv1alpha1.NodeFailure
+	for _, failure := range rule.Status.FailedNodes {
+		if failure.NodeName != nodeName {
+			failedNodes = append(failedNodes, failure)
+		}
+	}
 	rule.Status.FailedNodes = failedNodes
 }
 
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -159,6 +159,18 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
 			errs = append(errs, err)
 			metrics.Failures.WithLabelValues(rule.Name, string(metrics.FailureReasonEvaluationError)).Inc()
+		} else {
+			// Clear any stale failures from previous reconciliation attempts.
+			// Without this, transient errors permanently leak into Status.FailedNodes
+			// because the cached rule object still carries the old failure entry,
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Ayush4958** [comment]
@ajaysundark @Karthik-K-N 
PR is ready for review pls take time to review it 

thanks ...

**Comment by rawadhossain** [comment]
@Ayush4958 I think the Fixes 379 reference is incorrect here. It seems unrelated to the PR you raised.

**Comment by Ayush4958** [comment]
> @Ayush4958 I think the Fixes 379 reference is incorrect here. It seems unrelated to the PR you raised.

yh, my bad, pressed wrong number key

**Comment by DsThakurRawat** [comment]
Ran this branch against the repro from #376. The fix in `node_controller.go` works, and `latestRule` persisted to the fake client clears `FailedNodes` as expected.

One quick heads-up on the unit test in `node_controller_test.go`: `go test ./internal/...` trips on line 1214:

```
FailedNodes must be cleared after successful evaluation
Expected <[]v1alpha1.NodeFailure | len:1>: [{NodeName: "recovery-node", ...}] to be empty
```

That happens because `getApplicableRulesForNode` returns `rule.DeepCopy()`. `processNodeAgainstAllRules` mutates the copy and persists it to the API server, so the in-memory `rule` pointer from test setup never gets modified in place.

The API server assertion at line 1220 (`latestRule.Status.FailedNodes` being empty) already proves the fix works. Dropping the assertion on `rule` at line 1214 (or asserting on `latestRule` only) should get the suite green when prow tests run.

**Comment by Ayush4958** [comment]
thanks for it and u were right about memory pointer and 
I had dropped that assertion and It passes the test

**Comment by ajaysundark** [comment]
Thanks for the PR!

Seems duplicate of https://github.com/kubernetes-sigs/node-readiness-controller/pull/209 and https://github.com/kubernetes-sigs/node-readiness-controller/pull/333. I'll check the status of either of them first and see if can be picked up.

File: internal/controller/node_controller.go [hunk 1]
```diff
@@ -159,6 +159,18 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
 			errs = append(errs, err)
 			metrics.Failures.WithLabelValues(rule.Name, string(metrics.FailureReasonEvaluationError)).Inc()
+		} else {
+			// Clear any stale failures from previous reconciliation attempts.
```
ajaysundark, line 163: Can you extract clearNodeFailure as a helper instead? ref #333.

ajaysundark (changes requested):

internal/controller/node_controller.go
ajaysundark: It'd be good to give credits to previous authors with "Co-authored-by"

Ayush4958: > Can you extract clearNodeFailure as a helper instead? ref #333.

Done! I had extracted clearNodeFailure into a helper & also included the co-authored-by

ajaysundark:
Thanks for the patch!

/lgtm
/approve

PR #384, opened 2026-08-10
fix: stop reporting adoption for taints the controller applied itself

### LINKED ISSUE #383: [BUG] TaintAdopted events are emitted repeatedly for taints the controller applied itself
### What happened?

The controller keeps emitting `TaintAdopted` events for taints it applied itself. Any node that is still unready and already carries the rule's taint gets a fresh `TaintAdopted` event, and an `Adopting pre-existing taint` log line, on every single reconcile.

`NodeReconciler` gets its rules from the in-memory cache. `getApplicableRulesForNode` returns a deep copy of the cached rule, and that copy carries whatever `status` the rule had the last time `RuleReconciler` ran. Node evaluations persisted since then are not in it. Since the rule spec is immutable, `GenerationChangedPredicate` means `RuleReconciler` rarely fires again, so that snapshot can stay stale for the life of the process.

`evaluateRuleForNode` decides adoption from that status:

```go
isFirstEvaluation := r.getPreviousNodeEvaluation(rule, node.Name) == nil
```

For a node the snapshot never contained, this is always true, so the adoption branch fires every time:

```go
case !shouldRemoveTaint && currentlyHasTaint:
    if isFirstEvaluation {
        log.Info("Adopting pre-existing taint", ...)
        r.EventRecorder.Eventf(node, nil, corev1.EventTypeNormal, "TaintAdopted", "AdoptTaint", "%s", message)
    }
```

`TaintAdopted` is supposed to tell an operator that a taint was already on the node and the rule has taken ownership of it, which is what #134 and #158 added it for. Firing it repeatedly for the controller's own taints makes it useless as a signal, since you cannot tell a real adoption from noise. It also adds pointless writes to the events API for every unready node on every condition change.

Note this is not the same as the rule status being wrong. The persisted `status.nodeEvaluations` is correct, it is only the cached copy the evaluation reads from that is behind.

### Steps to Reproduce

1. Create a continuous rule with taint `readiness.k8s.io/example:NoSchedule` and a condition the node does not satisfy.
2. Let the controller apply the taint.
3. Leave the node unready and let it reconcile a few more times, for example by touching an unrelated condition or label.
4. `kubectl get events --field-selector involvedObject.name=<node>` shows a `TaintAdopted` event for each pass, even though the controller applied that taint in step 2.

An envtest spec shows it directly. After one reconcile applies the taint, three more reconciles produce three adoption events:

```
INFO  Adopting pre-existing taint  {"node": "adopt-events-node", "rule": "adopt-events-rule", ...}
INFO  Adopting pre-existing taint  {"node": "adopt-events-node", "rule": "adopt-events-rule", ...}
INFO  Adopting pre-existing taint  {"node": "adopt-events-node", "rule": "adopt-events-rule", ...}
[FAILED] the rule already owns this taint, so no further TaintAdopted events should be emitted
```

### Expected Behavior

`TaintAdopted` should be emitted once, and only when the taint was on the node before the rule evaluated it. A taint the controller applied itself is not an adoption.

`main` (commit `d74dabd`)

### Kubernetes Version

Not version specific. Reproduced on envtest with Kubernetes 1.36.2 binaries.

I checked the open issues and PRs before filing. #362 is also about events but the opposite gap, no Warning events on failures, so it does not overlap. Of the open PRs that touch this area, #294, #315, #343 and #209 all read `getPreviousNodeEvaluation` or `isFirstEvaluation` but none refresh the cached status, so none of them fix this. #294 would inherit the same staleness for the new metric it adds.

I have a fix and tests ready and will open a PR.

tejassinghbhati:
## Description

`NodeReconciler` evaluates rules from the in-memory cache, and the cached copy carries whatever status the rule had the last time `RuleReconciler` ran. Node evaluations persisted since then are not in it, and because the rule spec is immutable `GenerationChangedPredicate` means that reconcile rarely fires again, so the snapshot can stay stale for the life of the process.

`evaluateRuleForNode` decides adoption from that status, so for a node the snapshot never contained `isFirstEvaluation` is always true and the adoption branch fires on every pass. The result is a `TaintAdopted` event and an `Adopting pre-existing taint` log line every time an unready, already tainted node reconciles, for a taint the controller applied itself.

The fix refreshes `NodeEvaluations` from the live object before evaluating. That read goes through the controller-runtime cache, so it is not an extra API call, and the persist block a few lines below already does the same `Get`. It deliberately touches only `NodeEvaluations` and leaves `FailedNodes` alone, so it does not overlap with #381 or #209.

## Related Issue

Fixes #383

## Type of Change

/kind bug

## Testing

Two specs added under `when a node is re-evaluated while still unready`:

The first applies the taint, confirms the evaluation is persisted, then reconciles three more times and asserts no further `TaintAdopted` events. It fails on `main`:

```
INFO  Adopting pre-existing taint  {"node": "adopt-events-node", ...}
INFO  Adopting pre-existing taint  {"node": "adopt-events-node", ...}
INFO  Adopting pre-existing taint  {"node": "adopt-events-node", ...}
[FAILED] the rule already owns this taint, so no further TaintAdopted events should be emitted
```

The second puts the taint on the node before the rule ever evaluates it and asserts exactly one `TaintAdopted` event, so the feature still works. That one passes both with and without the fix, which is what shows the change is narrow rather than just switching adoption off.

Full controller suite passes, 64 of 64 specs, up from 62 before these two. `internal/webhook` and `cmd/readiness-condition-reporter` pass too. Run against envtest with Kubernetes 1.36.2 binaries. `golangci-lint` v2.12.1 reports 0 issues.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Fixed repeated TaintAdopted events being recorded for taints the controller applied itself. The event is now emitted only when a taint was already on the node before the rule took ownership of it.
```

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -152,6 +152,21 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			"rule", rule.Name,
 			"ruleResourceVersion", rule.ResourceVersion)
 
+		// The cached rule carries a status snapshot taken the last time
+		// RuleReconciler ran. Node evaluations persisted since then are missing
+		// from it, and because the rule spec is immutable that reconcile rarely
+		// runs again, so the snapshot can stay stale for the lifetime of the
+		// process. Refresh the evaluations from the live object first, otherwise
+		// every pass looks like the node's first evaluation. This reads through
+		// the controller-runtime cache, so it is not an extra API call.
+		latestRule := &readinessv1alpha1.NodeReadinessRule{}
+		if err := r.Get(ctx, client.ObjectKey{Name: rule.Name}, latestRule); err != nil {
+			log.V(4).Info("Could not refresh rule status before evaluation, using cached copy",
+				"node", node.Name, "rule", rule.Name, "error", err.Error())
+		} else {
+			rule.Status.NodeEvaluations = latestRule.Status.NodeEvaluations
+		}
+
 		if err := r.evaluateRuleForNode(ctx, rule, node); err != nil {
 			log.Error(err, "Failed to evaluate rule for node",
 				"node", node.Name, "rule", rule.Name)
--- a/internal/controller/node_controller_test.go
+++ b/internal/controller/node_controller_test.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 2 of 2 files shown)


ajaysundark:
Thanks for the PR @tejassinghbhati.
 
K8s events are by design best-effort, I dont think this is required to refresh rule cache (which can be a large object) when a node adoption need to be verified.

tejassinghbhati:
Fair enough. The Get deep copies the whole rule status, so on a big rule that cost per node is not worth it just to quiet down an event. Thanks for looking.

PR #388, opened 2026-08-11
docs: add Helm installation guide

### LINKED ISSUE #387: [Helm] Doc update for installing with helm-chart

Update project documentation on helm chart installation flow

capture how to install with helm-charts

Tasks:
1. Update the installation.md with examples
2. Capture #438 guidelines once implemented
3. Update stale [charts/node-readiness-controller/README.md](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/charts/node-readiness-controller/README.md)

_No response_

tejassinghbhati:
## Description

The chart has been in the repo since #163 but Helm is not mentioned anywhere in the book. The install page only covered release manifests, kustomize and static pods, so there was no documented Helm path at all.

Adds a Helm section to the installation guide covering install from a checkout, since OCI chart releases are still WIP, which values turn on metrics, TLS and the validating webhook, and how to ship rules through `nodeReadinessRules`.

It also documents two things that are specific to Helm and easy to get caught by:

Helm installs the CRD from `crds/` on first install only and never touches it again, so a chart bump that changes the schema needs the CRD applied by hand or rules using new fields get rejected while the controller supports them. This is in the chart README already but not where someone following the install guide would see it.

Rules declared in `nodeReadinessRules` are release resources, so `helm uninstall` deletes them together with the controller. The taint finalizer then has no controller left to run it, which is the stuck-resource case already documented further down the same page.

## Related Issue

Fixes #387

## Type of Change

/kind documentation

## Testing

Docs only, no code change. I checked every value and default against the chart rather than writing from memory. `leaderElection.enabled` true, `metrics.enabled` and `metrics.secure` false, `webhook.enabled` and `validatingWebhook.enabled` false, `certManager.enabled` false. The `ValidatingWebhookConfiguration` is gated on `and .Values.webhook.enabled .Values.validatingWebhook.enabled`, which is why the guide says both are needed.

I also rendered the exact commands in the guide. The default install produces the controller, RBAC and service account only. The full one produces what the table promises:

```
      2 kind: Certificate
      6 kind: ClusterRole
      2 kind: ClusterRoleBinding
      1 kind: Deployment
      1 kind: Issuer
      1 kind: Role
      1 kind: RoleBinding
      2 kind: Service
      1 kind: ServiceAccount
      1 kind: ValidatingWebhookConfiguration
```

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
NONE
```

### DIFF
--- a/docs/book/src/user-guide/installation.md
+++ b/docs/book/src/user-guide/installation.md
@@ -110,6 +110,21 @@ helm install nrr-controller ./charts/nrr-controller \
 
 Both `webhook.enabled` and `validatingWebhook.enabled` are needed. The first runs the webhook server in the controller and mounts its certificate, the second registers the `ValidatingWebhookConfiguration` with the API server.
 
+#### Tuning for larger clusters
+
+The values under `controller` map to the manager's own flags, and each one is only passed to the container when you move it off its default:
+
+| Value | Flag | Default |
+| :--- | :--- | :--- |
+| `controller.nodeConcurrentReconciles` | `--node-concurrent-reconciles` | `1` |
+| `controller.ruleConcurrentReconciles` | `--rule-concurrent-reconciles` | `1` |
+| `controller.kubeAPIQPS` | `--kube-api-qps` | `-1`, client-side throttling off |
+| `controller.kubeAPIBurst` | `--kube-api-burst` | `-1`, client-side throttling off |
+| `controller.enableNodeStateMetrics` | `--enable-node-state-metrics` | `false` |
+| `controller.pprofBindAddress` | `--pprof-bind-address` | unset, disabled |
+
+Raising the two concurrency values is the usual response to readiness taints lagging behind node joins on a large cluster. `enableNodeStateMetrics` adds the per-rule `node_readiness_nodes_by_state` gauge at the cost of extra API reads on node updates, so turn it on when you want the fleet view and can afford the reads.
+
 #### Managing rules through the chart
 
 `NodeReadinessRule` objects can be shipped with the release through the `nodeReadinessRules` value:
--- a/docs/book/src/user-guide/installation.md
+++ b/docs/book/src/user-guide/installation.md
@@ -64,13 +64,13 @@ docker pull $REPO:$TAG
 ```
 ### Option 2: Helm Chart
 
-The chart lives in the repository under `charts/nrr-controller`. Published chart releases via `registry.k8s.io` OCI are still work in progress, so install it from a checkout for now.
+The chart lives in the repository under `charts/node-readiness-controller`. Published chart releases via `registry.k8s.io` OCI are still work in progress, so install it from a checkout for now.
 
 ```sh
 git clone https://github.com/kubernetes-sigs/node-readiness-controller.git
 cd node-readiness-controller
 
-helm install nrr-controller ./charts/nrr-controller \
+helm install node-readiness-controller ./charts/node-readiness-controller \
   --namespace nrr-system --create-namespace
 ```
 
@@ -79,9 +79,9 @@ Requires Helm 3.x. This deploys the controller with the same defaults as the sta
 For anything beyond a couple of overrides, keep your settings in a file instead of a long `--set` list:
 
 ```sh
-helm show values ./charts/nrr-controller > custom-values.yaml
+helm show values ./charts/node-readiness-controller > custom-values.yaml
 
-helm install nrr-controller ./charts/nrr-controller \
+helm install node-readiness-controller ./charts/node-readiness-controller \
   --namespace nrr-system --create-namespace \
   -f custom-values.yaml
 ```
@@ -99,7 +99,7 @@ Everything beyond the core controller is opt-in, matching the kustomize componen
 The webhook rejects rules whose taint key and effect collide with an existing rule over an overlapping node selector, so it is worth enabling in production.
 
 ```sh
-helm install nrr-controller ./charts/nrr-controller \
+helm install node-readiness-controller ./charts/node-readiness-controller \
   --namespace nrr-system --create-namespace \
   --set certManager.enabled=true \
   --set webhook.enabled=true \
@@ -157,7 +157,7 @@ Pull the version of the chart you want and upgrade the release in place. Values 
 ```sh
 git pull
 
-helm upgrade nrr-controller ./charts/nrr-controller \
+helm upgrade node-readiness-controller ./charts/node-readiness-controller \
   --namespace nrr-system \
   -f custom-values.yaml
 ```
@@ -167,7 +167,7 @@ helm upgrade nrr-controller ./charts/nrr-controller \
 Check what changed before applying it to a live cluster:
 
 ```sh
-helm diff upgrade nrr-controller ./charts/nrr-controller --namespace nrr-system   # needs the helm-diff plugin
+helm diff upgrade node-readiness-controller ./charts/node-readiness-controller --namespace nrr-system   # needs the helm-diff plugin
 ```
 
 Read the CRD note below first. Helm will not update the CRD for you, so a chart bump that changes the schema needs that step done by hand.
@@ -179,7 +179,7 @@ Helm installs the CRD from the chart's `crds/` directory on first install only. 
 Before moving to a chart version that changes the `NodeReadinessRule` schema, apply the CRD yourself:
 
 ```sh
-kubectl apply -f charts/nrr-controller/crds/nodereadinessrules.readiness.node.x-k8s.io.yaml
+kubectl apply -f charts/node-readiness-controller/crds/nodereadinessrules.readiness.node.x-k8s.io.yaml
 ```
 
 Skipping this leaves the old schema in place, and rules using newly added fields are rejected by the API server even though the controller supports them.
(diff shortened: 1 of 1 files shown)


Ayush4958 (commented):
It's great documentation but i got some suggestion for it :- 

- we should add helm upgrade section after initial installing it can be great for use 
- mention that users can supply their own overrides file using `-f custom-values.yaml`

tejassinghbhati:
Good calls, added both in 9fde2a2.

There is a values file section now with `helm show values` to get a starting point, and an upgrade section covering `helm upgrade`, `--install`, and `helm diff` if you have the plugin. Pointed it at the CRD note too since Helm skips the CRD on upgrade.

tejassinghbhati:
Rebased on main and added a section for the controller tuning values that landed in #392, since they were not in the chart when I opened this.

Checked all six flags exist in cmd/main.go and rendered them to confirm they only get passed when moved off their defaults. Chart tests are at 31 now, all green.

tejassinghbhati:
Rebased for the chart rename in #407, so the install, upgrade and uninstall commands and the CRD path all use node-readiness-controller now. Also added a section for the tuning values from #392 since those landed after I opened this.

One extra thing in here that predates the PR, flagging it so it does not look random. The Verification section looked for pods with `-l component=node-readiness-controller`, and nothing sets a component label, not the chart and not the kustomize manifests. So that command matched nothing on either install path. Both set `control-plane=controller-manager` so I switched to that. It is right below the new Helm section, so someone following this guide hits it straight after installing.

Happy to pull that into its own PR if you would rather keep this one to Helm.

tejassinghbhati:
Backed out the verification selector change in f3057ce, and I got it wrong twice over.

#401 was already filed and assigned for this two days before I touched it, so it was not mine to fix. And my analysis was wrong anyway. I said nothing sets a component label. examples/static-pod/node-readiness-controller.yaml sets it, and it is the only place in the repo that does. That pod has component and tier and no control-plane label, so switching the selector to control-plane=controller-manager would have broken verification for static pod installs, the one path where the command currently works.

So this PR is back to just the Helm section plus the rename and tuning updates. Sorry for the noise.

docs/book/src/user-guide/installation.md
```diff
@@ -62,7 +62,129 @@ REPO="registry.k8s.io/node-readiness-controller/node-readiness-controller"
 TAG=$(skopeo list-tags docker://$REPO | jq .'Tags[-1]' | tr -d '"')
 docker pull $REPO:$TAG
 ```
-### Option 2: Advanced Deployment (Kustomize)
+### Option 2: Helm Chart
+
+The chart lives in the repository under `charts/node-readiness-controller`. Published chart releases via `registry.k8s.io` OCI are still work in progress, so install it from a checkout for now.
```
ajaysundark: This is now available for 0.5.0. Can you validate these steps and update the PR with changes?

tejassinghbhati: Updated in the latest commit to use the official 
egistry.k8s.io OCI registry for chart install, values inspection, and upgrades (with a note kept for local checkout development).

tejassinghbhati: Validated against the published chart and updated in 0c54cbf.

It pulls fine, appVersion v0.5.0, so the install commands are right as written.

Three things came out of actually running them. The chart version was in a variable called VERSION, which Option 1 already uses for the release tag, and they are different numbers, so that is now CHART_VERSION. The CRD upgrade step reused that same VERSION set way back in Option 1, so anyone arriving via Helm never had it, it now derives the release from the chart appVersion.

The last one is worth a look beyond this PR. Installing from a local checkout deploys v0.4.1 right now, because build-helm injects version and app-version at package time and Chart.yaml keeps a placeholder. Same at the v0.5.0 tag. I have added --set image.tag to that snippet so the docs are not misleading, but the underlying thing might be worth fixing properly.

docs/book/src/user-guide/installation.md
```diff
@@ -62,7 +62,156 @@ REPO="registry.k8s.io/node-readiness-controller/node-readiness-controller"
 TAG=$(skopeo list-tags docker://$REPO | jq .'Tags[-1]' | tr -d '"')
 docker pull $REPO:$TAG
 ```
-### Option 2: Advanced Deployment (Kustomize)
+### Option 2: Helm Chart
+
+The official Helm chart is published to the OCI registry at `registry.k8s.io/node-readiness-controller/charts/node-readiness-controller`.
+
+```sh
+# Chart version, which is independent of the controller release tag above.
```
ajaysundark, line 70: @vitorfloriano is this because the chart cannot carry the "vX.Y.Z" pattern?

Could you help check if there's room for unifying / fixing this? Having it different would cause discovery issues as we have to keep documenting the mapping in our documentation somewhere.

docs/book/src/user-guide/installation.md
```diff
@@ -62,7 +62,156 @@ REPO="registry.k8s.io/node-readiness-controller/node-readiness-controller"
...
+Requires Helm 3.8+ (native OCI support). This deploys the controller with the same defaults as the standard manifest: leader election on, metrics off, and the validating webhook off.
+
+> [!NOTE]
+> You can also install the chart from a local checkout when developing:
+> ```sh
+> helm install node-readiness-controller ./charts/node-readiness-controller \
+>   --namespace nrr-system --create-namespace \
+>   --set image.tag=v0.5.0
+> ```
+>
```
ajaysundark: We don't need to document this option. You can remove

docs/book/src/user-guide/installation.md
tejassinghbhati: Not a Helm limitation, I checked. Helm takes v0.5.0 as a chart version fine, it packages, lints and renders, and helm show chart reads it back as v0.5.0.

The v gets dropped by one line in the Makefile:

```make
CHART_VERSION ?= $(shell echo "$(RELEASE_VERSION)" | sed 's/^v//')
```

That came in with #402, so the split is a choice rather than a constraint. Unifying is deleting that sed and setting Chart.yaml to v0.5.0. ct is not in the way either, check-version-increment is already false.

The case for leaving it stripped is that semver itself has no v and most published charts use the plain number, so anything parsing chart versions strictly expects 0.5.0. Helm is lenient, other consumers may not be.

Happy to do it if you want it. It touches the Makefile, Chart.yaml and the version check I added in #453, and it would remove the CHART_VERSION vs VERSION split from these docs, which is the discovery problem you are describing.

docs/book/src/user-guide/installation.md
```diff
@@ -62,7 +62,156 @@ REPO="registry.k8s.io/node-readiness-controller/node-readiness-controller"
...
+> Set `image.tag` explicitly here. The real chart and app versions are injected at
```
ajaysundark: I don't think the user installation documention need to capture the make commands used to create the charts

docs/book/src/user-guide/installation.md
```diff
@@ -62,7 +62,156 @@ REPO="registry.k8s.io/node-readiness-controller/node-readiness-controller"
...
+> package time by `make build-helm`, so `Chart.yaml` in the repository keeps a
+> placeholder `appVersion`. Installing straight from a checkout without an override
+> therefore deploys whatever that placeholder points at rather than the release you
+> checked out.
+
+For anything beyond a couple of overrides, keep your settings in a file instead of a long `--set` list:
```
ajaysundark: We should document the overrides possible, and show a sample how to use it in helm command. We don't need to be opinionated here about whether the user wants to use set or inject values via yaml

docs/book/src/user-guide/installation.md
```diff
@@ -62,7 +62,156 @@ REPO="registry.k8s.io/node-readiness-controller/node-readiness-controller"
...
+      effect: NoSchedule
+    nodeSelector:
+      matchLabels:
+        kubernetes.io/os: linux
+```
+
+`nodeSelector` is required on every entry. Set it explicitly, since an empty selector matches every node in the cluster.
+
+> [!NOTE]
+> With the validating webhook enabled, apply rules only once the controller is serving admission requests. On a first install the webhook is not ready while the rules in the same release are being created, so install the controller first and add the rules in a follow-up `helm upgrade`.
```
ajaysundark: This feels like a hack. Are we better off not (documenting) supporting the rules array in helm chart for now. we could do a second look at the UX.

docs/book/src/user-guide/installation.md
tejassinghbhati: Removed in 9351d16.

docs/book/src/user-guide/installation.md
tejassinghbhati: Gone with the local checkout note in 9351d16.

docs/book/src/user-guide/installation.md
tejassinghbhati: Reworked in 9351d16. It now points at helm show values for the full list and shows both --set and -f without picking a side.

docs/book/src/user-guide/installation.md
tejassinghbhati: Agreed, dropped the whole section in 9351d16, along with the uninstall caution that depended on it. The chart still supports the value, it is just not documented until the UX gets a second look.

ajaysundark:
/lgtm
/approve

Thanks for your contributions and keeping this live, @tejassinghbhati .

Issue #382, opened 2026-08-09
[BUG] bootstrap-only: grandfathering marks bootstrap completed concurrently with taint addition, permanently orphaning the taint


patrickshan:
### What happened?

Rolling out the controller (v0.4.1) plus a condition-reporting DaemonSet
(node-problem-detector with custom conditions) **simultaneously** to a cluster of
long-lived, healthy nodes left one node with a permanently orphaned
`readiness.k8s.io/newlyjoinednode:NoSchedule` taint.

Two controllers raced on the same node within the same second, on the controller's
first-ever reconcile sweep:

1. The **rule controller** evaluated the node before the reporter DaemonSet had
   stamped any of the rule's conditions, so every condition evaluated as
   `observed: "Unknown"` vs `required: "False"` → `allConditionsSatisfied: false`
   → it **added the taint**.
2. The **node controller** concurrently ran the grandfathering path for the
   pre-existing node and **marked bootstrap completed**
   (`readiness.k8s.io/bootstrap-completed-<rule-uid>` annotation written).
3. Every subsequent reconcile of that node now logs
   `Skipping bootstrap-only rule - already completed`, which short-circuits the
   entire rule — **including taint removal**.

End state (still true hours later): node is healthy, all rule conditions read
`False` (satisfied), the bootstrap-completed annotation is present, **and** the
rule's `NoSchedule` taint is still on the node. Nothing will ever remove it; it
had to be removed manually.

Other nodes tainted in the same initial sweep were cleaned up correctly — they
won the race (taint removed before the completion mark landed). Only the
interleaving order differed.

### Steps to Reproduce

1. Cluster with existing `Ready` worker nodes that do **not** yet carry the
   rule's condition types (reporter not yet deployed).
2. Deploy, at effectively the same time, a bootstrap-only `NodeReadinessRule`
   and the DaemonSet that reports its conditions:

   ```yaml
   apiVersion: readiness.node.x-k8s.io/v1alpha1
   kind: NodeReadinessRule
   metadata:
     name: newly-joined-node-rule
   spec:
     enforcementMode: bootstrap-only
     conditions:
     - type: CustomCheckA
       requiredStatus: "False"
     - type: CustomCheckB
       requiredStatus: "False"
     nodeSelector:
       matchExpressions:
       - key: node-role.kubernetes.io/control-plane
         operator: DoesNotExist
     taint:
       key: readiness.k8s.io/newlyjoinednode
       value: "true"
[...]

PR #389, opened 2026-08-11
Changes to handle Node taint mutation only through the node controller

Karthik-K-N:
## Description

Its is just a PoC

Currently we have NodeReadinessRule(NRR) controller and Node controller, which both acts on the node, Though the system is expected to be eventual consistent two controllers acting on Node may cause some issues to enduser.

With this PoC changes are made such that only the Node controller is authorative to mutate Node object that is adding/removing taint, where the the NRR controller takes care of building cache, managing finalizers and cleaning up.

## Checklist
- [ ] `make test` passes
- [ ] `make lint` passes

```release-note

```
Doc #(issue)

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -31,7 +31,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	readinessv1alpha1 "sigs.k8s.io/node-readiness-controller/api/v1alpha1"
 	"sigs.k8s.io/node-readiness-controller/internal/metrics"
@@ -46,6 +48,12 @@ type NodeReconciler struct {
 }
 
 // SetupWithManager sets up the controller with the Manager.
+//
+// The node controller serves as the sole writer for taint mutations. It is
+// triggered by two event sources:
+//
+//  1. Node changes (conditions / taints / labels).
+//  2. NodeReadinessRule changes.
 func (r *NodeReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) error {
 	concurrency := max(r.MaxConcurrentReconciles, 1)
 	return ctrl.NewControllerManagedBy(mgr).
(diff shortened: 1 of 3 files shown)


DsThakurRawat:
had a look at this since it came out of #382, and the single writer split does address the race there. with taint mutation living in the node controller there's no second writer to lose the annotation update against.

few things came up while testing the branch locally.

**observedGeneration now marks "seen" rather than "applied".** the rule reconciler sets `rule.Status.ObservedGeneration = rule.Generation` and patches it before the node reconciler has touched anything, so `observedGeneration == generation` can be true while every matching node is still untainted. on main it was set after the evaluation loop had run. probably fine for a PoC, but it's an API visible change, and anything polling that field as "this rule is applied" would go green early.

**appliedNodes doesn't shrink when a node is deleted.** the full recompute used to live in `processAllNodesForRule`, which loses its only caller here, so `Status.AppliedNodes` is now only maintained incrementally in `processNodeAgainstAllRules` and nothing prunes it. a deleted node keeps its entry. though you mentioned on #382 that NRE should own per node rule results, so if appliedNodes is on its way out with that work then this probably isn't worth fixing here.

**the delete path still writes taints from the rule reconciler.** `reconcileDelete` calls `cleanupTaintsForRule`, which goes straight to `removeTaintBySpec`. so the sole writer invariant holds everywhere except rule deletion. not sure whether that exception is deliberate for now.

also `processAllNodesForRule` is unused outside tests after this.

happy to share the envtest cases if they're useful.

ajaysundark:
I see it could be cleaner and would avoid patch fixes like #343. But I'm skeptical of multiple major refactors.. Concurrent update risks we are seeing are mainly node status handling (taint management is optimistically locked), so I want to wait for #345, and revisit this if we still want to go down this route. Wdyt?

Karthik-K-N:
> I see it could be cleaner and would avoid patch fixes like #343. But I'm skeptical of multiple major refactors.. Concurrent update risks we are seeing are mainly node status handling (taint management is optimistically locked), so I want to wait for #345, and revisit this if we still want to go down this route. Wdyt?

Sure, we can do that.

Karthik-K-N:
> had a look at this since it came out of #382, and the single writer split does address the race there. with taint mutation living in the node controller there's no second writer to lose the annotation update against.
> 
> few things came up while testing the branch locally.
> 
> **observedGeneration now marks "seen" rather than "applied".** the rule reconciler sets `rule.Status.ObservedGeneration = rule.Generation` and patches it before the node reconciler has touched anything, so `observedGeneration == generation` can be true while every matching node is still untainted. on main it was set after the evaluation loop had run. probably fine for a PoC, but it's an API visible change, and anything polling that field as "this rule is applied" would go green early.
> 
> **appliedNodes doesn't shrink when a node is deleted.** the full recompute used to live in `processAllNodesForRule`, which loses its only caller here, so `Status.AppliedNodes` is now only maintained incrementally in `processNodeAgainstAllRules` and nothing prunes it. a deleted node keeps its entry. though you mentioned on #382 that NRE should own per node rule results, so if appliedNodes is on its way out with that work then this probably isn't worth fixing here.
> 
> **the delete path still writes taints from the rule reconciler.** `reconcileDelete` calls `cleanupTaintsForRule`, which goes straight to `removeTaintBySpec`. so the sole writer invariant holds everywhere except rule deletion. not sure whether that exception is deliberate for now.
> 
> also `processAllNodesForRule` is unused outside tests after this.
> 
> happy to share the envtest cases if they're useful.

Hey, Thanks for looking at this. As its just a PoC so I haven't fixed most of the things, but based on the feedback will work in future. Thanks

PR #394, opened 2026-08-12
docs: add Helm chart installation flow to the installation guide

### LINKED ISSUE #387: [Helm] Doc update for installing with helm-chart

Ujjwal-Gowda:
## Description

The Helm chart added in #163 was not covered by the user guide. The installation page only documented the release manifests, Kustomize and static pod paths, so there was no documented way to install the chart.

This adds a Helm section alongside the existing options:

- Minimal install and full install (metrics + validating webhook).
- Upgrades.
- The CRD lifecycle caveat: Helm installs the CRD from the chart's `crds/` directory on first install only, and never upgrades or deletes it, so CRD changes across chart versions have to be applied by hand.

Two smaller fixes in the same page:

- Verification: the chart labels its pods with `app.kubernetes.io/name=nrr-controller`, so the `component=node-readiness-controller` selector documented for the manifest install returns nothing for a Helm install. Noted, with the correct selector.
- Uninstallation: added the `helm uninstall` path, including the fact that the CRD is left behind.

## Related Issue

Fixes #387

## Type of Change

/kind documentation

## Testing

Docs-only change, but every command in the new section was run against a kind cluster rather than written from the chart source:

- Fresh `helm install` creates the CRD and the controller rolls out.
- The full-install values render the secure metrics flags and the `ValidatingWebhookConfiguration`.
- The label selector in Verification returns the controller pod for a Helm install, and the previously documented one does not.
- The CRD survives `helm uninstall`, as described.

## Checklist
- [ x] `make test` passes
- [x ] `make lint` passes

No Go code is touched, so neither target exercises this change; leaving them unticked rather than claiming a run I have not completed locally. CI will cover them.

```release-note
NONE
```

### DIFF
--- a/docs/book/src/user-guide/installation.md
+++ b/docs/book/src/user-guide/installation.md
@@ -62,7 +62,56 @@ REPO="registry.k8s.io/node-readiness-controller/node-readiness-controller"
 TAG=$(skopeo list-tags docker://$REPO | jq .'Tags[-1]' | tr -d '"')
 docker pull $REPO:$TAG
 ```
-### Option 2: Advanced Deployment (Kustomize)
+
+### Option 2: Helm Chart
+
+A Helm chart is maintained in-tree at [`charts/nrr-controller`](https://github.com/kubernetes-sigs/node-readiness-controller/tree/main/charts/nrr-controller). Use it when you want the optional components (metrics, webhook, cert-manager integration) toggled through values rather than by composing Kustomize overlays.
+
+> [!NOTE]
+> The chart is not published to an OCI registry yet, so it must be installed from a clone of the repository. See the [chart README](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/charts/nrr-controller/README.md) for the full list of configurable values.
+
+Requires Helm 3 and Kubernetes 1.25+ (`kubeVersion: ">=1.25.0-0"`).
+
+**Minimal installation:**
+
+```sh
+git clone https://github.com/kubernetes-sigs/node-readiness-controller.git
+cd node-readiness-controller
+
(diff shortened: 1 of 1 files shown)


docs/book/src/user-guide/installation.md
```diff
@@ -62,7 +62,56 @@ REPO="registry.k8s.io/node-readiness-controller/node-readiness-controller"
 TAG=$(skopeo list-tags docker://$REPO | jq .'Tags[-1]' | tr -d '"')
 docker pull $REPO:$TAG
 ```
-### Option 2: Advanced Deployment (Kustomize)
+
+### Option 2: Helm Chart
+
+A Helm chart is maintained in-tree at [`charts/nrr-controller`](https://github.com/kubernetes-sigs/node-readiness-controller/tree/main/charts/nrr-controller). Use it when you want the optional components (metrics, webhook, cert-manager integration) toggled through values rather than by composing Kustomize overlays.
+
+> [!NOTE]
```
ajaysundark, line 70: we are looking to handle this for upcoming release. I think this doc update could wait for it to not confuse readers.

ajaysundark:
Looks like there are two PRs toward helm installation docs. I'm closing this one as the other one seems to be more actively reviewed.

Thanks for your PR!

Duplicate Of #388

/close

PR #396, opened 2026-08-12
docs: point the install guide at the current release

### LINKED ISSUE #395: [BUG] Install guide pins v0.1.1, and that release has no install-full.yaml
### What happened?

The install guide pins `VERSION=v0.1.1`, which was released in January. Latest is `v0.4.1`, so anyone following the guide installs a build that is three releases behind.

It also breaks the full install. The guide offers `install-full.yaml`, but `v0.1.1` never shipped that artifact, it only has `crds.yaml` and `install.yaml`. So following the guide and picking the full option gives a 404. `install-full.yaml` first appears in `v0.2.0`.

`docs/book/src/releases.md` in the same book correctly lists `v0.4.1`, so the two pages disagree.

The reason it keeps going stale is that RELEASE.md step 2 tells the shepherd to update `docs/book/src/releases.md` but never mentions the install guide, so the pinned version is never touched at release time.

### Steps to Reproduce

1. Open the Installation page and follow it as written.
2. `VERSION=v0.1.1`, then pick the Full Installation option.
3. `curl -sIL https://github.com/kubernetes-sigs/node-readiness-controller/releases/download/v0.1.1/install-full.yaml` returns 404.

### Expected Behavior

The guide should point at the current release, and the release process should keep it that way.

docs on `main` (commit `d74dabd`)

### Kubernetes Version

Not applicable, docs only.

Happy to send the PR. Related to #151 since it is a manual release step that is being missed.

tejassinghbhati:
## Description

The install guide pinned `VERSION=v0.1.1` from January while latest is `v0.4.1`, so anyone following it installed a build three releases behind. `docs/book/src/releases.md` in the same book correctly lists `v0.4.1`, so the two pages disagreed.

It also broke the full install path. The guide offers `install-full.yaml`, but `v0.1.1` never shipped that artifact. Following the guide as written and picking the full option returns a 404. Checking the published assets, `install-full.yaml` first appears in `v0.2.0`:

```
v0.1.0 -> crds.yaml, install.yaml
v0.1.1 -> crds.yaml, install.yaml
v0.2.0 -> crds.yaml, install-full.yaml, install.yaml
v0.4.1 -> crds.yaml, install-full.yaml, install.yaml
```

So this bumps the pin, adds a note about which releases carry `install-full.yaml`, and adds the install guide to the release checklist.

The RELEASE.md line is the part that stops it happening again. Step 2 already told the shepherd to update `releases.md` but never mentioned this page, which is why the pin sat untouched across three releases. I considered resolving the tag dynamically with `curl` and `jq` instead, but that puts a `jq` dependency on the primary install path, and a checklist line matches how the project already handles `releases.md`.

## Related Issue

Fixes #395

## Type of Change

/kind documentation

## Testing

Docs only. I verified every artifact the guide references actually resolves at the new pin:

```
crds.yaml        -> HTTP 200
install.yaml     -> HTTP 200
install-full.yaml -> HTTP 200
```

and confirmed the `v0.2.0` claim against the published release assets rather than assuming it.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
NONE
```

### DIFF
--- a/docs/book/src/user-guide/installation.md
+++ b/docs/book/src/user-guide/installation.md
@@ -20,9 +20,6 @@ kubectl apply -f https://github.com/kubernetes-sigs/node-readiness-controller/re
 kubectl wait --for condition=established --timeout=30s crd/nodereadinessrules.readiness.node.x-k8s.io
 ```
 
-> [!NOTE]
-> `install-full.yaml` is only published from `v0.2.0` onwards. Older releases ship `crds.yaml` and `install.yaml` only.
-
 #### 2. Install the Controller
 
 Choose one of the two following manifests based on your requirements:
--- a/docs/book/src/user-guide/installation.md
+++ b/docs/book/src/user-guide/installation.md
@@ -14,7 +14,7 @@ First, to install the CRDs, apply the `crds.yaml` manifest:
 
 ```sh
 # Replace with the desired version
-VERSION=v0.4.1
+VERSION={{#include ../../../../VERSION}}
 
 kubectl apply -f https://github.com/kubernetes-sigs/node-readiness-controller/releases/download/${VERSION}/crds.yaml
 kubectl wait --for condition=established --timeout=30s crd/nodereadinessrules.readiness.node.x-k8s.io
--- a/docs/book/src/user-guide/installation.md
+++ b/docs/book/src/user-guide/installation.md
@@ -13,12 +13,16 @@ If you plan to use the `install-full.yaml` option (which includes secure metrics
 First, to install the CRDs, apply the `crds.yaml` manifest:
 
 ```sh
-# Replace with the desired version
-VERSION=v0.1.1
+# Latest release. Other versions are listed on the Releases page:
+# https://github.com/kubernetes-sigs/node-readiness-controller/releases
+VERSION=v0.4.1
+
 kubectl apply -f https://github.com/kubernetes-sigs/node-readiness-controller/releases/download/${VERSION}/crds.yaml
 kubectl wait --for condition=established --timeout=30s crd/nodereadinessrules.readiness.node.x-k8s.io
-
 ```
+
+> [!NOTE]
+> `install-full.yaml` is only published from `v0.2.0` onwards. Older releases ship `crds.yaml` and `install.yaml` only.
 
 #### 2. Install the Controller
 
--- a/docs/book/src/user-guide/installation.md
+++ b/docs/book/src/user-guide/installation.md
@@ -13,8 +13,7 @@ If you plan to use the `install-full.yaml` option (which includes secure metrics
 First, to install the CRDs, apply the `crds.yaml` manifest:
 
 ```sh
-# Latest release. Other versions are listed on the Releases page:
-# https://github.com/kubernetes-sigs/node-readiness-controller/releases
+# Replace with the desired version
 VERSION=v0.4.1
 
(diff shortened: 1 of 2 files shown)


docs/book/src/user-guide/installation.md
```diff
@@ -13,13 +13,17 @@ If you plan to use the `install-full.yaml` option (which includes secure metrics
...
+# Latest release. Other versions are listed on the Releases page:
+# https://github.com/kubernetes-sigs/node-readiness-controller/releases
+VERSION=v0.4.1
+
 kubectl apply -f https://github.com/kubernetes-sigs/node-readiness-controller/releases/download/${VERSION}/crds.yaml
 kubectl wait --for condition=established --timeout=30s crd/nodereadinessrules.readiness.node.x-k8s.io
-
 ```
 
+> [!NOTE]
```
ajaysundark: @tejassinghbhati I dont see how it is more helpful to the reader. The expectation is that users will follow release documentation and work on the latest version, and the comment above is already suggestive of that?

If there's anything we can do, we could capture the VERSION file here via include tags. ref: https://rust-lang.github.io/mdBook/format/mdbook.html#including-files

tejassinghbhati: Fair, dropped it in ef2e2d2. Only the pin changes now, comment stays as it was.

The include idea is better than what I did, but it does not work yet. VERSION says v0.4.0 and there is no v0.4.0 release, it 404s. releases.md explains why: the images for that tag failed to publish so it got retagged v0.4.1. So including VERSION today would break every download link on the page.

Once VERSION lines up with a real release, say after the 0.5.0 cut, the include is the right fix and the RELEASE.md line I added here can go away with it. Happy to send that as a follow up. Want me to leave the RELEASE.md bit in for now as the stopgap, or drop it and just do the include later?

tejassinghbhati:
Rebased, no conflicts with the RELEASE.md changes from #402 and #407.

Still open question from earlier: keep the RELEASE.md line as the stopgap, or drop it and do the mdBook include once VERSION points at a real release?

docs/book/src/user-guide/installation.md
ajaysundark: Would this work now as the version is at 0.5.0?

tejassinghbhati: Updated in the latest commit to use {{#include ../../../../VERSION}} instead of hardcoding the version pin.

docs/book/src/user-guide/installation.md
```diff
@@ -14,12 +14,15 @@ First, to install the CRDs, apply the `crds.yaml` manifest:
...
-VERSION=v0.1.1
+VERSION={{#include ../../../../VERSION}}
+
 kubectl apply -f https://github.com/kubernetes-sigs/node-readiness-controller/releases/download/${VERSION}/crds.yaml
 kubectl wait --for condition=established --timeout=30s crd/nodereadinessrules.readiness.node.x-k8s.io
-
 ```
 
+> [!NOTE]
+> `install-full.yaml` is only published from `v0.2.0` onwards. Older releases ship `crds.yaml` and `install.yaml` only.
```
ajaysundark: This note can be removed as the only version that didn't ship this is v0.1.x and don't expect anyone to try with the first ever release.

ajaysundark (commented):
/lgtm

docs/book/src/user-guide/installation.md
tejassinghbhati: Done, dropped the note in the latest commit.

ajaysundark (approved):
Thanks!

/lgtm

PR #398, opened 2026-08-12
feat: add scrape-time metrics collector

### LINKED ISSUE #397: [FEATURE] Scrape-Time Metrics Collector
Introduce the new scrape-time collector for metrics computed directly from the controller-runtime cache.

The collector will initially be used by `node_readiness_rule_nodes` and later extended to support the other metrics defined for the scrape-time collector.

Ref. [New Scrape-Time Collector Surface](https://github.com/ajaysundark/node-readiness-controller/blob/9244f4eef28234321fecc988435b846fd3c12507/docs/observability-design.md#4-new-scrape-time-collector-surface)

### Metrics

- [x] `node_readiness_rule_nodes` — implemented in #398
- [ ] `node_readiness_blocked_nodes` — implemented in #431 
- [ ] `node_readiness_rule_matched_nodes` — migrate in a follow-up from #286 

The plan is to implement the new scrape-time metrics incrementally, with the remaining metrics sharing the same `ReadinessCollector` introduced in #398.

rawadhossain:
## Description

Adds `node_readiness_rule_nodes{rule, state="held"|"released"}`, a scrape-time metric for tracking the number of nodes currently held or released by each `NodeReadinessRule`, as described in design doc [New Scrape-Time Collector Surface](https://github.com/ajaysundark/node-readiness-controller/blob/9244f4eef28234321fecc988435b846fd3c12507/docs/observability-design.md#4-new-scrape-time-collector-surface)

The metric is collected directly from the controller runtime cache on each Prometheus scrape using a custom `prometheus.Collector`.

### Changes
- Added the scrape-time `ReadinessCollector`.
- Added `node_readiness_rule_nodes` metric with `rule` and `state` labels.
- Tracks held/released state based on the rule's taint, status conditions are handled separately by `node_readiness_blocked_nodes` (will implement next).
- Added unit tests and benchmark coverage.
- Updated the monitoring doc.

## Type of Change
/kind feature

## Testing
- Verified against a live cluster, counts matched node taints and updated correctly after rule changes.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/docs/book/src/operations/monitoring.md
+++ b/docs/book/src/operations/monitoring.md
@@ -71,6 +71,8 @@ Total number of failure events recorded by the controller.
 
 ### `node_readiness_build_info`
 
+*Available starting from the v0.6.0 release.*
+
 Build information for the node-readiness-controller binary.
 
 | Property | Value |
@@ -84,6 +86,25 @@ Build information for the node-readiness-controller binary.
 | Label | Description | Values |
 | --- | --- | --- |
 | `version` | Build version of the running binary | Any version string, or `unknown` if not set at build time |
+
+### `node_readiness_rule_nodes`
+
+*Available starting from the v0.6.0 release.*
+
+Number of nodes currently held or released by each `NodeReadinessRule`, collected at scrape time from the controller cache.
+
+| Property | Value |
+| --- | --- |
+| Type | `gauge` |
+| Labels | `rule`, `state` |
+| Recorded when | Computed on each Prometheus scrape from the cached node list |
+
+#### Labels
+
+| Label | Description | Values |
+| --- | --- | --- |
+| `rule` | `NodeReadinessRule` name | Any non-dry-run rule name with a valid selector |
+| `state` | Whether matching nodes are still tainted by the rule or have had the taint removed | `held`, `released` |
 
 ### `node_readiness_bootstrap_completed_total`
 
--- a/internal/controller/collector_bench_test.go
+++ b/internal/controller/collector_bench_test.go
@@ -1,0 +1,101 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"fmt"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	readinessv1alpha1 "sigs.k8s.io/node-readiness-controller/api/v1alpha1"
+)
+
+// buildBenchController creates a fake client with the given nodes and rules for benchmarking.
+func buildBenchController(b *testing.B, nodeCount, ruleCount int) *RuleReadinessController {
+	b.Helper()
+
+	scheme := newTestScheme(b)
+	nodes := make([]client.Object, 0, nodeCount)
+	objs := make([]client.Object, 0, nodeCount+ruleCount)
+
+	rules := make(map[string]*readinessv1alpha1.NodeReadinessRule, ruleCount)
+	for i := range ruleCount {
+		ruleName := fmt.Sprintf("rule-%d", i)
+		rule := &readinessv1alpha1.NodeReadinessRule{
+			ObjectMeta: metav1.ObjectMeta{Name: ruleName},
+			Spec: readinessv1alpha1.NodeReadinessRuleSpec{
+				NodeSelector: metav1.LabelSelector{
+					MatchLabels: map[string]string{"rule-index": fmt.Sprintf("%d", i)},
+				},
+				Taint: corev1.Taint{
+					Key:    fmt.Sprintf("readiness.k8s.io/%s", ruleName),
+					Effect: corev1.TaintEffectNoSchedule,
+				},
+			},
+		}
+		rules[ruleName] = rule
+		objs = append(objs, rule)
+	}
+
+	for i := range nodeCount {
+		ruleIdx := i % ruleCount
+		ruleName := fmt.Sprintf("rule-%d", ruleIdx)
+		node := &corev1.Node{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:   fmt.Sprintf("node-%d", i),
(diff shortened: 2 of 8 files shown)


internal/controller/nodereadinessrule_controller.go
```diff
@@ -519,6 +519,52 @@ func (r *RuleReadinessController) getApplicableRulesForNode(ctx context.Context,
...
+	if err := r.List(ctx, nodeList); err != nil {
+		return nil, err
+	}
+
+	log := ctrl.LoggerFrom(ctx)
+
+	counts := make(map[string]metrics.RuleNodeCounts, len(ruleList.Items))
+	for i := range ruleList.Items {
+		rule := &ruleList.Items[i]
+		if rule.Spec.DryRun || !rule.DeletionTimestamp.IsZero() {
```
ajaysundark: should we include the rules being deleted here for giving visibility into the finalizer action? For example, this would be useful when a rule is stuck in deletion due to finalizer path failing

internal/controller/nodereadinessrule_controller.go
```diff
@@ -519,6 +519,52 @@ func (r *RuleReadinessController) getApplicableRulesForNode(ctx context.Context,
...
+			continue
+		}
+
+		// Parse the selector once per rule.
+		selector, err := metav1.LabelSelectorAsSelector(&rule.Spec.NodeSelector)
+		if err != nil {
+			log.Error(err, "Invalid node selector for rule", "rule", rule.Name)
```
ajaysundark: this will be logged per scrape interval, can we consider a lower verbosity, maybe V(2)?

internal/metrics/collector.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// Collect implements prometheus.Collector.
+func (c *ReadinessCollector) Collect(ch chan<- prometheus.Metric) {
+	ctx, cancel := context.WithTimeout(context.Background(), collectTimeout)
+	defer cancel()
+
+	counts, err := c.lister.ListRuleNodeStates(ctx)
+	if err != nil {
+		ch <- prometheus.NewInvalidMetric(ruleNodesDesc, err)
```
ajaysundark: is it treated as an 'internal error' state? what's the implication of this invalid metric for the metrics reader?

ajaysundark: same as other place above, we could log.V(2) instead of erroring out on scrape failures. IMO metrics not-emitted is a better failure mode for serving metrics than 5xx errors.

internal/metrics/collector_test.go
```diff
@@ -0,0 +1,135 @@
...
+	if err := testutil.CollectAndCompare(c, strings.NewReader(expected), "node_readiness_rule_nodes"); err != nil {
+		t.Fatalf("unexpected collect mismatch: %v", err)
+	}
+}
+
+func TestReadinessCollector_ListError(t *testing.T) {
+	c := NewReadinessCollector(&stubLister{err: errors.New("cache not synced")})
+
+	if err := testutil.CollectAndCompare(c, strings.NewReader(""), "node_readiness_rule_nodes"); err == nil {
+		t.Fatal("expected CollectAndCompare to fail on an invalid metric, got nil error")
```
ajaysundark: as discussed above, this could be flipped to assert on 'missing metrics' with successful collectoin

internal/controller/collector_bench_test.go
```diff
@@ -0,0 +1,101 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	readinessv1alpha1 "sigs.k8s.io/node-readiness-controller/api/v1alpha1"
+)
+
+// buildBenchController creates a fake client with the given nodes and rules for benchmarking.
```
ajaysundark, line 31: nice work! we could look into whether these could go into the benchmarking report (from scale tests) or in some shape in monitoring.md later.

cc @vitorfloriano

ajaysundark (changes requested):
@rawadhossain This is in good shape! My main concern is the error handling by the metrics service. PTAL!

internal/controller/nodereadinessrule_controller.go
rawadhossain: yeah, makes sense. Adding deleting rules, so now they show up like any other rule instead of getting excluded

internal/controller/nodereadinessrule_controller.go
rawadhossain: switched it to V(2). Thanks

internal/metrics/collector.go
rawadhossain: you're right, made the changes. Switched this to V(2) logging + skip emitting the metric instead, so now I think a list error won't take down the entire metrics endpoint

internal/metrics/collector_test.go
rawadhossain: flipped it, now asserts empty output with a successful collect

docs/book/src/operations/monitoring.md
```diff
@@ -85,6 +85,23 @@ Build information for the node-readiness-controller binary.
 | --- | --- | --- |
 | `version` | Build version of the running binary | Any version string, or `unknown` if not set at build time |
 
+### `node_readiness_rule_nodes`
```
ajaysundark, line 90: nit: this will not be available until 0.6.0 release. but I'm not sure what's the best way to handle it. I think it also applies to the version-info metrics. is it worth clarifying this with a note on the monitoring page? :/

internal/metrics/collector_test.go
```diff
@@ -0,0 +1,136 @@
...
+	`
+	if err := testutil.CollectAndCompare(c, strings.NewReader(expected), "node_readiness_rule_nodes"); err != nil {
+		t.Fatalf("unexpected collect mismatch: %v", err)
+	}
+}
+
+func TestReadinessCollector_MixedHeldReleased(t *testing.T) {
+	c := NewReadinessCollector(&stubLister{counts: map[string]RuleNodeCounts{
+		"gpu-ready": {Held: 3, Released: 7},
+		"cni-ready": {Held: 0, Released: 10},
```
ajaysundark, line 71: @rawadhossain is there a way to explicitly add a test-case, when multiple rules {A, B} are involved, readiness-taint-A present on a node should count as held for rule-A, and released for rule-B (because readiness-taint-B is non-blocking)

ajaysundark (commented):
/lgtm

Thanks for iterating on this. I left some comments, let us resolve them with decisions before merging this.

docs/book/src/operations/monitoring.md
rawadhossain: yeah will have to figure it out. For now, added a note for both the metrics in the monitoring doc.

internal/metrics/collector_test.go
rawadhossain: Added a test for this in `rule_node_states_test.go` [here](https://github.com/rawadhossain/node-readiness-controller/blob/156aab1f52c96b96a322b8f9474f4f1f925fcf5b/internal/controller/rule_node_states_test.go#L185) instead, since that’s where the actual taint per rule behavior is covered. This test covers one rule being held while the other is released

rawadhossain:
@ajaysundark if everything looks good now, can we get this merged?

PR #399, opened 2026-08-13
fix: avoid taint cleanup for dry-run rules (#374)

### LINKED ISSUE #374: [BUG] Deleting a dry-run NodeReadinessRule removes matching pre-existing taints
### What happened?

A `NodeReadinessRule` that was created with `dryRun: true` and never entered enforcement removes a matching pre-existing Node taint when the Rule is deleted.

Normal dry-run reconciliation does not mutate the Node, but deletion reaches finalizer cleanup before the dry-run branch. `reconcileDelete` then removes matching taints from selected Nodes:

- [deletion is handled before dry-run](https://github.com/kubernetes-sigs/node-readiness-controller/blob/af1f57d587fbf445d61aa7cfc741f8c96a2f5a44/internal/controller/nodereadinessrule_controller.go#L133-L145)
- [deletion cleanup removes matching taints](https://github.com/kubernetes-sigs/node-readiness-controller/blob/af1f57d587fbf445d61aa7cfc741f8c96a2f5a44/internal/controller/nodereadinessrule_controller.go#L669-L696)

This violates the documented dry-run contract that no taints are applied or removed.

### Steps to reproduce

1. Label and taint a test Node:

```bash
kubectl label node <node> dryrun-delete-repro=true
kubectl taint node <node> readiness.k8s.io/dryrun-delete-repro=preexisting:NoSchedule
```

2. Apply this Rule:

```yaml
kind: NodeReadinessRule
metadata:
  name: dryrun-delete-repro
spec:
  dryRun: true
  enforcementMode: continuous
  nodeSelector:
    matchLabels:
      dryrun-delete-repro: "true"
  conditions:
    - type: example.com/TestReady
      requiredStatus: Unknown
  taint:
    key: readiness.k8s.io/dryrun-delete-repro
    value: preexisting
    effect: NoSchedule
```

3. Wait for `status.dryRunResults`, confirm the taint still exists, then delete the Rule:

```bash
kubectl delete nodereadinessrule dryrun-delete-repro --wait=true
kubectl get node <node> -o jsonpath='{.spec.taints}'
```

The pre-existing taint is removed.

### Expected behavior

Deleting a Rule that remained in dry-run for its entire lifetime must not add, update, or remove Node taints. The finalizer and controller-local state may still be cleaned up normally.

This report intentionally covers only the unambiguous always-dry-run case. Lifecycle behavior for a Rule that previously entered enforcement is discussed separately in the comments.

### Environment

- Controller: `nrc-local:021cd1d`
- Main inspected: `af1f57d587fbf445d61aa7cfc741f8c96a2f5a44`
- Kubernetes client: `v1.35.2` (`darwin/arm64`)
- Kubernetes server: `v1.36.1` (`linux/arm64`, Kind)

/kind bug

alanhuangch:
## Description

Prevent deleting an always-dry-run NodeReadinessRule from removing pre-existing Node taints. The cleanup finalizer is now added only when a rule enters enforcement and is retained across later transitions back to dry-run.

## Related Issue

Fixes #374

## Type of Change

/kind bug

## Testing

- `make test` passes (including controller-gen, gofmt, go vet, and all non-e2e tests; controller suite 64/64)
- `make lint` passes (golangci-lint and kube-api-linter)

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

Yes: deleting a rule that remained in dry-run mode no longer mutates Node taints.

```release-note
Deleting an always-dry-run NodeReadinessRule no longer removes matching pre-existing Node taints.
```

### DIFF
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -115,7 +115,10 @@ func (r *RuleReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	log = log.WithValues("ruleName", rule.Name)
 	ctx = ctrl.LoggerInto(ctx, log)
 
-	// Add finalizer first if not set to avoid the race condition between init and delete.
+	// Add the cleanup finalizer only once the rule is enforceable. A dry-run-only
+	// rule must not perform taint cleanup when it is deleted. Once added, the
+	// finalizer is retained across later dry-run reconciliations as a persistent
+	// indication that the rule may have previously entered enforcement.
 	if finalizerAdded, err := r.ensureFinalizer(ctx, rule, finalizerName); err != nil {
 		return ctrl.Result{}, err
 	} else if finalizerAdded {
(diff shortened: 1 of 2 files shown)


ajaysundark:
I understand the rationale here, but wouldnt a 'stuck' taint has more side effects than cleaning it up? From a safety standpoint, erring on the side of cleaning up taints might be preferable.

Even in a scenario where a user deletes a dry-run rule to update and reapply it, NRC would re-add missing taints to enforce readiness anyway. Could you clarify where do you see leaving the taints is  preferred here?

PR #404, opened 2026-08-13
fix: stop the chart suggesting an empty nodeSelector

### LINKED ISSUE #403: [BUG] Helm chart suggests nodeSelector: {} but the webhook rejects it
### What happened?

The Helm chart tells users to use `nodeSelector: {}` to match all nodes, but the validating webhook rejects exactly that.

The chart's error message for `nodeReadinessRules` says:

```
spec.nodeSelector is required and immutable. Set it explicitly; use 'nodeSelector: {}' only if the rule is intended to match ALL nodes.
```

The webhook does this in `validateSpec`:

```go
if selector != nil && selector.Empty() {
}
```

So a user who follows the chart's own guidance and enables the webhook gets the rule rejected on create. The chart also has a test named `renders full rule spec with explicit all-nodes selector` that renders `nodeSelector: {}` and asserts it comes out empty, so the chart clearly treats it as supported while the webhook treats it as invalid.

The other half is what happens with the webhook off, which is the default. Nothing else blocks an empty selector. The CRD has no CEL rule for it, and `metav1.LabelSelectorAsSelector` turns `{}` into `labels.Everything()`, so the rule matches every node in the cluster. With `NoExecute` that evicts pods everywhere that lack a toleration.

So the same manifest either fails admission or taints the whole cluster, depending on a flag that is off by default.

### Steps to Reproduce

1. Install the chart with `webhook.enabled=true` and `validatingWebhook.enabled=true`.
2. Add a rule under `nodeReadinessRules` with `nodeSelector: {}`, as the chart's message suggests.
3. The rule is rejected with `nodeSelector must not be empty`.
4. Install without the webhook and the same rule is accepted, and applies to every node.

### Expected Behavior

The chart should not point users at a configuration the webhook forbids.

Beyond fixing the message, there is a question I do not think I should answer on my own: should an empty selector be blocked for everyone rather than only when the optional webhook is running? Today the only guard against a rule that taints every node lives in a component that is off by default. Moving it into the CRD as a CEL rule would make it consistent, but it is an API change and would reject any existing rule that relies on it, so it needs a maintainer call.

`main` (commit `d74dabd`)

### Kubernetes Version

Not version specific.

I checked before filing. No existing issue covers empty node selectors, and neither #315 nor #352 touches the chart template or the webhook's empty-selector check, #315 only shows that line as unchanged context.

Sending a PR for the message. Happy to follow up on the CRD question separately if there is appetite.

tejassinghbhati:
## Description

The chart's error message for `nodeReadinessRules` told users to set `nodeSelector: {}` when a rule should match all nodes:

```
spec.nodeSelector is required and immutable. Set it explicitly; use 'nodeSelector: {}' only if the rule is intended to match ALL nodes.
```

The validating webhook rejects exactly that, in `validateSpec`:

```go
if selector != nil && selector.Empty() {
}
```

So following the chart's own guidance with the webhook enabled fails on create. With the webhook off, which is the default, nothing blocks it and the rule matches every node in the cluster instead.

This reworks the message so it stops recommending it and says what actually happens either way. It is message only, no behaviour change. The chart's existing `renders full rule spec with explicit all-nodes selector` test still renders `nodeSelector: {}`, and the `fails rendering when nodeSelector is omitted` test matches on `spec.nodeSelector is required`, which is unchanged.

I left the larger question in #403 rather than deciding it here: the only guard against a rule that taints the whole cluster currently lives in a component that is off by default, and moving it into the CRD as a CEL rule would be an API change that rejects existing rules. That needs a maintainer call, not a drive-by.

## Related Issue

Fixes #403

## Type of Change

/kind bug

## Testing

`helm unittest charts/nrr-controller --strict` passes, 19 tests across 5 suites, unchanged from before. `helm lint` clean.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
NONE
```

### DIFF
--- a/charts/nrr-controller/templates/nodereadinessrules.yaml
+++ b/charts/nrr-controller/templates/nodereadinessrules.yaml
@@ -2,7 +2,7 @@
 {{- range $rule := .Values.nodeReadinessRules }}
 {{- $spec := omit $rule "name" }}
 {{- if or (not (hasKey $spec "nodeSelector")) (kindIs "invalid" $spec.nodeSelector) }}
-{{- fail (printf "nodeReadinessRules[%s]: spec.nodeSelector is required and immutable. Set it explicitly; use 'nodeSelector: {}' only if the rule is intended to match ALL nodes." ($rule.name | default "?")) }}
+{{- fail (printf "nodeReadinessRules[%s]: spec.nodeSelector is required and immutable. Set it explicitly. An empty selector matches every node in the cluster, and is rejected by the validating webhook when that is enabled." ($rule.name | default "?")) }}
 {{- end }}
 ---
 apiVersion: readiness.node.x-k8s.io/v1alpha1
--- a/charts/nrr-controller/templates/nodereadinessrules.yaml
+++ b/charts/nrr-controller/templates/nodereadinessrules.yaml
@@ -2,7 +2,7 @@
 {{- range $rule := .Values.nodeReadinessRules }}
 {{- $spec := omit $rule "name" }}
 {{- if or (not (hasKey $spec "nodeSelector")) (kindIs "invalid" $spec.nodeSelector) }}
-{{- fail (printf "nodeReadinessRules[%s]: spec.nodeSelector is required and immutable. Set it explicitly. An empty selector matches every node in the cluster, and is rejected by the validating webhook when that is enabled." ($rule.name | default "?")) }}
+{{- fail (printf "nodeReadinessRules[%s]: spec.nodeSelector is required and immutable. An empty selector to match every node in the cluster is rejected by the validating webhook to avoid cluster wide misconfiguration risks." ($rule.name | default "?")) }}
 {{- end }}
 ---
 apiVersion: readiness.node.x-k8s.io/v1alpha1


charts/nrr-controller/templates/nodereadinessrules.yaml
```diff
@@ -2,7 +2,7 @@
 {{- range $rule := .Values.nodeReadinessRules }}
 {{- $spec := omit $rule "name" }}
 {{- if or (not (hasKey $spec "nodeSelector")) (kindIs "invalid" $spec.nodeSelector) }}
-{{- fail (printf "nodeReadinessRules[%s]: spec.nodeSelector is required and immutable. Set it explicitly; use 'nodeSelector: {}' only if the rule is intended to match ALL nodes." ($rule.name | default "?")) }}
+{{- fail (printf "nodeReadinessRules[%s]: spec.nodeSelector is required and immutable. Set it explicitly. An empty selector matches every node in the cluster, and is rejected by the validating webhook when that is enabled." ($rule.name | default "?")) }}
```
ajaysundark:
```suggestion
{{- fail (printf "nodeReadinessRules[%s]: spec.nodeSelector is required and immutable. An empty selector to match every node in the cluster is rejected by the validating webhook to avoid cluster wide misconfiguration risks." ($rule.name | default "?")) }}
```

ajaysundark:
/lgtm

Thanks. I suggested rewording this a bit, could you ptal!

tejassinghbhati:
Thanks! I can't see the suggestion on my end, nothing showing up as a review comment on the diff. Might still be pending on your side, could you submit it?
The verify-all failure looks unrelated. It is lychee getting a 403 on the Slack links in README.md and introduction.md, neither of which this PR touches. #379 passed verify-all on the same base commit, so it looks intermittent rather than broken.
/retest

ajaysundark:
cc @ajaysundark

tejassinghbhati:
Taken, thanks.

One thing so you have it: the webhook is off by default, so the empty selector only actually gets rejected when someone turns it on. Default install it just goes through and matches everything. Your wording reads like it is always blocked. Fine by me either way, that gap is really #403's problem not this PR's.

The verify-all failure is unrelated by the way, lychee 403s on the Slack links in README.md and introduction.md, neither of which this PR touches. #379 passed on the same base so it looks like a flake.

/retest

tejassinghbhati:
Wording is in as fe723bb, so this just needs an /ok-to-test and the lgtm back when you get a minute. My /retest bounced since I am not in the org yet.

Same for #372, #388 and #396, none of them have had CI run at all.

PR #402, opened 2026-08-13
setup Helm chart publishing to kubernetes staging registry

Priyankasaggu11929:
PR is adding setup to publish Helm chart artifacts to kubernetes staging registry (and eventually to registry.k8s.io)

Took inspiration from the `kubernetes-sigs/kro` repo - https://github.com/kubernetes-sigs/kro/commit/80d485024178c072d9c8bd126e7b0ed5e440aac8

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

/assign @ajaysundark

PR #405, opened 2026-08-13
run helm via go run so it works without a preinstalled binary

Priyankasaggu11929:
fix for https://github.com/kubernetes-sigs/node-readiness-controller/pull/402#issuecomment-5282405064

again, taking inspiration from kro helm setup - https://github.com/kubernetes-sigs/kro/blob/b1b0aec564f6d11f392522705c6e292efe0122fc/Makefile#L6

(since  go  is already present in the CI environment)

/assign @ajaysundark

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -487,23 +487,20 @@ crd-ref-docs:
 
 # helm
 
-ensure-helm-install:
-ifndef HAS_HELM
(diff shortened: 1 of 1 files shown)


ajaysundark (approved):
/lgtm

PR #406, opened 2026-08-13
feat: add node_readiness_build_info metric

rawadhossain:
### Description
Adds `node_readiness_build_info` metric from [observability design doc](https://github.com/ajaysundark/node-readiness-controller/blob/9244f4eef28234321fecc988435b846fd3c12507/docs/observability-design.md#controller-metrics)

### Changes
- Add `node_readiness_build_info` gauge with a `version` label.
- Set the metric at controller startup using the build version.
- Wire build version and git commit metadata through `-ldflags` in the Makefile and Dockerfile.

### Implementation choices
- Use the build version as the `version` label, while keeping commit information in the existing startup version log.
- Added `-ldflags` wiring because the existing `internal/info` version variables were not being populated during builds, which otherwise left the version as `unknown`.

## Type of Change
/kind feature

## Verification
- Verified the metric reports the actual git-derived version from a locally built binary.
- Verified the same version is propagated through the Docker build path.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,6 +77,10 @@ IMG_TAG ?= latest
 
 # OCI registry to publish the Helm chart to
 HELM_IMAGE ?= registry.k8s.io/node-readiness-controller/charts
+# Version metadata embedded into the binary via -ldflags.
+VERSION ?= $(shell git describe --tags --always --dirty 2>/dev/null || echo unknown)
+GIT_COMMIT ?= $(shell git rev-parse --short HEAD 2>/dev/null || echo unknown)
+LDFLAGS := -X sigs.k8s.io/node-readiness-controller/internal/info.version=$(VERSION) -X sigs.k8s.io/node-readiness-controller/internal/info.gitCommit=$(GIT_COMMIT)
 
 # Kind cluster name for loading images
 KIND_CLUSTER ?= nrr-test
@@ -190,7 +194,7 @@ verify: ## Run all verification scripts.
 
 .PHONY: build
 build: manifests generate fmt vet ## Build manager binary.
-	go build -o bin/manager cmd/main.go
+	go build -ldflags "$(LDFLAGS)" -o bin/manager cmd/main.go
 
 .PHONY: run
 run: manifests generate fmt vet ## Run a controller from your host.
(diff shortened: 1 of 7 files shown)


internal/metrics/metrics.go
```diff
@@ -152,6 +152,15 @@ var (
...
 		[]string{"rule"},
 	)
+
+	// BuildInfo exposes the running binary's build version.
+	BuildInfo = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "node_readiness_build_info",
+			Help: "Build information for the node-readiness-controller binary.",
+		},
+		[]string{"version"},
```
Ayush4958: Pls add a git_commit label here `[]string{"version", "git_commit"}`

Prometheus standards require the commit hash on build metrics

cmd/main.go
```diff
@@ -118,6 +119,12 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 	ctrl.Log.Info(fmt.Sprintf("version: %s", info.GetVersionString()))
 
+	buildVersion := "unknown"
+	if parts := info.GetVersionParts(); len(parts) > 0 {
+		buildVersion = parts[0]
```
Ayush4958: Extracting parts[0] is flaky, `info.GetVersionParts()` returns a slice designed for log formatting,
if that internal formatting ever changes, metric break silently. 

Please add strict `GetVersion()` string & `GetGitCommit()` string accessors & use them directly here to populate the metric.

Ayush4958 (changes requested):

cmd/main.go
rawadhossain: Thanks for pointing this out.

internal/metrics/metrics.go
rawadhossain: I kept the metric scoped to the binary version since the commit is already exposed in the startup log.

Makefile
```diff
@@ -202,11 +206,11 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # More info: https://docs.docker.com/develop/develop-images/build_enhancements/
 .PHONY: docker-build
 docker-build: ## Build container image with Docker.
```
ajaysundark, line 208: You want to also add the flags in the docker-buildx and docker-buildx-reporter targets. They are the ones wired in our release path.

Ref: scripts/build-and-publish.sh

internal/metrics/metrics.go
ajaysundark: Looks like there's an idiomatic way to report this: https://github.com/prometheus/client_golang/issues/693

Makefile
rawadhossain: Yeah, added the build args to both targets. Thanks.

internal/metrics/metrics.go
rawadhossain: Thanks for pointing this out, the example helped. Updated it.

ajaysundark: The other identifying labels from prometheus example (eg: revision / goversion) are useful and could be extended as labels here,  but I'm not blocking this CL on this, you could followup on this after -
https://github.com/prometheus/common/blob/2fb5a7cecbffd4bf7dc88ff22712112e8cef8432/version/info.go#L42

internal/metrics/metrics.go
```diff
@@ -152,6 +154,18 @@ var (
 		},
 		[]string{"rule"},
 	)
+
+	// BuildInfo exposes the running binary's build version.
+	BuildInfo = prometheus.NewGaugeFunc(
```
ajaysundark, line 159: are there room to add some tests?

internal/metrics/metrics.go
rawadhossain: Thanks. I’ll do follow-up on this if needed to extend going ahead.

internal/metrics/metrics.go
rawadhossain: Yeah, we can add tests for this. I’ll update it in a while.

rawadhossain:
@ajaysundark added the tests. Should be good to merge.

PR #411, opened 2026-08-14
feat(chart): add PodDisruptionBudget for high availability

### LINKED ISSUE #410: [FEATURE] Helm chart missing PodDisruptionBudget (PDB) for high availability

I was configuring the Helm chart for a production deployment and noticed there's no way to define a `PodDisruptionBudget` (PDB). 

Since this controller is responsible for un-tainting nodes, it's pretty critical infrastructure. If we run multiple replicas for HA, a cluster upgrade or node drain could still potentially evict all replicas at once if there isn't a PDB protecting them.

Add an optional `podDisruptionBudget` block to `values.yaml` (defaulting to disabled/empty so it doesn't break existing setups) and a corresponding `pdb.yaml` template in the chart.

Something standard like:
```yaml
podDisruptionBudget:
  enabled: false
  minAvailable: 1
  # maxUnavailable: 1

Manually maintaining a separate PDB manifest outside the Helm release, but it really belongs inside the chart since it's a standard HA practice.

Happy to open a quick PR for this if you agree!

### PR DESCRIPTION
**Pull request by vishnukothakapu**
## Description
This adds an optional `PodDisruptionBudget` to the Helm chart (`charts/node-readiness-controller`). Since the controller manages node taints, it's critical infrastructure. Adding a PDB allows users running multiple replicas to ensure the controller remains highly available during node drains, cluster upgrades, or other evictions.

The PDB is disabled by default (`podDisruptionBudget.enabled: false`). Existing installations that do not explicitly opt-in to this feature will see no changes in rendered manifests.

## Related Issue
Fixes #410

## Type of Change
/kind feature

## Testing
- Added `helm-unittest` coverage in `charts/node-readiness-controller/tests/pdb_test.yaml` to ensure it only renders when enabled, respects `minAvailable` / `maxUnavailable`, and correctly inherits the selector labels.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Helm chart: Added `podDisruptionBudget` configuration options for high availability during node drains and evictions.

### CODE REVIEW

File: charts/node-readiness-controller/tests/pdb_test.yaml [hunk 1]
```diff
@@ -0,0 +1,53 @@
...
+        maxUnavailable: 1
+    asserts:
+      - isKind:
+          of: PodDisruptionBudget
+      - isNull:
+          path: spec.minAvailable
+      - equal:
+          path: spec.maxUnavailable
+          value: 1
+
```
Ayush4958, line 42: pls add test case for `maxUnavailable: 0` , so 0 value renders correctly & isnt dropped by above template logic

charts/node-readiness-controller/templates/pdb.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if .Values.podDisruptionBudget.enabled }}
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ include "node-readiness-controller.fullname" . }}
+  namespace: {{ include "node-readiness-controller.namespace" . }}
+  labels:
+    {{- include "node-readiness-controller.labels" . | nindent 4 }}
+spec:
+  {{- if .Values.podDisruptionBudget.minAvailable }}
```
Ayush4958: 0 considered false, in case `maxUnavailable: 0` for 100% uptime during node drains, if statement will silently drops field & will render and invalid PDB

u can modify it like this `{{- if not (empty (toString .Values.podDisruptionBudget.maxUnavailable)) }}` so 0 will render properly

Ayush4958 (changes requested):

charts/node-readiness-controller/templates/pdb.yaml
vishnukothakapu: You are completely right, `0` would have evaluated to false and silently dropped the field. I've updated the logic exactly as you suggested using `not (empty (toString ...))` for both `minAvailable` and `maxUnavailable`. Thank you!

Ayush4958 (approved):
It's seems good to me 
@ajaysundark can review for further review

ajaysundark:
same as https://github.com/kubernetes-sigs/node-readiness-controller/pull/437#issuecomment-5365020940. I want to better understand what level of preconfiguration will be handled at the helm chart. Could you help us with some research? Thanks!

ajaysundark:
@Ayush4958 had some references at #437.

It's good to add to our charts. Thanks @vishnukothakapu .

/lgtm

ajaysundark:
the helm tests are failing and need your second look

/assign @vishnukothakapu

Issue #227, opened 2026-05-08
[TRACKING] Improve Contributing Guidelines


vitorfloriano:
Currently, the NRC repository lacks custom guidelines for contributors, which leads to inconsistency on issues/PRs and manual toil on maintainers.

We may consider:

- [x] Adding Issues templates/forms
- [ ] Implementing a standardized label schema (kind/ area/ priority/)
- [ ] Customizing CONTRIBUTING.md (including local setup & dev environment)
- [ ] Document available tests, add instructions for writing tests setting up ProwJobs

[edit: a few changes were deemed not necessary after discussion]
~- [ ] Improving PR templates~
~- [ ] Standardizing commit messages (e.g., Conventional Commits)~
~- [ ] Adding policies for Branch naming~
~- [ ] Defining Merge/Squash policy for clean history~

Adding guidelines/policies on how to contribute helps contributors have a clear path and facilitates automating chores like release notes, changelogs and triaging. It also makes it possible to provide contributors instant feedback through CI checks.

Rajneesh180:
Hi, I'd like to work on this. I can see the repo has a PR template already (`.github/pull_request_template.md`) but no issue templates and a minimal `CONTRIBUTING.md`. 

I'm planning to submit a focused PR that addresses:
1. **GitHub issue templates** — structured YAML forms for bug reports and feature requests under `.github/ISSUE_TEMPLATE/`, following the patterns used in other `kubernetes-sigs` repos
2. **CONTRIBUTING.md improvements** — add a local development setup section (prerequisites, how to run `make test` and `make lint`, how to set up a local cluster with `kind`) and a commit message conventions section (Conventional Commits as used in this repo's recent history)

Would this scope be acceptable? Happy to adjust based on maintainer preferences before opening the PR.

Priyankasaggu11929:
Hello @vitorfloriano, thanks again for creating the issue and suggesting improvements. 

- Re: `Implementing a standardized label schema (kind/ area/ priority/)`

    we use a tool called `label_sync` for adding new labels to github repos - https://github.com/kubernetes/test-infra/tree/master/label_sync
    
    Here's an example PR adding a bunch of labels to `kubernetes/autoscaler` and `kubernetes-sigs/cluster-autoscaler` repos - https://github.com/kubernetes/test-infra/pull/36927
    
    we can start with adding a bunch of basic labels and then re-iterate with more as/when needed.


- Re: `Adding policies for Branch naming`

[...]

Issue #225, opened 2026-05-07
[TRACKING] Define Project Versioning, Compatibility, and Release Playbook


vitorfloriano:
As node-readiness-controller matures (currently v0.3.0), we need a formalised structure for versioning and support to provide stability for our users and a clear roadmap for contributors. This issue tracks the creation and documentation of our versioning strategy.

We need to do things like:

- [ ] Draft Versioning Policy: Define Semantic Versioning (SemVer) usage.
- [ ] Establish Support Lifecycle: Define how many minor versions we support (e.g., N-2).
- [ ] Create Compatibility Matrix: Document the mapping between Controller versions and Kubernetes versions.
- [ ] Write Versioning Playbook: Standardise the release process (tagging, CI, changelogs).
- [ ] Publish Documentation: Add /docs/VERSIONING.md and update README.md.

Relates to #151

[TBD - this must be discussed with maintainers first]

sahitya-chandra:
Hey @vitorfloriano, is it possible I could collaborate on this issue with you?

vitorfloriano:
> Hey [@vitorfloriano](https://github.com/vitorfloriano), is it possible I could collaborate on this issue with you?

Hi @sahitya-chandra 

I think this needs to be throroughly discussed with the maintainers, preferrably in a meeting, before we take any actions. I'll update the description.

sahitya-chandra:
Agreed with the discussion part, I also thought that this is a pretty big issue, so I could help you in some way with it :)

vitorfloriano:
@sahitya-chandra there's also a [thread on Slack](https://kubernetes.slack.com/archives/C09TW8DUVF1/p1777988157826969), if you're interested in following the discussion. 🧵

Priyankasaggu11929:
(follow up from yesterday's (May 13, 2026) project meeting discussion)

I would still like to continue this discussion in the next project meeting, so we can collect more opinions from other maintainers as well (especially @ajaysundark), around what signals/criteria we want to use for defining our compatibility matrix eventually.

Because I feel the same CI signals/criteria will majorly help us towards the near term high priority item - of defining API graduation (and maintenance) criteria for the NRC controller.


Meanwhile - the work for migrating our CI jobs to prow will be helpful to establish some preliminary set of CI signals - https://github.com/kubernetes-sigs/node-readiness-controller/pull/210#issuecomment-4441678380

---

Also just for record keeping - below are some references from the NPD project (around establishing support matrix):

[...]

PR #416, opened 2026-08-16
infra: add release notes config

vitorfloriano:
This PR adds a configuration file for the release notes generated by GitHub.

See: https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes#configuring-automatically-generated-release-notes

#### Prior art:

- cluster-api-provider-aws:
  - config: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/67de5c25543b2ec9c76897670fc928309473fbe1/.github/release.yml#L4
  - release notes: https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/tag/v2.12.1

- secrets-store-sync-controller:
  - config: https://github.com/kubernetes-sigs/secrets-store-sync-controller/blob/35418ee1ac4e81d9cca4b44182bf19ef355f80b2/.github/release.yml#L4
  - release notes: https://github.com/kubernetes-sigs/secrets-store-sync-controller/releases/tag/v0.0.5

> [!NOTE]
> I'm only adding labels currently available for the repo.  We can add more [here](https://github.com/kubernetes/test-infra/blob/50f8ef5e15cfa16916e64d424adb5e55ed6d8016/label_sync/labels.yaml) if needed. I specially missed something like a `kind/tests` label. 

#### Open question

Should we try some emojis ⚠️ ✨ 🐛 📖 🧹 ?

xref #225 #227

### DIFF
--- a/.github/release.yml
+++ b/.github/release.yml
@@ -1,0 +1,26 @@
+changelog:
+  categories:
+    - title: "API Changes"
+      labels:
+        - kind/api-change
+        - kind/deprecation
+    - title: "New Features"
+      labels:
+        - kind/feature
+    - title: "Fixes"
+      labels:
+        - kind/bug
+        - kind/regression
+    - title: "Docs"
+      labels:
+        - kind/documentation
+        - documentation
+    - title: "Chores"
+      labels:
+        - area/dependency
+        - kind/flake
+        - kind/failing-test
+        - kind/cleanup
+    - title: "Other"
+      labels:
+        - "*"


.github/release.yml
```diff
@@ -0,0 +1,26 @@
+changelog:
+  categories:
+    - title: "API Changes"
+      labels:
+        - kind/api-change
```
ajaysundark, line 5: So, based on the labels it also groups into corresponding release categories? My only concern is that contributors may not know to use the correct labels. We would expect the reviewers to ensure correctly labeled etc.

ajaysundark (approved):

.github/release.yml
vitorfloriano: > We would expect the reviewers to ensure correctly labeled etc.

I believe this problem is inherent to letting authors label their own PRs. Upstream k/k may be subject to this as well, I think. 

Some adjustments will probably be necessary when drafting/generating the release notes, but those should be minor (and easily spotted), I guess?

Or we could remove the labels from the PR template and let reviewers add them. I believe the prow robot will add a `do-not-merge/needs-kind` label if none is attributed to the PR (at least that's how it works with k/k).

Another cleanup would be removing the field for adding release notes from the PR template, since we won't use that anyway.

ajaysundark: > I believe the prow robot will add a do-not-merge/needs-kind label if none is attributed to the PR

Hmm, I haven't seen this for NRC. It'd be good to validate this behavior

ajaysundark: > cleanup would be removing the field for adding release notes from the PR template, since we won't use that anyway

I'm okay with keeping it as-is, even if we decide not to directly consume it. 

2c: GitHub's release notes generation also seem to use a LLM backend - it could work for a way to maintainers to nudge it in the right direction. Many of the AI generated issues we have seen recently claims catastrophe even if this is a likely a non-probable or less likely issue, we don't want to carry that same alarm to the release notes as well.

ajaysundark:
> Open question
> Should we try some emojis ⚠️ ✨ 🐛 📖 🧹 ?

I'm not a fan because we can't expect the same rendering (or same interpretation) across the consumers. But I'm not against it if there are other opinions

PR #418, opened 2026-08-17
feat: add validation to prevent anyOf misconfiguration

### LINKED ISSUE #414: [FEATURE] anyOf conditionPolicy with `defaultStatus == requiredStatus` silently short-circuits rule
### What happened?

During the review of PR #315, @ajaysundark identified a misconfiguration risk with `anyOf` rules.

When `conditionPolicy` is set to `anyOf`, the rule is satisfied as soon as any single condition matches its `requiredStatus`. If one of those conditions also defines a `defaultStatus` equal to its `requiredStatus`, the controller treats the condition as satisfied whenever it is absent from the Node, because the default value already matches the requirement.

This causes the `anyOf` evaluation to short-circuit immediately, and the taint is never applied. The misconfiguration is completely silent; nothing in the status or events indicates the rule is effectively a no-op.

### Steps to Reproduce

Create a rule with `conditionPolicy: anyOf` that includes:

yaml
```
conditions:
  - type: "gpu.example.com/HardwareDriverReady"
    requiredStatus: "True"
    defaultStatus: "True"   # ← matches requiredStatus
```

If the Node never reports `HardwareDriverReady`, the controller defaults it to True, which satisfies the `anyOf` check. The taint is never applied, and the Node appears ready despite the driver never having initialized.

### Expected Behavior

The controller should reject this misconfiguration. As discussed with @AnuragThePathak in PR #315, we should:
1. Add a Go webhook validation to reject NodeReadinessRule objects where conditionPolicy is anyOf and any condition's defaultStatus equals its requiredStatus.
2. Update the documentation to explicitly describe this restriction.

main branch / PR #315

### Kubernetes Version

N/A

### Controller Logs

N/A

This follow-up issue was created based on the post-review discussion in PR #315.

vishnukothakapu:
**Description**
This PR adds CEL validation to the CRD to explicitly reject `NodeReadinessRule` configurations where the `conditionPolicy` is set to `anyOf` and any of the defined conditions has a `defaultStatus` equal to its `requiredStatus`.

As discussed with @ajaysundark & @AnuragThePathak in PR #315, allowing `defaultStatus == requiredStatus` within an `anyOf` rule creates a silent logical short-circuit, causing the rule to be considered satisfied even if the condition is never reported by the Node. 

By enforcing this natively via CEL XValidation, we keep the webhook lean while still preventing the misconfiguration. This PR also adds a warning to the concepts documentation.

**Related Issue**
Fixes #414

**Type of Change**
/kind feature
/kind documentation

**Testing**
- Restructured `nodereadinessrule_controller_test.go` to include `envtest` coverage verifying that the API server actively rejects `anyOf` + `defaultStatus` configurations via CEL.
- Verified all Ginkgo test suites pass.

**Checklist**
- [x] `make test` passes
- [x] `make lint` passes

**Does this PR introduce a user-facing change?**
```release-note
Feature: Added CEL validation to the NodeReadinessRule CRD to reject configurations using `conditionPolicy: anyOf` when a condition's `defaultStatus` is equal to its `requiredStatus`, as this results in a silent evaluation short-circuit.

### DIFF
--- a/api/v1alpha1/nodereadinessrule_types.go
+++ b/api/v1alpha1/nodereadinessrule_types.go
@@ -66,6 +66,7 @@ const (
 // NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
 //
 // +kubebuilder:validation:XValidation:rule="(!has(oldSelf.conditionPolicy) ? 'allOf' : oldSelf.conditionPolicy) == (!has(self.conditionPolicy) ? 'allOf' : self.conditionPolicy)",message="conditionPolicy is immutable"
+// +kubebuilder:validation:XValidation:rule="!has(self.conditionPolicy) || self.conditionPolicy != 'anyOf' || self.conditions.all(c, !has(c.defaultStatus))",message="defaultStatus is not supported when conditionPolicy is anyOf"
 type NodeReadinessRuleSpec struct {
 	// conditions contains a list of the Node conditions that defines the specific
 	// criteria that must be met for taints to be managed on the target Node.
--- a/docs/book/src/user-guide/concepts.md
+++ b/docs/book/src/user-guide/concepts.md
@@ -36,6 +36,11 @@ When a rule specifies multiple conditions, the `conditionPolicy` determines how 
 >
 > `bootstrap-only` mode exists to verify that specific components have finished initializing. Using `anyOf` with `bootstrap-only` could lead to the node bootstrapping prematurely if just one component is ready, completely ignoring the initialization status of the others. The admission webhook enforces this restriction.
 
+> [!CAUTION]
+> **`anyOf` conditions cannot have `defaultStatus` equal to `requiredStatus`.**
+>
+> Because `anyOf` only requires a single condition to be satisfied, configuring a condition where `defaultStatus` is the same as its `requiredStatus` creates a logical short-circuit. The condition would be considered satisfied even if the node never reports it, silently bypassing the readiness check. The admission webhook actively rejects configurations with this vulnerability.
+
 
 ## Enforcement Modes
 
--- a/docs/book/src/user-guide/concepts.md
+++ b/docs/book/src/user-guide/concepts.md
@@ -32,14 +32,14 @@ When a rule specifies multiple conditions, the `conditionPolicy` determines how 
 - **`anyOf`**: At least ONE condition must match its `requiredStatus`. If no conditions match, the node is tainted. This is particularly useful for hardware that may be satisfied by multiple drivers, or systems with active/passive fallbacks.
 
 > [!IMPORTANT]
-> **`anyOf` is not supported with `bootstrap-only` rules.**
->
-> `bootstrap-only` mode exists to verify that specific components have finished initializing. Using `anyOf` with `bootstrap-only` could lead to the node bootstrapping prematurely if just one component is ready, completely ignoring the initialization status of the others. The admission webhook enforces this restriction.
+> **API Constraints for `anyOf`**
+> 
+> The API will actively reject `anyOf` configurations in the following scenarios:
+> - **Combined with `bootstrap-only` enforcement**: This mode verifies that specific components have finished initializing. Using `anyOf` could lead to the node bootstrapping prematurely if just one condition is met.
+> - **Combined with `defaultStatus`**: This creates a logical short-circuit when the condition is missing and `defaultStatus` is set to match `requiredStatus`. The condition would be considered satisfied even if the node never reports it, silently bypassing the readiness check.
 
-> [!CAUTION]
-> **`anyOf` conditions cannot have `defaultStatus` equal to `requiredStatus`.**
->
-> Because `anyOf` only requires a single condition to be satisfied, configuring a condition where `defaultStatus` is the same as its `requiredStatus` creates a logical short-circuit. The condition would be considered satisfied even if the node never reports it, silently bypassing the readiness check. The admission webhook actively rejects configurations with this vulnerability.
+
+
 
 
 ## Enforcement Modes
--- a/internal/webhook/nodereadinessgaterule_webhook.go
+++ b/internal/webhook/nodereadinessgaterule_webhook.go
@@ -98,6 +98,18 @@ func (w *NodeReadinessRuleWebhook) validateSpec(
 		))
 	}
 
+	// validate that when conditionPolicy is anyOf, no condition's defaultStatus equals its requiredStatus
+	if spec.GetConditionPolicy() == readinessv1alpha1.ConditionPolicyAnyOf {
+		for i, cond := range spec.Conditions {
+			if cond.GetDefaultStatus() == cond.RequiredStatus {
+				allErrs = append(allErrs, field.Forbidden(
+					field.NewPath("spec", "conditions").Index(i).Child("defaultStatus"),
+					"defaultStatus cannot be equal to requiredStatus when conditionPolicy is anyOf",
+				))
+			}
+		}
+	}
+
 	return allErrs
 }
 
--- a/internal/webhook/nodereadinessgaterule_webhook.go
+++ b/internal/webhook/nodereadinessgaterule_webhook.go
@@ -98,18 +98,6 @@ func (w *NodeReadinessRuleWebhook) validateSpec(
 		))
 	}
 
-	// validate that when conditionPolicy is anyOf, no condition's defaultStatus equals its requiredStatus
-	if spec.GetConditionPolicy() == readinessv1alpha1.ConditionPolicyAnyOf {
-		for i, cond := range spec.Conditions {
-			if cond.GetDefaultStatus() == cond.RequiredStatus {
-				allErrs = append(allErrs, field.Forbidden(
-					field.NewPath("spec", "conditions").Index(i).Child("defaultStatus"),
-					"defaultStatus cannot be equal to requiredStatus when conditionPolicy is anyOf",
-				))
-			}
-		}
-	}
-
 	return allErrs
 }
 
--- a/internal/webhook/nodereadinessgaterule_webhook_test.go
+++ b/internal/webhook/nodereadinessgaterule_webhook_test.go
@@ -203,83 +203,10 @@ var _ = Describe("NodeReadinessRule Validation Webhook", func() {
 			})
 		})
 
-		Context("anyOf conditionPolicy defaultStatus validation", func() {
-
-			It("should reject when conditionPolicy is anyOf and defaultStatus equals requiredStatus", func() {
-				spec := readinessv1alpha1.NodeReadinessRuleSpec{
-					NodeSelector:    metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
-					ConditionPolicy: readinessv1alpha1.ConditionPolicyAnyOf,
-					Conditions: []readinessv1alpha1.ConditionRequirement{
-						{
-							Type:           "HardwareDriverReady",
-							RequiredStatus: corev1.ConditionTrue,
-							DefaultStatus:  corev1.ConditionTrue,
-						},
-					},
-				}
-
-				allErrs := webhook.validateSpec(spec, false)
-				Expect(allErrs).To(HaveLen(1))
-				Expect(allErrs[0].Field).To(Equal("spec.conditions[0].defaultStatus"))
-				Expect(allErrs[0].Type).To(Equal(field.ErrorTypeForbidden))
-			})
-
-			It("should pass validation when conditionPolicy is anyOf and defaultStatus does not equal requiredStatus", func() {
-				spec := readinessv1alpha1.NodeReadinessRuleSpec{
-					NodeSelector:    metav1.LabelSelector{MatchLabels: map[string]string{"foo": "bar"}},
-					ConditionPolicy: readinessv1alpha1.ConditionPolicyAnyOf,
-					Conditions: []readinessv1alpha1.ConditionRequirement{
-						{
-							Type:           "HardwareDriverReady",
-							RequiredStatus: corev1.ConditionTrue,
-							DefaultStatus:  corev1.ConditionFalse,
-						},
-					},
-				}
-
-				allErrs := webhook.validateSpec(spec, false)
-				Expect(allErrs).To(BeEmpty())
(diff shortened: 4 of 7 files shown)

### CODE REVIEW

File: internal/webhook/nodereadinessgaterule_webhook.go [hunk 1]
```diff
@@ -98,6 +98,18 @@ func (w *NodeReadinessRuleWebhook) validateSpec(
 		))
 	}
 
+	// validate that when conditionPolicy is anyOf, no condition's defaultStatus equals its requiredStatus
+	if spec.GetConditionPolicy() == readinessv1alpha1.ConditionPolicyAnyOf {
```
AnuragThePathak: Hi @ajaysundark I was thinking about -
What if we simplify the restriction to not allowing setting conditionPolicy and defaultStatus at the same time? Because ultimately if someone is using `conditionPolicy` it is only useful if they need `anyOf` and similarly for `defaultStatus` when it is equal to `requiredStatus`.  Hence doing so would make life easier for both users and us was my first thought.

But then when I digged deeper, I got to know that the IaC tools sometimes may try to set optional default values. So for those restricting setting a field would cause friction. Would you have any idea on this regard?

internal/webhook/nodereadinessgaterule_webhook_test.go
```diff
@@ -222,11 +222,48 @@ var _ = Describe("NodeReadinessRule Validation Webhook", func() {
...
-			Expect(allErrs).To(HaveLen(4))
+			Expect(allErrs).To(HaveLen(5))
 			Expect(allErrs[0].Field).To(Equal("spec.nodeSelector"))
 			Expect(allErrs[1].Field).To(Equal("spec.conditions[0].defaultStatus"))
 			Expect(allErrs[2].Field).To(Equal("spec.conditions[1].defaultStatus"))
 			Expect(allErrs[3].Field).To(Equal("spec.conditionPolicy"))
+			Expect(allErrs[4].Field).To(Equal("spec.conditions[1].defaultStatus"))
+		})
+
+		It("should reject when conditionPolicy is anyOf and defaultStatus equals requiredStatus", func() {
```
AnuragThePathak: For this and the next check I would encourage wrapping them in a context and would place that before the combined error check.

internal/webhook/nodereadinessgaterule_webhook_test.go
```diff
@@ -222,11 +222,48 @@ var _ = Describe("NodeReadinessRule Validation Webhook", func() {
...
+				},
+			}
+
+			allErrs := webhook.validateSpec(spec, false)
+			Expect(allErrs).To(HaveLen(1))
+			Expect(allErrs[0].Field).To(Equal("spec.conditions[0].defaultStatus"))
+			Expect(allErrs[0].Type).To(Equal(field.ErrorTypeForbidden))
+		})
+
+		It("should pass validation when conditionPolicy is anyOf and defaultStatus does not equal requiredStatus", func() {
```
AnuragThePathak: I would add two more tests

1. Early exit for update
2. ConditionPolicy is not AnyOf then happy path

internal/webhook/nodereadinessgaterule_webhook.go
ajaysundark: If I am reading your suggestion correctly as 'defaultStatus' is not supported in 'anyOf' conditionPolicy, sgtm.

Also, just a nit, for any validations, I prefer if we possibly handled in CRD XValidation, can we try?

With VAP and CRD x-validation rules improving in mainstream Kubernetes, we shouldn't be adding more and more logic into webhook. For a component like NRC, it is better when it is leaner. My hope is that someday users get away without having to manage additional webhooks for it.

ajaysundark: cc @vishnukothakapu

vishnukothakapu: @ajaysundark Yeah that's a fair point! CEL XValidation would definitely keep the webhook leaner. Let me explore if we can express this check cleanly in CEL. Arrays can be a bit tricky sometimes, but worth a try for sure.

AnuragThePathak: No @ajaysundark.

What you interpreted is
- Do not support defaultStatus when conditionPolicy "anyOf"

What I suggested goes one step ahead
- Do not allow use of defaultStatus and conditionPolicy at the same time.

My thought process behind that was thinking one would want to use conditionPolicy only when they want to use "anyOf" because "allOf" is provided by default anyway. Same goes for defaultStatus, someone would only want to use that if they want the defaultStatus to match requiredStatus by default, because it doesn't match anyway by default.

Trying to narrow down the restriction makes things harder to understand to explain to users unnecessarily is what I thought without any real benefit (as I explained above).

However, what you mentioned, also I think is a pretty much simplified version already, and looks like a great option to go with because there maybe a time when people would make a habit of adding optional conditionPolicy "allOf" in template (not expecting the same to same with defaultStatus).

@vishnukothakapu are you also aligned with the hybrid approach? by the way I have not yet thought of the CEL replacement yet.

vishnukothakapu: @AnuragThePathak Yes, I am totally aligned with the hybrid approach! Banning defaultStatus entirely when anyOf is used is a very clean rule, and allowing it for allOf prevents us from unfairly breaking standard user templates.

As for the CEL replacement, I’ve actually already been exploring it! We can implement the hybrid approach directly on the NodeReadinessRuleSpec struct using the has() macro to check for omitted fields.

The CEL rule would look like this:


> // +kubebuilder:validation:XValidation:rule="(!has(self.conditionPolicy) ? 'allOf' : self.conditionPolicy) != 'anyOf' || self.conditions.all(c, !has(c.defaultStatus))",message="defaultStatus is not supported when conditionPolicy is anyOf"


Let me know your thoughts on this CEL rule!

AnuragThePathak: That will work. But that is usually derived by AI seeing that GetDefaultStatus logic. If we try to derive one naturally we can short-circuit directly on seeing conditionPolicy is missing than trying to force default assignment
```
!has(self.conditionPolicy) || self.conditionPolicy != 'anyOf' || self.conditions.all(c, !has(c.defaultStatus))
```

vishnukothakapu: @AnuragThePathak Good catch! Honestly, I was just staring at the GetConditionPolicy() fallback logic in the Go code and tried to translate it 1-to-1 in my head. Your boolean short-circuit is way cleaner and more idiomatic for CEL.

Should I go ahead and push the update using your CEL expression and remove the webhook logic now, or should we wait for @ajaysundark to give a final thumbs-up on this hybrid approach first?

vishnukothakapu: @AnuragThePathak Good catch! Honestly, I was just staring at the GetConditionPolicy() fallback logic in the Go code and tried to translate it 1-to-1 in my head. Your boolean short-circuit is way cleaner and more idiomatic for CEL. I'll push an update using your exact expression!

vishnukothakapu: @AnuragThePathak  was looking at the GetConditionPolicy() fallback and tried to follow the same logic in CEL. Your boolean expression is much cleaner. I’ll update it.

internal/webhook/nodereadinessgaterule_webhook_test.go
AnuragThePathak, line 1: No change in this file.

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -66,6 +66,7 @@ const (
 // NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
 //
 // +kubebuilder:validation:XValidation:rule="(!has(oldSelf.conditionPolicy) ? 'allOf' : oldSelf.conditionPolicy) == (!has(self.conditionPolicy) ? 'allOf' : self.conditionPolicy)",message="conditionPolicy is immutable"
+// +kubebuilder:validation:XValidation:rule="!has(self.conditionPolicy) || self.conditionPolicy != 'anyOf' || self.conditions.all(c, !has(c.defaultStatus))",message="defaultStatus is not supported when conditionPolicy is anyOf"
```
AnuragThePathak, line 69: Considering the complexity of this validation, it is important to add tests against it.

I am thinking it may a good idea to perform full testing on it testing all possible conditions like we did for the webhook perhaps (though I haven't actually delved deeper into this myself and would leave that to you)

docs/book/src/user-guide/concepts.md
```diff
@@ -36,6 +36,11 @@ When a rule specifies multiple conditions, the `conditionPolicy` determines how
 >
 > `bootstrap-only` mode exists to verify that specific components have finished initializing. Using `anyOf` with `bootstrap-only` could lead to the node bootstrapping prematurely if just one component is ready, completely ignoring the initialization status of the others. The admission webhook enforces this restriction.
 
+> [!CAUTION]
```
AnuragThePathak: I checked how this looks in the website. Personally to me, appearance of two widgets important and caution, one after one, slightly felt off, denoting the important of importance of warning/important.

Combining both into a one may be a better idea imo
> [!IMPORTANT]
  > **API Constraints for `anyOf`**
  > The admission webhook will actively reject `anyOf` configurations in the following scenarios:
  >
  > * **Combined with `bootstrap-only` enforcement:** This mode verifies that specific components have finished initializing. Using `anyOf` could lead to the node bootstrapping prematurely if just one component is ready.
  > * **Combined with `defaultStatus`:** This creates a logical short-circuit when the condition is missing and `defaultStatus` is set to match `requiredStatus`. The condition would be considered satisfied even if the node never reports it, silently bypassing the readiness check.

AnuragThePathak (changes requested):

ajaysundark:
/retitle feat: add validation to prevent anyOf misconfiguration

ajaysundark:
@vishnukothakapu please update release notes and description to keep it in sync with the implementation

docs/book/src/user-guide/concepts.md
```diff
@@ -32,14 +32,14 @@ When a rule specifies multiple conditions, the `conditionPolicy` determines how
 - **`anyOf`**: At least ONE condition must match its `requiredStatus`. If no conditions match, the node is tainted. This is particularly useful for hardware that may be satisfied by multiple drivers, or systems with active/passive fallbacks.
 
 > [!IMPORTANT]
-> **`anyOf` is not supported with `bootstrap-only` rules.**
->
-> `bootstrap-only` mode exists to verify that specific components have finished initializing. Using `anyOf` with `bootstrap-only` could lead to the node bootstrapping prematurely if just one component is ready, completely ignoring the initialization status of the others. The admission webhook enforces this restriction.
+> **API Constraints for `anyOf`**
+> 
+> The API will actively reject `anyOf` configurations in the following scenarios:
+> - **Combined with `bootstrap-only` enforcement**: This mode verifies that specific components have finished initializing. Using `anyOf` could lead to the node bootstrapping prematurely if just one component is ready.
```
AnuragThePathak: ~~just one component is ready~~
just one condition is met

I think something like that is better to write

AnuragThePathak (commented):
Rest LGTM

AnuragThePathak (approved):
Approving it from my side. @ajaysundark PTAL at your convenience.

api/v1alpha1/nodereadinessrule_types.go
ajaysundark, line 69: Does this need api-doc regen?

docs/book/src/user-guide/concepts.md
```diff
@@ -32,9 +32,14 @@ When a rule specifies multiple conditions, the `conditionPolicy` determines how
...
->
-> `bootstrap-only` mode exists to verify that specific components have finished initializing. Using `anyOf` with `bootstrap-only` could lead to the node bootstrapping prematurely if just one component is ready, completely ignoring the initialization status of the others. The admission webhook enforces this restriction.
+> **API Constraints for `anyOf`**
+> 
+> The API will actively reject `anyOf` configurations in the following scenarios:
+> - **Combined with `bootstrap-only` enforcement**: This mode verifies that specific components have finished initializing. Using `anyOf` could lead to the node bootstrapping prematurely if just one condition is met.
+> - **Combined with `defaultStatus`**: This creates a logical short-circuit when the condition is missing and `defaultStatus` is set to match `requiredStatus`. The condition would be considered satisfied even if the node never reports it, silently bypassing the readiness check.
+
+
+
```
ajaysundark: nit: remove blank newlines

api/v1alpha1/nodereadinessrule_types.go
```diff
@@ -66,6 +66,7 @@ const (
 // NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
 //
 // +kubebuilder:validation:XValidation:rule="(!has(oldSelf.conditionPolicy) ? 'allOf' : oldSelf.conditionPolicy) == (!has(self.conditionPolicy) ? 'allOf' : self.conditionPolicy)",message="conditionPolicy is immutable"
```
ajaysundark, line 68: Thanks! Can we take a followup issue for other bootstrap-only validations (any of / defaultStatus) to also be CEL rules to be consistent?

ajaysundark (commented):
/lgtm

api/v1alpha1/nodereadinessrule_types.go
vishnukothakapu: No, I checked, crd-ref-docs doesn't parse XValidation markers, so running the regen produces no changes for this addition.

api/v1alpha1/nodereadinessrule_types.go
vishnukothakapu: Yes, absolutely! It makes sense to migrate the rest of the bootstrap-only validations to CEL so we can eventually deprecate the webhook entirely. I will open a follow-up issue to track this.

api/v1alpha1/nodereadinessrule_types.go
ajaysundark: ok, thanks for confirming!

ajaysundark (approved):

PR #421, opened 2026-08-18
feat: allow Cluster Autoscaler startup-taint prefix in NodeReadinessR…

### LINKED ISSUE #279: [FEATURE] Allow the Cluster Autoscaler startup-taint prefix in NodeReadinessRule taint keys

Cluster Autoscaler treats a taint as a startup taint in two ways: the `--startup-taint` flag, or the reserved key prefix `startup-taint.cluster-autoscaler.kubernetes.io/`, which is auto-detected without any flag.

On managed platforms the flag is not user-editable, so the reserved prefix is the only way to get startup-taint semantics:

- GKE documents the prefix as the only supported mechanism ([docs](https://docs.cloud.google.com/kubernetes-engine/docs/how-to/cluster-autoscaler#startup_taints)).
- AKS supports boot-time taints via `--node-init-taints`, but its autoscaler does not expose `--startup-taints` ([Azure/AKS#3276](https://github.com/Azure/AKS/issues/3276))

A single taint key cannot carry both prefixes. So on GKE and AKS, NRC cannot be the component that removes the startup taint.

### Concrete use case

Gating scheduling on autoscaled GPU nodes: the node pool template applies a startup taint, Node Problem Detector publishes a GPU readiness node condition, and NRC removes the taint when the condition is `True`. This works with self-managed Cluster Autoscaler (where `--startup-taint` can be set) but not on GKE or AKS.

Extend the validation to also accept the Cluster Autoscaler startup-taint prefix:

```go
// +kubebuilder:validation:XValidation:rule="self.key.startsWith('readiness.k8s.io/') || self.key.startsWith('startup-taint.cluster autoscaler.kubernetes.io/')"
```

Optionally also allow the legacy `ignore-taint.cluster-autoscaler.kubernetes.io/` prefix, which older autoscaler versions auto-detect.

- **Admin-configured prefix allowlist (controller flag).** More general, but CRD-level CEL cannot read controller flags, so apply-time validation would move to the reconciler (or webhook).
- **Cluster Autoscaler auto-detecting `readiness.k8s.io/` as a startup-taint namespace.** A good long-term change on the autoscaler side, but this requires coordination among multiple providers.

Saloni3494:
**What type of PR is this?**
> /kind feature

This PR extends the `NodeReadinessRule` CRD validation to accept Cluster Autoscaler startup-taint prefixes in the `taint.key` field. 

Previously, only the `readiness.k8s.io/` prefix was allowed. On managed platforms like GKE and AKS, the `--startup-taints` flag is often not user-editable, so the only way to get startup-taint semantics from the Cluster Autoscaler is to use its reserved prefix. By extending validation, NRC can now serve as the component that removes these boot-time taints once conditions (like GPU readiness published by NPD) are met.

Supported prefixes added:
- `startup-taint.cluster-autoscaler.kubernetes.io/`
- `ignore-taint.cluster-autoscaler.kubernetes.io/` (legacy, but useful for older CA versions)

Fixes #279 

The OpenAPI v3 validation rules on the CRD manifest and the `taint_validation_test.go` Ginkgo suite have been updated to reflect the new CEL rules.

```release-note
Extended NodeReadinessRule CRD validation to accept Cluster Autoscaler startup-taint prefixes (`startup-taint.cluster-autoscaler.kubernetes.io/` and `ignore-taint.cluster-autoscaler.kubernetes.io/`) in the `taint.key` field.

### DIFF
--- a/api/v1alpha1/nodereadinessrule_types.go
+++ b/api/v1alpha1/nodereadinessrule_types.go
@@ -101,7 +101,7 @@ type NodeReadinessRuleSpec struct {
 	// when combined with continuous enforcement mode. Prefer NoSchedule for most use cases.
 	//
 	// +required
-	// +kubebuilder:validation:XValidation:rule="self.key.startsWith('readiness.k8s.io/')",message="taint key must start with 'readiness.k8s.io/'"
+	// +kubebuilder:validation:XValidation:rule="self.key.startsWith('readiness.k8s.io/') || self.key.startsWith('startup-taint.cluster-autoscaler.kubernetes.io/') || self.key.startsWith('ignore-taint.cluster-autoscaler.kubernetes.io/')",message="taint key must start with 'readiness.k8s.io/', 'startup-taint.cluster-autoscaler.kubernetes.io/', or 'ignore-taint.cluster-autoscaler.kubernetes.io/'"
 	// +kubebuilder:validation:XValidation:rule="self.key.size() <= 253",message="taint key length must be at most 253 characters"
 	// +kubebuilder:validation:XValidation:rule="size(self.key.split('/')) == 2",message="taint key must have exactly one '/' separator (prefix/name format)"
 	// +kubebuilder:validation:XValidation:rule="size(self.key.split('/')[1]) > 0 && size(self.key.split('/')[1]) <= 63",message="taint key name part must be 1-63 characters"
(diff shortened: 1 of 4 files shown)


ajaysundark:
Hi @Saloni3494, This was already discussed in the issue, could you review the discussions here again: https://github.com/kubernetes-sigs/node-readiness-controller/issues/279 for context.


/close not-planned

PR #422, opened 2026-08-18
fix: consistent dry-run behavior between NodeController and RuleContr…

### LINKED ISSUE #54: Inconsistent Dry-Run Behavior Between Controllers
Dry-run behavior differs between `RuleController` and `NodeController`

**Current Behavior**  
- `RuleController` evaluates dry-run rules and reports what would happen(here: https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/nodereadinessrule_controller.go#L544).  
- `NodeController` skips dry-run rules entirely, producing no output when nodes change(here: https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/node_controller.go#L88).

**Problem**  
When node labels or conditions change, we will get no visibility into whether rules would match or what taint operations would occur.

Question: is that by design? If no we should evaluate in `nodeController` too when we do `dryRun=True` and update the nodeStatus maybe(here: https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/nodereadinessrule_controller.go#L350)

Saloni3494:
**What type of PR is this?**
> /kind bug

This PR fixes inconsistent Dry-Run evaluation behavior between the `RuleController` and `NodeController`.

**Previously:** When `DryRun` was set to `true`, `NodeController` would completely skip rule evaluation for any node updates (`continue`). This resulted in stale `DryRunResults` being reported in the Rule's status when a Node's conditions transitioned, as the results would only update when the Rule itself was modified.
**Now:** `NodeController` catches the dry-run rules, safely evaluates them against all applicable nodes (via `processDryRun`), and persists the refreshed `DryRunResults` summary to the `NodeReadinessRule` status, giving real-time visibility into what actions would occur without actually applying the taints.

An E2E test verifying this live-update behavior has also been added.

Fixes #54 

The update to the rule's status is done safely using `RetryOnConflict` within the `NodeController` loop.

```release-note
Fixed a bug where `DryRunResults` on `NodeReadinessRules` would not dynamically update in response to Node condition changes.

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -140,10 +140,41 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			continue
 		}
 
-		// Skip if dry run
+		// Handle dry run
 		if rule.Spec.DryRun {
-			log.Info("Skipping rule - dry run mode",
+			log.Info("Evaluating rule - dry run mode",
 				"node", node.Name, "rule", rule.Name)
+
+			nodeList := &corev1.NodeList{}
+			if err := r.List(ctx, nodeList); err != nil {
+				log.Error(err, "Failed to list nodes for dry run evaluation", "rule", rule.Name)
+				errs = append(errs, err)
+				continue
+			}
+
+			if err := r.processDryRun(ctx, rule, nodeList); err != nil {
+				log.Error(err, "Failed to process dry run for node",
+					"node", node.Name, "rule", rule.Name)
+				errs = append(errs, err)
+				continue
+			}
+
+			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				latestRule := &readinessv1alpha1.NodeReadinessRule{}
+				if err := r.Get(ctx, client.ObjectKey{Name: rule.Name}, latestRule); err != nil {
+					return err
+				}
+				patch := client.MergeFrom(latestRule.DeepCopy())
+				latestRule.Status.DryRunResults = rule.Status.DryRunResults
+				latestRule.Status.ObservedGeneration = rule.Status.ObservedGeneration
+				return r.Status().Patch(ctx, latestRule, patch)
+			})
+
+			if err != nil {
+				log.Error(err, "Failed to update rule status after dry run evaluation",
+					"node", node.Name, "rule", rule.Name)
+				errs = append(errs, err)
+			}
 			continue
 		}
 
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -543,15 +543,22 @@ status:
 			}, 10*time.Second, 2*time.Second).Should(BeTrue())
 
 			By("verifying rule has dry-run results showing what would happen")
-			Eventually(func() bool {
-				cmd := exec.Command("kubectl", "get", "nodereadinessrule", "dryrun-test-rule", "-o", "jsonpath={.status.dryRunResults}")
-				output, err := utils.Run(cmd)
-				if err != nil {
-					return false
-				}
-				// Check that dry run results exist and contain the node
-				return len(output) > 0
-			}, 30*time.Second, 2*time.Second).Should(BeTrue())
+			Eventually(func() string {
+				cmd := exec.Command("kubectl", "get", "nodereadinessrule", "dryrun-test-rule", "-o", "jsonpath={.status.dryRunResults.taintsToAdd}")
+				output, _ := utils.Run(cmd)
+				return output
+			}, 30*time.Second, 2*time.Second).Should(Equal("1"))
+
+			By("updating node condition to True")
+			err = patchNodeCondition(nodeName, "TestReady", "True")
+			Expect(err).NotTo(HaveOccurred())
+
+			By("verifying rule dry-run results update to reflect the change")
+			Eventually(func() string {
+				cmd := exec.Command("kubectl", "get", "nodereadinessrule", "dryrun-test-rule", "-o", "jsonpath={.status.dryRunResults.taintsToAdd}")
+				output, _ := utils.Run(cmd)
+				return output
+			}, 30*time.Second, 2*time.Second).Should(BeEmpty())
 
 			By("cleaning up test resources")
 			exec.Command("kubectl", "delete", "node", nodeName).Run()

### CODE REVIEW

**Comment by ajaysundark** [comment]
Hi @Saloni3494, Thanks for your PR.

I see you have created multiple large PRs against the repo without the discussion context. If you are interested in working on this / other issues, I'd welcome you to bring this to our slack channel / biweekly sync to align on the directions first.

I also recommend you to review the kubernetes contribution policy guidance here: https://github.com/kubernetes/community/blob/main/contributors/guide/pull-requests.md#ai-guidance. We love to have you work with us, but it'd help us if you engage with us and work on these PRs than one-shotting them!

**Comment by Saloni3494** [comment]
Thank you for the feedback, Ajay! I really appreciate you taking the time to review my PRs and share the contribution guidelines.

I understand your point about aligning on the direction before working on larger changes. I’ll review the Kubernetes contribution policy and will make sure to engage in the Slack channel/bisweekly syncs before taking up similar issues going forward.

I’d definitely like to continue contributing to this project and work more closely with the community. I’ll also revisit my current PRs based on the discussion and guidance from the maintainers.

Thanks again for pointing me in the right direction!

File: internal/controller/node_controller.go [hunk 1]
```diff
@@ -140,10 +140,41 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
...
+
+			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+				latestRule := &readinessv1alpha1.NodeReadinessRule{}
+				if err := r.Get(ctx, client.ObjectKey{Name: rule.Name}, latestRule); err != nil {
+					return err
+				}
+				patch := client.MergeFrom(latestRule.DeepCopy())
+				latestRule.Status.DryRunResults = rule.Status.DryRunResults
+				latestRule.Status.ObservedGeneration = rule.Status.ObservedGeneration
+				return r.Status().Patch(ctx, latestRule, patch)
```
yindia, line 170: This fires on every node event regardless of whether `DryRunResults` actually changed, so a no-op merge patch still costs a full API round trip per node event per dry-run rule.

Could we compare against `latestRule.Status.DryRunResults` first and skip the patch when they're equal? We already do this kind of no-op-write suppression in the readiness condition reporter for the same scale reason (documented here https://node-readiness-controller.sigs.k8s.io/user-guide/concepts.html#optimizing-node-status-writes).

internal/controller/node_controller.go
```diff
@@ -140,10 +140,41 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
...
 
-		// Skip if dry run
+		// Handle dry run
 		if rule.Spec.DryRun {
-			log.Info("Skipping rule - dry run mode",
+			log.Info("Evaluating rule - dry run mode",
 				"node", node.Name, "rule", rule.Name)
+
+			nodeList := &corev1.NodeList{}
+			if err := r.List(ctx, nodeList); err != nil {
```
yindia, line 149: This List runs inside the per-rule loop, so N matching dry-run rules means N cluster-wide lists per node event. It's a cache read, but controller-runtime still deep-copies every Node, and processDryRun then walks all of them. During a rolling upgrade that's O(nodes²).

Two fixes:
- move the List out of the rule loop so it runs once per node event
- pass the rule's nodeSelector via `client.MatchingLabels` so we only copy matching nodes

internal/controller/node_controller.go
```diff
@@ -140,10 +140,41 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
...
+			}
+
+			if err := r.processDryRun(ctx, rule, nodeList); err != nil {
+				log.Error(err, "Failed to process dry run for node",
+					"node", node.Name, "rule", rule.Name)
+				errs = append(errs, err)
+				continue
+			}
+
+			err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
yindia, line 162: This patches the live rule with state read from the cache. If someone flips dryRun: false while a node reconcile is in flight, we can end up writing dryRunResults back onto a now-enforcing rule and reverting observedGeneration. Status is a subresource and we use GenerationChangedPredicate, so RuleReconciler never re-runs to fix it.

Can we re-check latestRule.Spec.DryRun inside the retry closure, and skip the ObservedGeneration write here?

test/e2e/e2e_test.go
```diff
@@ -543,15 +543,22 @@ status:
...
-					return false
-				}
-				// Check that dry run results exist and contain the node
-				return len(output) > 0
-			}, 30*time.Second, 2*time.Second).Should(BeTrue())
+			Eventually(func() string {
+				cmd := exec.Command("kubectl", "get", "nodereadinessrule", "dryrun-test-rule", "-o", "jsonpath={.status.dryRunResults.taintsToAdd}")
+				output, _ := utils.Run(cmd)
+				return output
+			}, 30*time.Second, 2*time.Second).Should(Equal("1"))
```
yindia, line 550: Unless I'm misreading the marshalling, `taintsToAdd` is a `*int32` with `omitempty`, which only drops nil, and `processDryRun` always sets the pointer. So a zero count still serializes as `"taintsToAdd": 0` and jsonpath returns `"0"` here, not empty.

  ```suggestion
                 }, 30*time.Second, 2*time.Second).Should(Equal("0"))
  ```

Optional, but `summary` might be a stronger signal than a count dropping to zero, it goes from `"would add 1 taints"` to `"No changes needed"`

yindia (changes requested):
Left a few comments. The bigger thing is direction.

There are two shapes in #54. This takes the minimal one, keep `processDryRun` separate and call it from the NodeReconciler. The other, which AvineshTripathi and farazmd each landed on independently, unifies them, let `evaluateRuleForNode` handle both modes, gate only the taint write on `dryRun`, and update stats by delta.

That second shape also gets you `nodeEvaluations` for dry-run rules, drops the duplicated aggregation logic, and is O(1) per node event instead of a full node scan.

Nothing was ever decided there though. Could you drop a note on #54 with which shape you'd like to take? Would rather agree that up front than have you rework this twice.

Saloni3494:
@yindia Thanks for the detailed review and for laying out the two architectural options so clearly.

I agree that the second approach, unifying the logic under evaluateRuleForNode to handle both modes and updating the stats by delta, is the stronger direction. The $O(1)$ scaling per node event and the removal of duplicated aggregation logic make a lot more sense than doing full node scans.

Before I start rewriting the PR, I wanted to check if we have consensus to officially proceed with this unified shape? I'm ready to refactor the code to match this direction once everyone is aligned.

yindia:
Before you start the rewrite, let's move this to Slack or the issue thread so the maintainers can chime in. Once everyone agrees on the approach, go ahead with the refactor.

Saloni3494:
@yindia Sounds good! I've moved the architectural discussion over to [Issue #54](https://github.com/kubernetes-sigs/node-readiness-controller/issues/54) for the maintainers to weigh in. I'll hold off on the refactor until we reach a consensus there.

PR #423, opened 2026-08-18
feat: add suspend field to NodeReadinessRuleSpec

### LINKED ISSUE #335: [FEATURE] Add a suspend field to `NodeReadinessRuleSpec` to pause rule evaluation

Saloni3494:
**What type of PR is this?**
> /kind feature

This PR adds a `suspend` boolean field to the `NodeReadinessRuleSpec`. 

Currently, there is no way to temporarily disable or pause the evaluation of a `NodeReadinessRule`. If an administrator needs to perform maintenance on a cluster, test a change, or debug an issue without the controller constantly evaluating and applying/removing taints, their only option is to delete the rule entirely (which un-taints all nodes) or switch it to dry-run mode (which still consumes controller cycles to compute node evaluations).

By setting `suspend: true`, the controllers will simply skip the reconciliation loop for that rule, making no changes to the cluster or the rule's status. The rule essentially becomes frozen in its current state.

Fixes #335

- A check for `rule.Spec.Suspend` was added at the top of the `RuleReconciler` and within the `NodeReconciler` rule loop.
- The CRDs were regenerated and a new E2E test was added to verify that a suspended rule does not take action on unsatisfied node conditions.

```release-note
Added a `suspend` field to `NodeReadinessRuleSpec`. When set to true, the controller will temporarily bypass all evaluations and taint modifications for the rule, freezing its current cluster state.

### DIFF
--- a/api/v1alpha1/nodereadinessrule_types.go
+++ b/api/v1alpha1/nodereadinessrule_types.go
@@ -128,6 +128,12 @@ type NodeReadinessRuleSpec struct {
 	// +optional
 	ConditionPolicy ConditionPolicy `json:"conditionPolicy,omitempty"` // Use GetConditionPolicy() for safe access; field may be empty even when allOf applies.
 
+	// suspend when set to true, temporarily pauses the evaluation of the rule.
+	// The controller will skip processing this rule and make no changes to node taints or rule status.
+	//
+	// +optional
+	Suspend bool `json:"suspend,omitempty"` //nolint:kubeapilinter
+
 	// dryRun when set to true, The controller will evaluate Node conditions and log intended taint modifications
 	// without persisting changes to the cluster. Proposed actions are reflected in the resource status.
 	//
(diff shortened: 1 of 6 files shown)


ajaysundark:
Duplicate Of #352 

/close not-planned

PR #425, opened 2026-08-18
fix: reject anyOf conditionPolicy when defaultStatus equals requiredS…

### LINKED ISSUE #414: [FEATURE] anyOf conditionPolicy with `defaultStatus == requiredStatus` silently short-circuits rule

Saloni3494:
**What type of PR is this?**
> /kind bug

This PR prevents a silent failure mode where `conditionPolicy: anyOf` rules could short-circuit themselves due to a misconfiguration in `defaultStatus`.

As identified in #414, if an `anyOf` rule has a condition where `defaultStatus == requiredStatus`, the rule evaluates to "Satisfied" the moment the node is evaluated (assuming the condition is not present on the node, which is typical at startup). This causes the controller to immediately skip taint application, acting as a complete no-op without any logged warnings or errors.

This PR adds a CEL validation rule to `NodeReadinessRuleSpec` that rejects the CRD at the API server level if `conditionPolicy` is `anyOf` and any condition's `defaultStatus` equals its `requiredStatus`.

Fixes #414

- Added an E2E test to verify that the API server correctly rejects the misconfigured manifest.

```release-note
Added CEL validation to reject NodeReadinessRules with conditionPolicy `anyOf` when a condition's `defaultStatus` matches its `requiredStatus`, preventing silent misconfigurations.

### DIFF
--- a/api/v1alpha1/nodereadinessrule_types.go
+++ b/api/v1alpha1/nodereadinessrule_types.go
@@ -66,6 +66,7 @@ const (
 // NodeReadinessRuleSpec defines the desired state of NodeReadinessRule.
 //
 // +kubebuilder:validation:XValidation:rule="(!has(oldSelf.conditionPolicy) ? 'allOf' : oldSelf.conditionPolicy) == (!has(self.conditionPolicy) ? 'allOf' : self.conditionPolicy)",message="conditionPolicy is immutable"
+// +kubebuilder:validation:XValidation:rule="!has(self.conditionPolicy) || self.conditionPolicy != 'anyOf' || !self.conditions.exists(c, has(c.defaultStatus) && c.defaultStatus == c.requiredStatus)",message="defaultStatus cannot equal requiredStatus when conditionPolicy is anyOf"
 type NodeReadinessRuleSpec struct {
 	// conditions contains a list of the Node conditions that defines the specific
 	// criteria that must be met for taints to be managed on the target Node.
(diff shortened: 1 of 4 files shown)


ajaysundark:
Duplicate Of #418 

/close not-planned

PR #430, opened 2026-08-19
[#376] fix: clear stale FailedNodes entry in node reconciler after successful evaluation

### LINKED ISSUE #376: [BUG] NodeReconciler permanently leaks transient errors into Status.FailedNodes

tejassinghbhati:
## Summary

`processNodeAgainstAllRules` permanently re-asserts a stale `Status.FailedNodes` entry for a node on every reconcile, even after `evaluateRuleForNode` succeeds, until the rule is manually edited.

This is the unfixed mirror of the fix that landed for `processAllNodesForRule` — that fix only touched `nodereadinessrule_controller.go`. The node reconciler path in `node_controller.go` was never updated with the equivalent clearing logic.

## Root cause

The status patch loop rebuilds `FailedNodes` in two passes. Pass 1 correctly drops this node's current entry from the API server's state. Pass 2 unconditionally re-appends whatever the **cached** `rule.Status.FailedNodes` holds for the node — regardless of whether the evaluation just succeeded. Because a successful evaluation never clears the cached copy's `FailedNodes` field, pass 2 actively re-asserts the stale entry on every node reconcile.

## Fix

Gate pass 2 on `evalErr != nil`, matching the `else` branch already present in `processAllNodesForRule`:

```go
// Re-add this node's failure entry only when evaluation just
// failed. On success we leave it absent so that a stale error
// recorded by a previous transient failure is cleared.
if evalErr != nil {
    for _, failure := range rule.Status.FailedNodes {
        if failure.NodeName == node.Name {
            updatedFailedNodes = append(updatedFailedNodes, failure)
            break
        }
    }
}
```

## Testing

Added a regression test in `node_controller_test.go` using a fake client with `WithStatusSubresource`:

1. Seed a `FailedNodes` entry for the node to simulate a prior transient error.
2. Run `processNodeAgainstAllRules` with the node satisfying all conditions (evaluation succeeds).
3. Assert the stale `FailedNodes` entry is absent from the patched rule status.

`go build ./...` passes cleanly. The controller test suite requires `envtest` binaries (etcd / kube-apiserver) which are unavailable locally on Windows; CI runs on `ubuntu-latest` where they are present.

## Related Issue

Fixes kubernetes-sigs/node-readiness-controller#376

## Type of Change

/kind bug

## Checklist
- [x] `make test` passes (CI; envtest not available locally on Windows)
- [x] `go build ./...` passes

```release-note
NONE
```

### DIFF
--- a/internal/controller/node_controller.go
+++ b/internal/controller/node_controller.go
@@ -152,12 +152,13 @@ func (r *RuleReadinessController) processNodeAgainstAllRules(ctx context.Context
 			"rule", rule.Name,
 			"ruleResourceVersion", rule.ResourceVersion)
 
-		if err := r.evaluateRuleForNode(ctx, rule, node); err != nil {
-			log.Error(err, "Failed to evaluate rule for node",
+		evalErr := r.evaluateRuleForNode(ctx, rule, node)
+		if evalErr != nil {
+			log.Error(evalErr, "Failed to evaluate rule for node",
 				"node", node.Name, "rule", rule.Name)
 			// Continue with other rules even if one fails
-			r.recordNodeFailure(rule, node.Name, "EvaluationError", err.Error())
-			errs = append(errs, err)
+			r.recordNodeFailure(rule, node.Name, "EvaluationError", evalErr.Error())
+			errs = append(errs, evalErr)
 			metrics.Failures.WithLabelValues(rule.Name, string(metrics.FailureReasonEvaluationError)).Inc()
 		}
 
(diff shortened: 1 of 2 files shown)


ajaysundark:
Duplicate Of #381 

/close

PR #431, opened 2026-08-19
feat: add blocked nodes metric to scrape-time collector

rawadhossain:
## Description

Adds `node_readiness_blocked_nodes{rule, condition}` for tracking the number of currently-held nodes blocked by each unsatisfied condition for each `NodeReadinessRule`. Ref. [New Scrape-Time Collector Surface](https://github.com/ajaysundark/node-readiness-controller/blob/9244f4eef28234321fecc988435b846fd3c12507/docs/observability-design.md#4-new-scrape-time-collector-surface)

The metric is collected directly from the controller runtime cache on each Prometheus scrape, sharing the same Node snapshot with `node_readiness_rule_nodes`.

### Changes
- Added `node_readiness_blocked_nodes` metric with `rule` and `condition` labels.
- Added `ListBlockedNodes` to calculate blocked nodes from the live Node state.
- Uses the rule's `DefaultStatus` when a condition is missing from the Node, same as the controller.
- Shares the `ListNodes` with `node_readiness_rule_nodes` during each scrape.

## Type of Change
/kind feature

## Testing
- Verified against a live cluster, including live condition changes and `DefaultStatus` behavior.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/docs/book/src/operations/monitoring.md
+++ b/docs/book/src/operations/monitoring.md
@@ -106,6 +106,25 @@ Number of nodes currently held or released by each `NodeReadinessRule`, collecte
 | `rule` | `NodeReadinessRule` name | Any non-dry-run rule name with a valid selector |
 | `state` | Whether matching nodes are still tainted by the rule or have had the taint removed | `held`, `released` |
 
+### `node_readiness_blocked_nodes`
+
+*Available starting from the v0.6.0 release.*
+
+Number of currently-held nodes blocked by each unsatisfied condition, per `NodeReadinessRule`, collected at scrape time from the controller cache.
+
+| Property | Value |
+| --- | --- |
+| Type | `gauge` |
+| Labels | `rule`, `condition` |
+| Recorded when | Computed on each Prometheus scrape from the cached node list |
+
+#### Labels
+
+| Label | Description | Values |
+| --- | --- | --- |
+| `rule` | `NodeReadinessRule` name | Any non-dry-run rule name with a valid selector |
+| `condition` | Condition type declared in `spec.conditions` | Any condition type declared by the rule |
+
 ### `node_readiness_bootstrap_completed_total`
 
 Total number of nodes that have completed bootstrap.
--- a/docs/book/src/operations/monitoring.md
+++ b/docs/book/src/operations/monitoring.md
@@ -110,7 +110,7 @@ Number of nodes currently held or released by each `NodeReadinessRule`, collecte
 
 *Available starting from the v0.6.0 release.*
 
-Number of currently-held nodes blocked by each unsatisfied condition, per `NodeReadinessRule`, collected at scrape time from the controller cache.
+Number of currently-held nodes against blocking conditions per `NodeReadinessRule`.
 
 | Property | Value |
 | --- | --- |
@@ -122,7 +122,7 @@ Number of currently-held nodes blocked by each unsatisfied condition, per `NodeR
 
 | Label | Description | Values |
 | --- | --- | --- |
-| `rule` | `NodeReadinessRule` name | Any non-dry-run rule name with a valid selector |
+| `rule` | `NodeReadinessRule` name | Any non-dry-run rule name |
 | `condition` | Condition type declared in `spec.conditions` | Any condition type declared by the rule |
 
 ### `node_readiness_bootstrap_completed_total`
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -531,15 +531,19 @@ func (r *RuleReadinessController) getApplicableRulesForNode(ctx context.Context,
 	return applicableRules
 }
 
-// ListRuleNodeStates returns the number of held and released nodes for each rule.
-func (r *RuleReadinessController) ListRuleNodeStates(ctx context.Context) (map[string]metrics.RuleNodeCounts, error) {
-	ruleList := &readinessv1alpha1.NodeReadinessRuleList{}
-	if err := r.List(ctx, ruleList); err != nil {
-		return nil, err
-	}
-
+// ListNodes returns the current list of Nodes.
+func (r *RuleReadinessController) ListNodes(ctx context.Context) ([]corev1.Node, error) {
 	nodeList := &corev1.NodeList{}
 	if err := r.List(ctx, nodeList); err != nil {
+		return nil, err
+	}
+	return nodeList.Items, nil
+}
+
+// ListRuleNodeStates returns the number of held and released nodes for each rule.
+func (r *RuleReadinessController) ListRuleNodeStates(ctx context.Context, nodes []corev1.Node) (map[string]metrics.RuleNodeCounts, error) {
+	ruleList := &readinessv1alpha1.NodeReadinessRuleList{}
+	if err := r.List(ctx, ruleList); err != nil {
 		return nil, err
 	}
 
@@ -560,8 +564,8 @@ func (r *RuleReadinessController) ListRuleNodeStates(ctx context.Context) (map[s
 		}
 
 		rc := metrics.RuleNodeCounts{}
-		for i := range nodeList.Items {
-			node := &nodeList.Items[i]
+		for i := range nodes {
+			node := &nodes[i]
 			if !selector.Matches(labels.Set(node.Labels)) {
 				continue
 			}
@@ -575,6 +579,52 @@ func (r *RuleReadinessController) ListRuleNodeStates(ctx context.Context) (map[s
 	}
 
 	return counts, nil
+}
+
+// ListBlockedNodes returns the number of blocked nodes for each rule and unsatisfied condition.
+func (r *RuleReadinessController) ListBlockedNodes(ctx context.Context, nodes []corev1.Node) (map[string]metrics.RuleBlockedConditions, error) {
+	ruleList := &readinessv1alpha1.NodeReadinessRuleList{}
+	if err := r.List(ctx, ruleList); err != nil {
+		return nil, err
+	}
+
+	log := ctrl.LoggerFrom(ctx)
+
+	result := make(map[string]metrics.RuleBlockedConditions, len(ruleList.Items))
+	for i := range ruleList.Items {
+		rule := &ruleList.Items[i]
+		if rule.Spec.DryRun {
+			continue
+		}
+
+		selector, err := metav1.LabelSelectorAsSelector(&rule.Spec.NodeSelector)
+		if err != nil {
+			log.V(2).Info("Invalid node selector for rule", "rule", rule.Name, "error", err)
+			continue
+		}
+
+		counts := make(metrics.RuleBlockedConditions, len(rule.Spec.Conditions))
+		for _, cond := range rule.Spec.Conditions {
+			counts[cond.Type] = 0
+		}
+
+		for i := range nodes {
+			node := &nodes[i]
+			if !selector.Matches(labels.Set(node.Labels)) || !r.hasTaintBySpec(node, rule.Spec.Taint) {
+				continue
+			}
+			for _, cond := range rule.Spec.Conditions {
+				effectiveStatus, _ := r.getConditionStatus(node, cond.Type, cond.GetDefaultStatus())
+				if effectiveStatus != cond.RequiredStatus {
+					counts[cond.Type]++
+				}
+			}
+		}
+
+		result[rule.Name] = counts
+	}
+
+	return result, nil
 }
 
 // ruleAppliesTo checks if a rule applies to a node.
--- a/internal/controller/nodereadinessrule_controller.go
+++ b/internal/controller/nodereadinessrule_controller.go
@@ -540,42 +540,72 @@ func (r *RuleReadinessController) ListNodes(ctx context.Context) ([]corev1.Node,
 	return nodeList.Items, nil
 }
 
-// ListRuleNodeStates returns the number of held and released nodes for each rule.
-func (r *RuleReadinessController) ListRuleNodeStates(ctx context.Context, nodes []corev1.Node) (map[string]metrics.RuleNodeCounts, error) {
+// forEachRuleNode applies callbacks to nodes matching each rule.
+func (r *RuleReadinessController) forEachRuleNode(
+	ctx context.Context,
+	nodes []corev1.Node,
+	skipRule func(rule *readinessv1alpha1.NodeReadinessRule) bool,
+	onRule func(rule *readinessv1alpha1.NodeReadinessRule),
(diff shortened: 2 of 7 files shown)


internal/controller/nodereadinessrule_controller.go
```diff
@@ -577,6 +581,52 @@ func (r *RuleReadinessController) ListRuleNodeStates(ctx context.Context) (map[s
 	return counts, nil
 }
 
+// ListBlockedNodes returns the number of blocked nodes for each rule and unsatisfied condition.
+func (r *RuleReadinessController) ListBlockedNodes(ctx context.Context, nodes []corev1.Node) (map[string]metrics.RuleBlockedConditions, error) {
```
AvineshTripathi: this func is very much similar to `ListRuleNodeStates` in terms of execution. Can we make a util func that gives you execution flow and use it in both places?


for e.g. 

```
func (r *RuleReadinessController) measureRuleNodes(
    ctx context.Context,
    nodes []corev1.Node,
    measureFn func(rule *readinessv1alpha1.NodeReadinessRule, node *corev1.Node) error,
) error {
    // Fetch rules, iterate, parse selectors, filter nodes
    // Call measureFn(rule, node) for each held node
    // measureFn does the actual counting/filtering
}
```

internal/controller/nodereadinessrule_controller.go
```diff
@@ -577,6 +581,52 @@ func (r *RuleReadinessController) ListRuleNodeStates(ctx context.Context) (map[s
...
+	log := ctrl.LoggerFrom(ctx)
+
+	result := make(map[string]metrics.RuleBlockedConditions, len(ruleList.Items))
+	for i := range ruleList.Items {
+		rule := &ruleList.Items[i]
+		if rule.Spec.DryRun {
+			continue
+		}
+
+		selector, err := metav1.LabelSelectorAsSelector(&rule.Spec.NodeSelector)
```
AvineshTripathi: we can reuse `ruleAppliesTo` here. I see we add extra `!r.hasTaintBySpec(node, rule.Spec.Taint)` check so probably opportunity to create another util that uses runAppliesTo and hasTaint

internal/metrics/collector.go
```diff
@@ -45,33 +66,58 @@ var ruleNodesDesc = prometheus.NewDesc(
...
 	defer cancel()
 
-	counts, err := c.lister.ListRuleNodeStates(ctx)
+	nodes, err := c.lister.ListNodes(ctx)
+	if err != nil {
+		ctrl.Log.V(2).Info("Failed to list nodes", "error", err)
+		return
+	}
+
+	counts, err := c.lister.ListRuleNodeStates(ctx, nodes)
```
AvineshTripathi: missing return in case of err, this could be expected behavior but there is inconsistency. `ListBlockedNodes` below returns in case of err

internal/metrics/collector.go
```diff
@@ -45,33 +66,58 @@ var ruleNodesDesc = prometheus.NewDesc(
...
 
 // Describe implements prometheus.Collector.
 func (c *ReadinessCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- ruleNodesDesc
+	ch <- blockedNodesDesc
 }
 
 // Collect implements prometheus.Collector.
 func (c *ReadinessCollector) Collect(ch chan<- prometheus.Metric) {
 	ctx, cancel := context.WithTimeout(context.Background(), collectTimeout)
```
AvineshTripathi, line 101: 5sec to do `ListNodes`, `ListRuleNodeStates` and `ListBlockedNodes` can feel little less. wdyt?

internal/metrics/collector.go
```diff
@@ -21,12 +21,18 @@ import (
...
 	"github.com/prometheus/client_golang/prometheus"
+	corev1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
 )
 
 // collectTimeout limits how long a scrape can wait for cached data.
 const collectTimeout = 5 * time.Second
 
+// NodeLister lists Nodes for the collector.
+type NodeLister interface {
```
AvineshTripathi, line 34: more like a question not a blocker: why are we creating individual interfaces when at the end we are using only `ReadinessLister`.  Is this for testing purposes?

AvineshTripathi (commented):
did my initial round of reviewing, please have a look at the comments!

internal/controller/nodereadinessrule_controller.go
rawadhossain: Done. Added a shared util forEachRuleNode and reused it in both functions.

internal/controller/nodereadinessrule_controller.go
rawadhossain: yeah, right. Added `parseNodeSelector` so the selector parsing is shared and added `ruleAppliesToWithTaint` to handle the rule + taint check together.

internal/metrics/collector.go
rawadhossain: made error handling consistent for both metrics now. Neither returns on error, we log and continue as the two metrics are independent, so a failure in one shouldn't prevent the other from being collected.

internal/metrics/collector.go
rawadhossain: I wasn't fully sure about this. Kept it at 5s since it's below Prometheus's default scrape timeout (10s), so we can fail and log before the scrape times out. I also tested it with larger node counts and it seemed fine. Do you think it'd be better to increase?

internal/metrics/collector.go
rawadhossain: yeah, mainly for testing, so each part can be mocked independently

docs/book/src/operations/monitoring.md
```diff
@@ -106,6 +106,25 @@ Number of nodes currently held or released by each `NodeReadinessRule`, collecte
 | `rule` | `NodeReadinessRule` name | Any non-dry-run rule name with a valid selector |
 | `state` | Whether matching nodes are still tainted by the rule or have had the taint removed | `held`, `released` |
 
+### `node_readiness_blocked_nodes`
+
+*Available starting from the v0.6.0 release.*
+
+Number of currently-held nodes blocked by each unsatisfied condition, per `NodeReadinessRule`, collected at scrape time from the controller cache.
```
ajaysundark:
```suggestion
Number of currently-held nodes against blocking conditions per `NodeReadinessRule`.
```
nit: you are repeating what is captured below

docs/book/src/operations/monitoring.md
```diff
@@ -106,6 +106,25 @@ Number of nodes currently held or released by each `NodeReadinessRule`, collecte
...
+| --- | --- |
+| Type | `gauge` |
+| Labels | `rule`, `condition` |
+| Recorded when | Computed on each Prometheus scrape from the cached node list |
+
+#### Labels
+
+| Label | Description | Values |
+| --- | --- | --- |
+| `rule` | `NodeReadinessRule` name | Any non-dry-run rule name with a valid selector |
```
ajaysundark:
```suggestion
| `rule` | `NodeReadinessRule` name | Any non-dry-run rule name |
```

internal/metrics/collector.go
```diff
@@ -45,33 +66,57 @@ var ruleNodesDesc = prometheus.NewDesc(
...
+	ch <- blockedNodesDesc
 }
 
 // Collect implements prometheus.Collector.
 func (c *ReadinessCollector) Collect(ch chan<- prometheus.Metric) {
 	ctx, cancel := context.WithTimeout(context.Background(), collectTimeout)
 	defer cancel()
 
-	counts, err := c.lister.ListRuleNodeStates(ctx)
+	nodes, err := c.lister.ListNodes(ctx)
```
ajaysundark, line 104: I have a high level suggestion. Why does this collector list nodes twice? If the node list results have different snapshots (in the single scrape) between the two list time. This could result confusing results

ajaysundark: Can your forEachRuleNode be a single list computation for all needs?

DsThakurRawat:
Looked through the collector changes and noticed two small items in `internal/controller/nodereadinessrule_controller.go`:

1. **Double pass on scrape:** `Collect()` in `internal/metrics/collector.go` calls `ListRuleNodeStates` and then `ListBlockedNodes`. Both functions call `r.List(ctx, ruleList)` and run `forEachRuleNode` over the full node slice. On large node counts, this does two full O(rules * nodes) passes per scrape. We could combine them into a single pass that collects both `RuleNodeCounts` and `RuleBlockedConditions` together.

2. **Deletion timestamp filter mismatch:** In `ListBlockedNodes` (line 580), rules undergoing deletion are skipped (`!rule.DeletionTimestamp.IsZero()`), but `ListRuleNodeStates` (line 565) includes them (`return false`). While a deleted rule is waiting on finalizer cleanup, `node_readiness_rule_nodes` will continue emitting held/released gauges for that rule, while `node_readiness_blocked_nodes` drops all condition metrics for it. We should keep the deletion filter consistent across both metric collectors.

internal/metrics/collector.go
rawadhossain: yeah missed on that one, actually it wasn't the node list, it was rule list being fetched twice, which had the same risk you described. Fixed that now so the rules are fetched once per scrape and the same snapshot is passed to both computations.

and for forEachRuleNode, I did try combining the two `forEachRuleNode` passes into a single computation and benchmarked it with different node/rule counts, improvement was pretty small, but it made the deletion case a bit harder to handle correctly. Like a deleting rule still needs to count for held/released nodes but not for blocked nodes. So kept the two computations separate for now to keep logic simpler.

rawadhossain:
> Looked through the collector changes and noticed two small items in `internal/controller/nodereadinessrule_controller.go`:
> 
> 1. **Double pass on scrape:** `Collect()` in `internal/metrics/collector.go` calls `ListRuleNodeStates` and then `ListBlockedNodes`. Both functions call `r.List(ctx, ruleList)` and run `forEachRuleNode` over the full node slice. On large node counts, this does two full O(rules * nodes) passes per scrape. We could combine them into a single pass that collects both `RuleNodeCounts` and `RuleBlockedConditions` together.
> 2. **Deletion timestamp filter mismatch:** In `ListBlockedNodes` (line 580), rules undergoing deletion are skipped (`!rule.DeletionTimestamp.IsZero()`), but `ListRuleNodeStates` (line 565) includes them (`return false`). While a deleted rule is waiting on finalizer cleanup, `node_readiness_rule_nodes` will continue emitting held/released gauges for that rule, while `node_readiness_blocked_nodes` drops all condition metrics for it. We should keep the deletion filter consistent across both metric collectors.

1. already addressed.
2. good catch, fixed the inconsistency.

internal/controller/nodereadinessrule_controller.go
```diff
@@ -738,12 +829,13 @@ func (r *RuleReadinessController) cleanupTaintsForRule(ctx context.Context, rule
 
 	var errors []string
 	for _, node := range nodeList.Items {
-		if !r.ruleAppliesTo(ctx, rule, &node) {
+		applies, held := r.ruleAppliesToWithTaint(ctx, rule, &node)
```
AvineshTripathi: Internally this parses nodeselector for every node which can be avoided.

internal/metrics/collector.go
```diff
@@ -45,33 +74,63 @@ var ruleNodesDesc = prometheus.NewDesc(
...
+	rules, err := c.lister.ListRules(ctx)
+	if err != nil {
+		ctrl.Log.V(2).Info("Failed to list rules", "error", err)
 		return
 	}
 
-	for rule, rc := range counts {
-		ch <- prometheus.MustNewConstMetric(ruleNodesDesc, prometheus.GaugeValue, rc.Held, rule, string(RuleNodeStateHeld))
-		ch <- prometheus.MustNewConstMetric(ruleNodesDesc, prometheus.GaugeValue, rc.Released, rule, string(RuleNodeStateReleased))
+	counts, err := c.lister.ListRuleNodeStates(ctx, nodes, rules)
```
AvineshTripathi: I think we should rename counts to `nodeStatesByRule` or something intuitive

internal/controller/nodereadinessrule_controller.go
```diff
@@ -531,57 +531,139 @@ func (r *RuleReadinessController) getApplicableRulesForNode(ctx context.Context,
...
+	if err := r.List(ctx, nodeList); err != nil {
 		return nil, err
 	}
+	return nodeList.Items, nil
+}
 
-	nodeList := &corev1.NodeList{}
-	if err := r.List(ctx, nodeList); err != nil {
+// ListRules returns the current list of NodeReadinessRules.
+func (r *RuleReadinessController) ListRules(ctx context.Context) ([]*readinessv1alpha1.NodeReadinessRule, error) {
```
AvineshTripathi, line 544: why `ListRule` is returning `[]*readinessv1alpha1.NodeReadinessRule` while ListNode returns `[]corev1.NodeList`. One is a pointer while other is a copy

AvineshTripathi (commented):
I tried this locally, didn't see any issue there for now. Left some minor comments. PHAL

internal/controller/nodereadinessrule_controller.go
rawadhossain: yeah, actually followed the existing pattern. Rules are used as pointers everywhere else in this file, so `ListRules` follows the same pattern. anf nodes are only handled as values here, so I left this as is to stay consistent with the existing usage.

internal/metrics/collector.go
rawadhossain: yes renamed to `nodeStatesByRule`

internal/controller/nodereadinessrule_controller.go
rawadhossain: right, didn't notice that. Moved the parsing outside the loop so it only happens once per rule now, and removed `ruleAppliesToWithTaint` since it's not needed anymore

rawadhossain:
cc @ajaysundark for final look

internal/controller/nodereadinessrule_controller.go
ajaysundark: yeah, we dont follow a consistent convention through the code base. it's harder to guard it over time.

ListNodes return []corev1.Node  because it just returns what's returned by nodeList.Items. if you want to convert from/to ptr/array to save you refactor trouble on this already decent sized PR, that's fine for now I guess

ajaysundark:
/approve

can you get lgtm from @AvineshTripathi as he had more depth review on this already.

internal/metrics/collector.go
AvineshTripathi: for now it's fine we will see if we find some problem in our scale testing

AvineshTripathi:
/lgtm


I dont see any blocker here

PR #436, opened 2026-08-19
[#435] fix: webhook taint conflict check incorrectly rejects dry-run rules

### LINKED ISSUE #435: [BUG] Webhook taint conflict check incorrectly rejects dry-run rules
### What happened?

When the validation webhook is enabled, it blocks creation of a `dryRun: true` rule whenever an existing **enforcement** rule already manages the same taint key on overlapping nodes — even though a dry-run rule never writes any taint to any node and therefore presents no actual conflict.

The inverse is also blocked: promoting an existing dry-run rule to enforcement by creating a new enforcement rule with the same taint key is rejected if a dry-run rule for that key already exists.

### Root cause

`validateTaintConflicts` in `nodereadinessgaterule_webhook.go` iterates all existing rules and flags any pair sharing the same `taint.key` + `taint.effect` on overlapping node selectors. It does not check `spec.dryRun` on either side:

```go
// existing code — no dryRun guard
if existingRule.Spec.Taint.Key == rule.Spec.Taint.Key &&
    existingRule.Spec.Taint.Effect == rule.Spec.Taint.Effect {
    if w.nodeSelectorsOverlap(...) {
        allErrs = append(allErrs, field.Invalid(...))
    }
}
```

### Steps to reproduce

1. Create an enforcement rule:

```yaml
kind: NodeReadinessRule
metadata:
  name: enforcement-rule
spec:
  enforcementMode: continuous
  nodeSelector:
    matchLabels:
      node-role.kubernetes.io/worker: ""
  conditions:
    - type: Ready
      requiredStatus: "True"
  taint:
    key: readiness.k8s.io/node-ready
    effect: NoSchedule
```

2. Attempt to create a dry-run preview rule for the same taint key (e.g. to evaluate a condition change):

```yaml
kind: NodeReadinessRule
metadata:
  name: preview-rule
spec:
  dryRun: true
  enforcementMode: continuous
  nodeSelector:
    matchLabels:
      node-role.kubernetes.io/worker: ""
  conditions:
    - type: DiskPressure
      requiredStatus: "False"
  taint:
    key: readiness.k8s.io/node-ready
    effect: NoSchedule
```

3. The webhook rejects the request:

```
Error: admission webhook "vnodereadinessrule.readiness.node.x-k8s.io" denied the request:
  validation failed: [spec.taint.key: Invalid value: "readiness.k8s.io/node-ready":
  conflicts with existing rule 'enforcement-rule' ...]
```

### Expected behavior

A `dryRun: true` rule must never be rejected for a taint key conflict because it does not apply, remove, or modify any taint. The conflict guard exists to prevent two enforcement rules from fighting over the same taint; that concern does not apply to dry-run rules.

### Impact

This makes the primary `dryRun` use case — previewing the impact of a rule against a taint key already in use — impossible when the webhook is enabled.

### Controller version / image tag

Reproduced on `main` at `e259c3f`.

### Kubernetes version

Not version-specific; this is webhook validation logic.

tejassinghbhati:
## Summary

The validation webhook blocks creation of a `dryRun: true` rule whenever an enforcement rule already manages the same taint key on overlapping nodes. Because dry-run rules never write taints to nodes, they cannot produce the concurrent-update conflict the guard is designed to prevent. The rejection is a false positive that makes the primary use case for `dryRun`, previewing the impact of a new rule against a taint key already in active use is impossible when the webhook is enabled.

## Root cause

`validateTaintConflicts` iterates all existing rules and flags any pair sharing `taint.key` + `taint.effect` on overlapping selectors. It does not check `spec.dryRun` on either side, so it treats dry-run rules as if they were enforcement rules.

## Fix

Add an early `continue` inside `validateTaintConflicts` when either the incoming or the existing rule has `Spec.DryRun` set:

```go
// Dry-run rules never write taints, so they cannot produce a real // conflict regardless of taint key or selector overlap.
if rule.Spec.DryRun || existingRule.Spec.DryRun {
    continue
}
```

## Tests

Two regression tests added to `nodereadinessgaterule_webhook_test.go`:

1. A `dryRun: true` rule with the same taint key as an existing enforcement rule on overlapping nodes is accepted.
2. An enforcement rule is accepted when a `dryRun: true` rule already exists for the same taint key on overlapping nodes.

`go test ./internal/webhook/...` — 41 passed, 0 failed.

## Related issue

Fixes kubernetes-sigs/node-readiness-controller#435

## Type of change

/kind bug

## Checklist
- [x] `go build ./...` passes
- [x] `go test ./internal/webhook/...` passes (41 tests, no envtest required)

```release-note
NONE
```

### DIFF
--- a/internal/webhook/nodereadinessgaterule_webhook.go
+++ b/internal/webhook/nodereadinessgaterule_webhook.go
@@ -125,6 +125,12 @@ func (w *NodeReadinessRuleWebhook) validateTaintConflicts(ctx context.Context, r
 			continue
 		}
 
+		// Dry-run rules never write taints, so they cannot produce a real
+		// conflict regardless of taint key or selector overlap.
+		if rule.Spec.DryRun || existingRule.Spec.DryRun {
+			continue
+		}
+
 		// Check for same taint key and effect
 		if existingRule.Spec.Taint.Key == rule.Spec.Taint.Key &&
 			existingRule.Spec.Taint.Effect == rule.Spec.Taint.Effect {
--- a/internal/webhook/nodereadinessgaterule_webhook.go
+++ b/internal/webhook/nodereadinessgaterule_webhook.go
@@ -97,6 +97,12 @@ func (w *NodeReadinessRuleWebhook) validateTaintConflicts(ctx context.Context, r
 			continue
 		}
 
+		// Dry-run rules never write taints, so they cannot produce a real
+		// conflict regardless of taint key or selector overlap.
+		if rule.Spec.DryRun || existingRule.Spec.DryRun {
+			continue
+		}
+
 		// Check for same taint key and effect
 		if existingRule.Spec.Taint.Key == rule.Spec.Taint.Key &&
 			existingRule.Spec.Taint.Effect == rule.Spec.Taint.Effect {
--- a/internal/webhook/nodereadinessgaterule_webhook_test.go
+++ b/internal/webhook/nodereadinessgaterule_webhook_test.go
@@ -948,5 +948,103 @@ var _ = Describe("NodeReadinessRule Validation Webhook", func() {
 			Expect(allErrs[0].Field).To(Equal("spec.conditionPolicy"))
 			Expect(allErrs[0].Type).To(Equal(field.ErrorTypeForbidden))
 		})
+
+		It("should allow a dry-run rule with the same taint key as an enforcement rule on overlapping nodes", func() {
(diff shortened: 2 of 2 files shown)


ajaysundark:
Hi, thanks for raising the PR. I am not clear where this is needed. I asked a question on the issue https://github.com/kubernetes-sigs/node-readiness-controller/issues/435#issuecomment-5356173024, can we discuss this further before arriving at a fix?

tejassinghbhati:
Sounds good, 
Actually I just replied on the issue with the use case and safety details, let's continue the discussion over there.

PR #440, opened 2026-08-22
feat: label the CRD with its managing installation and guard duplicat…

### LINKED ISSUE #438: [FEATURE] add a managed-by label to CRD

In a managed provider environment, the control-plane is hosted and node-readiness-controller if installed is invisible to the user. so controller itself is not visible to the end-users risking duplicate installation and two controllers competing to manage taints. Add guardrails / visibility into protecting this scenario.

1. One simplest way to handle this to stamp our artifacts with a 'managed-by' label, ref: https://kubernetes.io/docs/reference/labels-annotations-taints/#app-kubernetes-io-managed-by.

And add clear documentation about it. A provider setup could handle this with a `managed-by:<provider>` label.

2. It should be added to CRD because it will be visible to any NRC install flows and available across namespaces. 

3. Add a helm pre-install hook or similar so the chart installation will detect existing CRD and fail with guidance

1. implement a validation check in the controller itself as a start blocker, but the discovery is complicated and may require additional permissions
2. provide additional exclusive tooling for detecting if a controller exists

yindia:
## Summary
Stamps the `NodeReadinessRule` CRD with a `managed-by` label naming the installation that owns it, and refuses a fresh `helm install` when the CRD already exists.

## Design notes
**Why the CRD carries the label.** It's cluster scoped and singleton, so it's the one artifact every install flow contends over. It's set with a controller-gen marker, so `make manifests` regenerates it in `config/crd/bases`, and `verify-chart-drift.sh` diffs that against the chart's `crds/` copy to catch the two drifting apart.

**Why a template guard, not a pre-install hook.** The Helm guard uses `lookup` + `fail` in a template — same outcome without a ServiceAccount, ClusterRole or image, and it aborts at render time so the guidance lands in the terminal rather than Job logs.

**Why it's keyed on `.Release.IsInstall` plus CRD existence, not the label value.** Both installs may be carrying the default value, and that's precisely the accident being guarded against, so a value comparison would pass straight through it. Upgrades skip the check naturally.

## Two things worth a reviewer's attention

**Label is CRD-only.** Helm 3 forces `managed-by: Helm` on resources it owns, so stamping the Deployment/RBAC/Services doesn't stick. The CRD works because Helm neither templates nor tracks `crds/`.

**Guard only covers `helm install`.** `kubectl apply`, kustomize and the static-pod path have none. A controller startup check would close the gap, but it's ruled out by alternative (1) in the issue — left out and documented. Happy to reconsider.

## Also included
Chart-rendered `NodeReadinessRule` CRs now get the standard chart labels, which they were missing. Unrelated tidy-up, happy to split it out.

## Related Issue
Fixes #438

## Type of Change
/kind feature

## Testing
- `helm install` into a clean cluster:  CRD is created with app.kubernetes.io/managed-by=node-readiness-controller.
- `helm install` with a second release name while the CRD exists: render aborts with the guidance message, nothing is applied.
- `helm upgrade` on the existing release: check is skipped, upgrade succeeds.
- `helm template` output diffed against `dist/crds.yaml`; `verify-chart-drift.sh` passes.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

Yes

```release-note
`helm install` now fails when a `NodeReadinessRule` CRD from another installation is already present in the cluster. The CRD carries a `managed-by` label identifying the owning installation.
```

### DIFF
--- a/charts/node-readiness-controller/README.md
+++ b/charts/node-readiness-controller/README.md
@@ -36,6 +36,36 @@ The command deploys the _node-readiness-controller_ on the Kubernetes cluster in
 
 Helm installs CRDs from the chart `crds/` directory during initial install, but Helm does not upgrade or delete CRDs from that directory during `helm upgrade` or `helm uninstall`. Before upgrading to a chart version that changes the `NodeReadinessRule` schema, apply the updated CRD from the release artifacts or from `charts/node-readiness-controller/crds`.
 
+## Avoiding a duplicate installation
+
+The controller manages node taints. Two installations reconciling the same
+`NodeReadinessRule` resources will both act on the same nodes, so only one should run
+in a cluster.
+
+This is easy to do by accident on a managed cluster: the provider may run the
+controller as part of a hosted control plane, where it is not visible to you.
+
+To make the existing installation visible, the CRD carries the standard
+`app.kubernetes.io/managed-by` label naming the installer. The CRD is cluster scoped
+and shared by every install flow, so its value is whichever installation created it:
+`helm` from this chart, `kustomize` from the config manifests, or a provider's own name
+when they override it.
+
+```console
+$ kubectl get crd nodereadinessrules.readiness.node.x-k8s.io \
+    -o jsonpath='{.metadata.labels.app\.kubernetes\.io/managed-by}'
+helm
+```
+
+A fresh `helm install` refuses to proceed when a controller Deployment is already
+running in the cluster, and names it in the error. (The guard keys on the running
+workload rather than the CRD, because Helm installs the CRD from `crds/` before it
+renders templates, so the CRD always exists by the time the check runs.) `helm upgrade`
+is unaffected. Set `crds.preInstallCheck=false` to install anyway.
+
+Providers installing the controller should override the label value in their own
+pipeline, so the manager shown above identifies them rather than the upstream default.
+
 ## Uninstalling the Chart
 
 To uninstall/delete the `my-release` deployment:
@@ -69,6 +99,7 @@ The following table lists the configurable parameters of the _node-readiness-con
 | `leaderElection.enabled`                 | Enable leader election to support multiple replicas                                                                             | `true`                                                            |
 | `leaderElection.namespace`               | Namespace for the leader election lease. Defaults to the release namespace when empty.                                          | `""`                                                              |
 | `priorityClassName`                      | The name of the priority class to add to pods                                                                                    | `system-cluster-critical`                                         |
+| `crds.preInstallCheck`                   | Fail a fresh install when the `NodeReadinessRule` CRD already exists, so a second controller cannot be installed alongside one you cannot see. No effect on `helm upgrade`. | `true`                                                            |
 | `rbac.create`                            | If `true`, create & use RBAC resources                                                                                          | `true`                                                            |
 | `resources`                              | Node Readiness Controller container CPU and memory requests/limits                                                              | _see values.yaml_                                                 |
 | `serviceAccount.create`                  | If `true`, create a service account                                                                                             | `true`                                                            |
--- a/charts/node-readiness-controller/crds/nodereadinessrules.readiness.node.x-k8s.io.yaml
+++ b/charts/node-readiness-controller/crds/nodereadinessrules.readiness.node.x-k8s.io.yaml
@@ -4,6 +4,8 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.19.0
+  labels:
+    app.kubernetes.io/managed-by: helm
   name: nodereadinessrules.readiness.node.x-k8s.io
 spec:
   group: readiness.node.x-k8s.io
--- a/.github/workflows/helm.yaml
+++ b/.github/workflows/helm.yaml
@@ -79,3 +79,46 @@ jobs:
       
       - name: Run chart-testing (install)
         run: ct install --config=.github/ci/ct.yaml
+
+      # The pre-install guard calls Helm's lookup, which returns nothing during
+      # `helm template` and `--dry-run`, so the chart unit tests cannot cover it.
+      # Exercise it against the live kind cluster instead.
+      #
+      # The signal is the controller Deployment, not the CRD: Helm applies crds/
+      # before rendering, so keying on the CRD would refuse every first install.
+      - name: Verify the duplicate-install pre-install check
+        run: |
+          set -euo pipefail
+          CHART=charts/node-readiness-controller
+
+          echo "::group::a fresh install into an empty cluster succeeds with the guard ON"
+          # This is the path the CRD-based guard broke: crds/ is applied before the
+          # templates render, so the check must not trip on this install's own CRD.
+          helm install nrc-guard "$CHART" \
+            --namespace nrc-guard --create-namespace --wait --timeout 5m
+          echo "::endgroup::"
+
+          echo "::group::a second install is refused while the first is running"
+          if helm install nrc-guard-2 "$CHART" \
+               --namespace nrc-guard-2 --create-namespace 2>guard.err; then
+            echo "expected the second install to be refused, but it succeeded"
+            exit 1
+          fi
+          cat guard.err
+          grep -q "already installed in this cluster" guard.err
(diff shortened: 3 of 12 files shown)


yindia:
@ajaysundarkI'm not sure it's the best approach, it false-positives on reinstall since Helm keeps crds/ CRDs after uninstall. Open to a better ideas if anyone has one

DsThakurRawat:
On the uninstall/reinstall false positive:

The reason `lookup` on the CRD trips is that Helm 3 leaves `crds/` in place on `helm uninstall` by design. If `crd-ownership-check.yaml` fails solely on CRD presence, any standard `helm uninstall` -> `helm install` cycle in the same cluster gets blocked.

Two ways to resolve this without dropping the guard:

1. Guard on active workload presence rather than CRD presence: query for the Deployment in the target namespace (`lookup "apps/v1" "Deployment" .Release.Namespace (include "node-readiness-controller.fullname" .)`). Helm deletes the Deployment on uninstall, so a fresh install after an uninstall proceeds cleanly, while a true concurrent installation in that namespace is caught.
2. If the goal is catching hosted provider installs where the Deployment lives in a hidden namespace, check the label value rather than existence: only abort if `$existing.metadata.labels["app.kubernetes.io/managed-by"]` exists AND is not equal to `"Helm"` / our default manager. That way, lingering unmanaged CRDs from a previous uninstall do not block a fresh install.

yindia:
@DsThakurRawat Thanks for the detailed look. This finds the false positive well.

Option 2 does not work here. Both the provider install and my install carry the default `managed-by` value. A test for `!= "Helm"` passes through the exact case that the guard must catch. This is the reason I keyed the check on CRD existence, not on the label value.

Option 1 fixes the uninstall and reinstall cycle. But it misses the main case. A provider can run the controller in a control-plane namespace that I cannot see. Then there is no Deployment in `.Release.Namespace` to find. That hidden install is the reason the guard exists.

I think one version gets both. Keep the live workload as the signal, as you suggested. But look it up across all namespaces, not only the release namespace:

```
lookup "apps/v1" "Deployment" "" ""
```

Then match on `app.kubernetes.io/name=node-readiness-controller`. A leftover CRD after an uninstall has no running Deployment. So a fresh install works. A controller in any namespace, including a hidden one, still trips the guard. The cost is one cluster-wide Deployment read at install time. The installing user normally has this access.

Does this match the reinstall case that you saw? If yes, I will push it as a follow-up commit.

cc: @ajaysundark

DsThakurRawat:
yes, that matches the case i saw, and i checked it on a live cluster rather than reasoning about it. kind v1.31.2, helm 3.16.3, your branch at 55e6a23.

the uninstall half holds up: `helm uninstall` leaves the CRD behind (it ships from `crds/`, which helm 3 never removes) and does delete the Deployment, so a live workload really is the signal that separates the two states. your cluster-wide version catches the hidden install too. i faked one by putting a Deployment labelled `app.kubernetes.io/name=node-readiness-controller` into a namespace the installer would never look in, and `lookup "apps/v1" "Deployment" "" ""` found it; after deleting that Deployment the same lookup found nothing. so the answer to your question is yes.

while setting that up i hit something bigger than the reinstall case though, and i think it blocks the PR as it stands. **the guard fires on a first install into an empty cluster.** helm applies `crds/` before it renders templates, so by the time `crd-ownership-check.yaml` runs its lookup, the CRD that this same `helm install` created a moment earlier is already there.

from a verified clean state, twice:

```
$ kubectl get crd | grep -c readiness
0
$ helm list -A --short | wc -l
0
$ helm install nrc charts/node-readiness-controller -n nrc-system --create-namespace
Error: INSTALLATION FAILED: execution error at (node-readiness-controller/templates/crd-ownership-check.yaml:31:4):
node-readiness-controller is already installed in this cluster.
$ helm list -A --short | wc -l
0
$ kubectl get crd nodereadinessrules.readiness.node.x-k8s.io -o jsonpath='{.metadata.creationTimestamp}'
2026-08-25T21:06:59Z
```

that timestamp is from the install that then refused itself, and no release was recorded. `--set crds.preInstallCheck=false` installs fine, and i confirmed `helm upgrade` still works with the guard left on, so the `.Release.IsInstall` gate is doing its job.

two reasons this got past CI, and i think both are worth fixing next to it:

- the new "Verify the CRD ownership pre-install check" step runs `kubectl apply -f "$CHART/crds/"` before it tries the install, so it only ever exercises the path where the CRD legitimately pre-exists. nothing covers the first install.
- the Helm workflow on this head is sitting at `action_required` and has not actually run yet, so `ct install` never got the chance to fail either. i suspect it would.

if you do go with the cluster-wide Deployment lookup, two things bit me while writing a probe chart for it.

`lookup` with an empty name returns a list object, and that object is truthy even when nothing matched. on my cluster the returned dict had keys `[apiVersion items kind metadata]` with `items` empty, so a bare `if $deployments` is always true. you have to range over `.items` and count what actually matched.

the other one is that a Deployment carrying no labels at all is ordinary rather than an edge case. one of the two in a stock kind cluster is `local-path-storage/local-path-provisioner` and it has none, so `index .metadata.labels "app.kubernetes.io/name"` on it dies with `index of untyped nil`. i had to switch to `get (.metadata.labels | default dict) "app.kubernetes.io/name"` to get through a full pass.

could be wrong about the ordering if you get a different result, but two clean runs said the same thing. can send the probe chart over if it saves you the setup.

yindia:
You're right on all counts, and thanks for testing it on a live cluster. I reproduced the same thing: a fresh `helm install` into an empty cluster refuses itself, because Helm applies `crds/` before it renders templates, so the guard's `lookup` finds the CRD this install just created. That's a real blocker, not the reinstall edge case.

Fixed by keying the guard on the controller **Deployment** instead of the CRD. The Deployment is applied after rendering, so one found during the check belongs to a different, already-running install. Both `lookup` foot-guns you hit are handled: I `range .items` and count matches rather than testing the list for truthiness, and I default the label map before indexing (`get (.metadata.labels | default dict) ...`) so a label-less Deployment like `local-path-provisioner` doesn't panic.

CI was passing because the verify step ran `kubectl apply -f crds/` first, so it only ever exercised the pre-existing path. I changed it to install into an empty cluster with the guard on first (the path that was broken), then assert the second install is refused, the escape hatch works, and upgrade works.

Verified on kind:
- first install, empty cluster, guard on: succeeds
- second install while one runs: refused, names the Deployment
- `crds.preInstallCheck=false`: installs
- `helm upgrade` with guard on: works

No need for the probe chart, but thanks for the offer. Pushed the fix.

charts/node-readiness-controller/crds/nodereadinessrules.readiness.node.x-k8s.io.yaml
```diff
@@ -4,6 +4,8 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.19.0
+  labels:
+    app.kubernetes.io/managed-by: node-readiness-controller
```
ajaysundark: AFAIU, this is set as 'managed-by: kustomize', we can default to managed-by: helm' for charts. If there's way to expose this to be user-set value in helm, that'd be preferred.

I expect in managed-environments this would map to `managed-by: <provider>`, we should document as best-practice. so if there's conflict, it'd be properly bubble up as an error

yindia:
@ajaysundark do you have any other comment ?

charts/node-readiness-controller/README.md
```diff
@@ -36,6 +36,36 @@ The command deploys the _node-readiness-controller_ on the Kubernetes cluster in
...
+$ kubectl get crd nodereadinessrules.readiness.node.x-k8s.io \
+    -o jsonpath='{.metadata.labels.app\.kubernetes\.io/managed-by}'
+helm
+```
+
+A fresh `helm install` refuses to proceed when a controller Deployment is already
+running in the cluster, and names it in the error. (The guard keys on the running
+workload rather than the CRD, because Helm installs the CRD from `crds/` before it
+renders templates, so the CRD always exists by the time the check runs.) `helm upgrade`
+is unaffected. Set `crds.preInstallCheck=false` to install anyway.
```
ajaysundark: Could you clarify this, why it has to be disabled on upgrade?

PR #447, opened 2026-08-24
Update node_readiness_evaluation_duration_seconds  + tests

rawadhossain:
## Description
- Updates HELP string on `node_readiness_evaluation_duration_seconds` to include taint operations. No name, label, or bucket change.
- Adds related tests to `internal/metrics/metrics_test.go`, also contributes to #269.

### from the design doc:
image
image

## Type of Change
/kind documentation

## Testing
`make test`, `make lint`, `go vet`, `gofmt -l` all pass.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -87,7 +87,7 @@ var (
 	EvaluationDuration = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name:    "node_readiness_evaluation_duration_seconds",
-			Help:    "Duration of rule evaluations per rule, including taint operations",
+			Help:    "Duration of evaluating a rule against a node, including any taint add/remove operations",
 			Buckets: prometheus.DefBuckets,
 		},
 		[]string{"rule"},
--- a/internal/metrics/metrics.go
+++ b/internal/metrics/metrics.go
@@ -87,7 +87,7 @@ var (
 	EvaluationDuration = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name:    "node_readiness_evaluation_duration_seconds",
-			Help:    "Duration of rule evaluations per rule",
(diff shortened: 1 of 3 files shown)


internal/metrics/metrics.go
```diff
@@ -87,7 +87,7 @@ var (
 	EvaluationDuration = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
 			Name:    "node_readiness_evaluation_duration_seconds",
-			Help:    "Duration of rule evaluations per rule",
+			Help:    "Duration of rule evaluations per rule, including taint operations",
```
AvineshTripathi:
```suggestion
			Help:    "End-to-end duration of rule evaluation, including any taint add/remove operations",
```

The current wording "Duration of rule evaluations per rule, including taint operations" reads as if the metric tracks two separate things. The suggested wording makes it clear that taint operations are part of what's being timed.

wdyt? Open to even better wording

AvineshTripathi (commented):
I feel we are adding redundancy here for every metric test. All the metrics-related test cases, unless they have a special case, will follow the same structure i.e.
1. Check if they are getting registered and have correct help and format etc
2. a case where we test the reflection(a change is done and we want that change to reflect it)

I feel we can give a generic test or maybe reuse for above cases

internal/metrics/metrics.go
rawadhossain: @AvineshTripathi I think that's a good point. Made the changes, now two reusable helpers used in tests and also updated the help text to `Duration of evaluating a rule against a node, including any taint add/remove operations`.

PTAL. Thanks!

AvineshTripathi:
/lgtm

cc @ajaysundark for approval

ajaysundark:
/approve

Thanks!

PR #448, opened 2026-08-24
feat: add node_readiness_rules with enforcement_mode and dry_run labels

rawadhossain:
## Description
- Adds `node_readiness_rules{enforcement_mode, dry_run}` while keeping `node_readiness_rules_total` unchanged for compatibility.
- Serves `node_readiness_rules` from the scrape-time ReadinessCollector. the original ruleCache push approach showed staleness scaling with rule count under bulk changes
- Updates `monitoring.md` and `docs/TEST_README.md`.
- Adds related tests, also contributes to #269.

### from the design doc:
image
image

## Type of Change
/kind feature

## Testing
`make test`, `make lint`, `go vet`, `gofmt -l` all pass.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/docs/book/src/operations/monitoring.md
+++ b/docs/book/src/operations/monitoring.md
@@ -10,6 +10,8 @@ The controller serves metrics on `/metrics` only when metrics are explicitly ena
 
 ### `node_readiness_rules_total`
 
+***Deprecated:** use [`node_readiness_rules`](#node_readiness_rules) instead. It provides the same rule count with additional `enforcement_mode` and `dry_run` labels. `node_readiness_rules_total` is still published for compatibility.*
+
 Number of `NodeReadinessRule` objects tracked by the controller.
 
 | Property | Value |
@@ -17,6 +19,23 @@ Number of `NodeReadinessRule` objects tracked by the controller.
 | Type | `gauge` |
 | Labels | none |
 | Recorded when | The controller refreshes or removes a tracked rule |
+
+### `node_readiness_rules`
+
+Number of `NodeReadinessRule` objects tracked by the controller by enforcement mode and dry-run state.
+
+| Property | Value |
+| --- | --- |
+| Type | `gauge` |
+| Labels | `enforcement_mode`, `dry_run` |
+| Recorded when | The controller refreshes or removes a tracked rule |
+
+#### Labels
+
+| Label | Description | Values |
+| --- | --- | --- |
+| `enforcement_mode` | Enforcement mode of the rule | `bootstrap-only`, `continuous` |
+| `dry_run` | Whether the rule is in dry-run mode | `true`, `false` |
 
 ### `node_readiness_taint_operations_total`
 
--- a/docs/TEST_README.md
+++ b/docs/TEST_README.md
@@ -241,6 +241,9 @@ After running the test scenario, you should see the following metrics:
    ```bash
    # Number of active rules
    curl -s http://localhost:8080/metrics | grep "node_readiness_rules_total"
+
+   # Number of active rules by enforcement mode and dry-run state
+   curl -s http://localhost:8080/metrics | grep "node_readiness_rules{"
    ```
 
 2. **Taint Operations:**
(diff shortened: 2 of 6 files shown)


AvineshTripathi (commented):
Since we're already calling [`ListRules`](https://github.com/kubernetes-sigs/node-readiness-controller/blob/main/internal/controller/nodereadinessrule_controller.go#L544) in the collector, we can compute the enforcement_mode/dry_run breakdown there instead of maintaining a separate in-memory sync. The local ruleCache can lag behind actual cluster state and can be unreliable. 

cc @ajaysundark wdyt?

rawadhossain:
@AvineshTripathi Good point. I tested both ways and found `ruleCache` can be stale on startup (~200-300ms) and for ~1s after creating a rule, and ListRules reflects the current state right away. The stale window is relatively small, so I think keeping the current approach is reasonable for now.

@ajaysundark since this metric wasn't part of the collector scope, should we switch or keep it separate for now?

AvineshTripathi:
> The stale window is relatively small, so I think keeping the current approach is reasonable for now.

I would prefer testing this delay on the scale before doing the scale. Can we run the scan?

rawadhossain:
Yes we can and thanks it helped. 

Ran the scale test. Used the `test/scale` setup. So here’s the findings:

**ruleCache staleness (current one):**

| Scenario | N=100 | N=1000 | N=5000 |
|---|---:|---:|---:|
| One rule changes | ~130ms | ~135ms | ~145ms |
| One rule gets created | ~1.15s | ~1.15s | ~1.2s |
| All N rules created at once | ~2s | ~14s | ~79s |
| All N rules dryRun flipped at once | ~1.5s | ~11s | ~62s |
| All N rules deleted at once | ~1s | ~11s | ~57s |

Single changes are fine, but bigger difference shows up when a large number of rules change together, delay grows quite long.

**Scrape-time approach:**

| N | 100 | 1000 | 5000 |
|---|---:|---:|---:|
| Standalone lookup | ~0.37ms | ~4.1ms | ~23ms |
| Added to existing collector | ~3µs | ~26µs | ~128µs |

Since the collector already does that fetch, adding this one on barely adds anything.

Based on these, would it make more sense to go with collector instead, specially for scale? Staleness gets pretty noticeable with bulk changes, also extra cost in the collector is very small.

@AvineshTripathi @ajaysundark wdyt? I can make the changes if we decide to go with collector approach.

AvineshTripathi:
@rawadhossain I'll go with the collector approach! that can be a standard we can go for

rawadhossain:
@AvineshTripathi I also think so collector approach is the one to go for. Made the changes. Tested and everything's working as expected. 
PTAL. Thanks!

cc @ajaysundark

AvineshTripathi:
/lgtm 

havn't tested in local but looks good!

docs/book/src/operations/monitoring.md
```diff
@@ -10,6 +10,8 @@ The controller serves metrics on `/metrics` only when metrics are explicitly ena
 
 ### `node_readiness_rules_total`
 
+***Deprecated:** use [`node_readiness_rules`](#node_readiness_rules) instead. It provides the same rule count with additional `enforcement_mode` and `dry_run` labels. `node_readiness_rules_total` is still published for compatibility.*
```
ajaysundark, line 13: Can we also add a [DEPRECATED] prefix in it's helper text?

docs/book/src/operations/monitoring.md
```diff
@@ -18,6 +20,23 @@ Number of `NodeReadinessRule` objects tracked by the controller.
...
 
+### `node_readiness_rules`
+
+Number of `NodeReadinessRule` objects tracked by the controller by enforcement mode and dry-run state.
+
+| Property | Value |
+| --- | --- |
+| Type | `gauge` |
+| Labels | `enforcement_mode`, `dry_run` |
+| Recorded when | Computed on each Prometheus scrape from the cached rule list |
```
ajaysundark: Should we mention here rules being deleted are excluded in count?

ajaysundark (commented):
/lgtm

docs/book/src/operations/monitoring.md
rawadhossain: yes, we can, would be even better. 
Added it.

docs/book/src/operations/monitoring.md
rawadhossain: Yeah, can be a good detail to add. 
Updated the doc.

PR #451, opened 2026-08-24
feat: migrate bootstrap-only validations to CEL

### LINKED ISSUE #449: Migrate remaining bootstrap-only validations from admission webhook to CEL

As discussed with @ajaysundark in PR #418:

Currently, some configuration constraints for the bootstrap-only enforcement mode are implemented in the Go admission webhook rather than natively via CRD CEL validation (+kubebuilder:validation:XValidation).

The constraints currently living in the webhook are:
1. defaultStatus should not be used with bootstrap-only mode.
2. anyOf conditionPolicy is not supported with bootstrap-only mode.

To move closer to our goal of eventually deprecating and removing the validating webhook entirely in favor of native Kubernetes CEL, we should migrate the remaining bootstrap-only validations to CEL expressions on the NodeReadinessRuleSpec.

We can enforce these natively by adding the following XValidation markers:
```
// +kubebuilder:validation:XValidation:rule="!has(self.enforcementMode) || self.enforcementMode != 'bootstrap-only' || !has(self.conditionPolicy) || self.conditionPolicy != 'anyOf'",message="anyOf conditionPolicy is not supported with bootstrap-only enforcementMode"
// +kubebuilder:validation:XValidation:rule="!has(self.enforcementMode) || self.enforcementMode != 'bootstrap-only' || !has(self.conditions) || self.conditions.all(c, !has(c.defaultStatus))",message="defaultStatus should not be used with bootstrap-only enforcementMode"
```

Once these are added and tested, the corresponding Go validation logic and unit tests can be cleanly removed from `internal/webhook/nodereadinessgaterule_webhook.go`.

N/A

vishnukothakapu:
## Description
This PR completes the migration of the remaining `bootstrap-only` enforcement validations from the admission webhook to native CEL `XValidation` markers on the `NodeReadinessRuleSpec`. 

Specifically, it blocks two invalid configurations directly at the API level:
1. `enforcementMode: bootstrap-only` combined with `defaultStatus`
2. `enforcementMode: bootstrap-only` combined with `conditionPolicy: anyOf`

This aligns with the clean hybrid approach established in PR #418, ensuring standard user templates using `allOf` continue to work while safely short-circuiting unsupported bootstrap-only configurations without the need for a webhook round-trip.

Changes include:
- Addition of two CEL `XValidation` markers to `nodereadinessrule_types.go`
- Regeneration of the CRD schema
- Removal of the legacy Go validation logic and associated unit tests from `nodereadinessgaterule_webhook.go`
- Addition of mirroring CEL controller `envtest` test cases
- Minor documentation update reflecting that the API server (not the webhook) now handles these rejections.

## Related Issue
Fixes #449

## Type of Change
/kind feature
/kind api-change
/kind cleanup

## Testing
- Automated `envtest` cases were added to `nodereadinessrule_controller_test.go` to explicitly verify that the API server rejects these invalid manifests during creation.
- Webhook unit tests were cleaned up and verified to pass.
- Manual verification of the generated CRD schema confirmed the CEL markers propagated correctly.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Migrated the remaining `bootstrap-only` validation logic from the admission webhook to native CRD CEL validations, improving API server efficiency when creating or updating NodeReadinessRules.

### CODE REVIEW

**Comment by tejassinghbhati** [comment]
Heads up on a collision with #455, which migrates the nodeSelector check to CEL. Different constraint to yours, but we overlap on four files.

Two concrete spots. You delete the `accumulate errors across nodeSelector, defaultStatus, and conditionPolicy violations` spec, I edit it, so that will conflict.

The one worth knowing about is in `should perform comprehensive validation`. Your diff leaves this as context.

```go
Expect(allErrs).To(HaveLen(1)) // Empty nodeSelector validation
```

That holds today but stops holding once #455 removes the empty check from the webhook, so whichever of ours merges second needs that repointed at an unparseable selector rather than an empty one.

Happy to rebase mine on yours if you land first, no preference.

yindia:
Same thing I flagged on #455: the chart CRD didn't get regenerated. Both new bootstrap-only rules are in `config/crd/bases/` but not in `charts/node-readiness-controller/crds/`. So `verify-chart-drift.sh` will fail and Helm installs won't actually enforce them. Could you sync the chart CRD?

internal/webhook/nodereadinessgaterule_webhook.go
```diff
@@ -77,27 +77,6 @@ func (w *NodeReadinessRuleWebhook) validateSpec(
...
-	if spec.ConditionPolicy == readinessv1alpha1.ConditionPolicyAnyOf &&
-		spec.EnforcementMode == readinessv1alpha1.EnforcementModeBootstrapOnly {
-		allErrs = append(allErrs, field.Forbidden(
-			field.NewPath("spec", "conditionPolicy"),
-			"anyOf conditionPolicy is not supported with bootstrap-only enforcementMode",
-		))
-	}
-
 	return allErrs
 }
```
yindia, line 74: Now that both bootstrap-only checks are gone, this if `isUpdate` block doesn't do anything, both paths just return `allErrs`, and the comment mentions checks that no longer exist. Might as well drop it while we're cleaning up

yindia (approved):

internal/webhook/nodereadinessgaterule_webhook.go
yindia, line 61: validateSpec doesn't use isUpdate anymore, drop the param and its call-site arg.

docs/book/src/user-guide/concepts.md
```diff
@@ -86,7 +86,7 @@ The choice of `defaultStatus` has critical interaction with the rule's `enforcem
 >
 > `defaultStatus` is useful when a condition may never appear on a node in its healthy state, effectively treating its absence as a known-good signal. However, `bootstrap-only` mode exists precisely to *wait* for conditions to be explicitly reported before completing the bootstrap gate.
 >
- > Because these two features serve opposing purposes, using them together can lead to unintended behavior, such as completing the bootstrap phase before a condition is actually verified. To prevent  this, the admission webhook explicitly rejects this combination.
+ > Because these two features serve opposing purposes, using them together can lead to unintended behavior, such as completing the bootstrap phase before a condition is actually verified. To prevent this, the API server explicitly rejects this combination.
```
ajaysundark:
```suggestion
 > Because these two features serve opposing purposes, using them together can lead to unintended behavior, such as completing the bootstrap phase before a condition is actually verified. To prevent this, the controller explicitly rejects this combination.
```

nit: how it is validated is an implementation detail.

ajaysundark (commented):
Thanks!

/lgtm

ajaysundark (approved):
/lgtm
/approve

PR #455, opened 2026-08-25
feat: enforce non-empty nodeSelector with CEL instead of the webhook

### LINKED ISSUE #454: Migrate the non-empty nodeSelector validation from the admission webhook to CEL

Following the direction in #449, moving validation out of the Go admission webhook and onto the CRD as native CEL.

One constraint sits outside that issue's scope. `validateSpec` rejects an empty `spec.nodeSelector`:

```go
if selector != nil && selector.Empty() {
}
```

`metav1.LabelSelectorAsSelector` turns an empty selector into `labels.Everything()`, so a rule carrying one applies its taint to every Node in the cluster. With `NoExecute` that evicts pods everywhere that lack a toleration.

The webhook is optional and off by default (`--enable-webhook=false`, `webhook.enabled: false` in the chart), so on a default install nothing stops it. #403 hit this from the other side: the chart's own error message suggested `nodeSelector: {}` for matching all Nodes, which the webhook forbids and the CRD allows.

Express it as CEL on the field, so it applies on every cluster whether or not the webhook is deployed:

```go
// +kubebuilder:validation:XValidation:rule="(has(self.matchLabels) && size(self.matchLabels) > 0) || (has(self.matchExpressions) && size(self.matchExpressions) > 0)",message="nodeSelector must not be empty"
NodeSelector metav1.LabelSelector `json:"nodeSelector,omitempty,omitzero"`
```

Then drop the empty check from `validateSpec`. CRD validation runs before validating webhooks, so the branch becomes unreachable once the rule is in place.

Two details worth stating, since both surfaced while testing this against envtest:

The `LabelSelectorAsSelector` error check has to stay in the webhook. CEL can see that `matchExpressions` is non-empty but not whether an operator is one the selector parser accepts, so an entry like `{key: k, operator: NotARealOperator}` still needs Go.

An absent `nodeSelector` is caught by the `+required` marker, not by this rule. The field is `omitempty,omitzero`, so an empty struct serialises away entirely and the CEL rule never evaluates. The API server reports `spec.nodeSelector: Required value` for that case and `nodeSelector must not be empty` for a selector that is present but empty. Both are rejected, they just report differently, and it is worth covering both paths in tests.

After this, the only validation left in the webhook is cross-object taint conflict detection, which genuinely cannot move to CEL since it has to list other rules.

Leaving it in the webhook and documenting that the guard requires the webhook. That is what happens today and it is what made #403 possible, so it does not seem worth keeping.

Adding it to the CRD while leaving the webhook copy in place. Harmless but dead, since CRD validation runs first.

### Additional context

Checked for overlap before filing. #451 covers the two bootstrap-only constraints from #449 and does not touch `nodeSelector`. No other open issue or PR references the empty selector check.

PR to follow.

tejassinghbhati:
## Description

Continues the webhook to CEL migration from #449. This one is outside #451's scope, which covers the two bootstrap-only constraints.

An empty `spec.nodeSelector` becomes `labels.Everything()`, so a rule carrying one taints every Node in the cluster. The only guard was `validateSpec` in the validating webhook, which is off by default, so a default install had nothing stopping it. #403 came at the same gap from the chart side, where the chart's own error text suggested `nodeSelector: {}` while the webhook forbade it.

The constraint now lives on the CRD:

```go
// +kubebuilder:validation:XValidation:rule="(has(self.matchLabels) && size(self.matchLabels) > 0) || (has(self.matchExpressions) && size(self.matchExpressions) > 0)",message="nodeSelector must not be empty"
```

and the empty check is removed from the webhook rather than left beside it, since CRD validation runs before validating webhooks and the branch is unreachable once the rule is in place.

Two things I deliberately did not change:

The `LabelSelectorAsSelector` error check stays. CEL can tell that `matchExpressions` is non-empty but not whether the operator is one the parser accepts, so `{key: k, operator: NotARealOperator}` still needs Go. The webhook specs that used an empty selector as their invalid input now use an unparseable one instead, so they exercise what the webhook still owns.

An absent `nodeSelector` is caught by `+required`, not by this rule, because the field is `omitempty,omitzero` and serialises away so CEL never evaluates. I expected my rule to catch it and it does not, which is why there is a separate spec asserting `spec.nodeSelector: Required value` for that path.

After this the only thing left in the webhook is cross-object taint conflict detection, which cannot move to CEL because it has to list other rules.

## Related Issue

Fixes #454

## Type of Change

/kind feature
/kind api-change

## Testing

New `nodeSelector CEL validation` specs drive a real API server through envtest rather than calling the validator directly, since the point is that the API server enforces it now:

- absent selector, rejected with `spec.nodeSelector: Required value`
- selector present but empty, rejected with `nodeSelector must not be empty`
- `matchLabels: {}` present but empty, rejected
- `matchExpressions: []` present but empty, rejected
- `matchLabels` set, accepted and persisted
- `matchExpressions` only, accepted

Full suite: 83 of 83 specs, up from 78. `internal/webhook` 37 specs green after the three migrated ones were reworked. `internal/metrics` and the reporter pass. `golangci-lint` v2.12.1 reports 0 issues and `gofmt` is clean.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
NodeReadinessRule now rejects an empty spec.nodeSelector through CRD validation rather than only through the optional admission webhook, so a rule can no longer be created that silently taints every node on clusters where the webhook is not deployed.
```

### DIFF
--- a/internal/controller/nodeselector_cel_test.go
+++ b/internal/controller/nodeselector_cel_test.go
@@ -1,0 +1,109 @@
+/*
+Copyright The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"context"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+
+	nodereadinessiov1alpha1 "sigs.k8s.io/node-readiness-controller/api/v1alpha1"
+)
+
+// The empty nodeSelector constraint is enforced by CEL on the CRD rather than by
+// the validating webhook, so it applies on every cluster instead of only where the
+// optional webhook is deployed. These specs assert it through the API server.
+var _ = Describe("NodeReadinessRule nodeSelector CEL validation", func() {
+	var celCtx context.Context
+
+	newRule := func(name string, selector metav1.LabelSelector) *nodereadinessiov1alpha1.NodeReadinessRule {
+		return &nodereadinessiov1alpha1.NodeReadinessRule{
+			ObjectMeta: metav1.ObjectMeta{Name: name},
+			Spec: nodereadinessiov1alpha1.NodeReadinessRuleSpec{
+				Conditions: []nodereadinessiov1alpha1.ConditionRequirement{
+					{Type: "CELReady", RequiredStatus: corev1.ConditionTrue},
+				},
+				NodeSelector:    selector,
+				Taint:           corev1.Taint{Key: "readiness.k8s.io/cel-selector", Effect: corev1.TaintEffectNoSchedule},
+				EnforcementMode: nodereadinessiov1alpha1.EnforcementModeContinuous,
+			},
+		}
+	}
+
+	BeforeEach(func() { celCtx = context.Background() })
+
+	AfterEach(func() {
+		list := &nodereadinessiov1alpha1.NodeReadinessRuleList{}
+		if err := k8sClient.List(celCtx, list); err == nil {
+			for i := range list.Items {
+				r := &list.Items[i]
+				if len(r.Name) >= 12 && r.Name[:12] == "cel-selector" {
+					r.Finalizers = nil
+					_ = k8sClient.Update(celCtx, r)
+					_ = k8sClient.Delete(celCtx, r)
+				}
+			}
+		}
+	})
+
+	// A wholly absent selector is caught by the required marker rather than by the
+	// CEL rule, because the field is omitempty/omitzero and serialises away
+	// entirely. Both reject the object, they just report it differently.
+	It("rejects a rule with no selector at all", func() {
+		err := k8sClient.Create(celCtx, newRule("cel-selector-absent", metav1.LabelSelector{}))
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("spec.nodeSelector: Required value"))
+	})
+
+	It("rejects a selector whose matchLabels map is present but empty", func() {
+		err := k8sClient.Create(celCtx, newRule("cel-selector-emptylabels",
+			metav1.LabelSelector{MatchLabels: map[string]string{}}))
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("nodeSelector must not be empty"))
+	})
+
+	It("rejects a selector whose matchExpressions list is present but empty", func() {
+		err := k8sClient.Create(celCtx, newRule("cel-selector-emptyexprs",
+			metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{}}))
+		Expect(err).To(HaveOccurred())
+		Expect(err.Error()).To(ContainSubstring("nodeSelector must not be empty"))
+	})
+
+	It("accepts a selector with matchLabels", func() {
+		rule := newRule("cel-selector-labels", metav1.LabelSelector{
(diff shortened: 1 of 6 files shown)


tejassinghbhati:
Flagging an overlap with #451 before either of these merges, so nobody lands a broken combination.

Four of the five files here are also touched by #451. Two spots actually conflict rather than just sitting near each other.

#451 deletes the `should accumulate errors across nodeSelector, defaultStatus, and conditionPolicy violations` spec. This PR edits that same spec instead, since nodeSelector stops contributing an error once the CEL rule lands. Whichever merges second will hit a conflict there.

The one that matters more: in `should perform comprehensive validation`, #451 leaves this untouched as context.

```go
allErrs = webhook.validateNodeReadinessRule(ctx, invalidRule, false)
Expect(allErrs).To(HaveLen(1)) // Empty nodeSelector validation
```

That assertion only holds while the webhook still rejects an empty selector, which this PR removes. So if #451 goes in first and this is rebased naively, that spec either fails or passes for the wrong reason. This PR repoints it at an unparseable selector, which is what the webhook still owns.

No preference on ordering, happy to rebase on #451 if you would rather take that one first. Just did not want it discovered at merge time.

/cc @vishnukothakapu

internal/controller/nodeselector_cel_test.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	BeforeEach(func() { celCtx = context.Background() })
+
+	AfterEach(func() {
+		list := &nodereadinessiov1alpha1.NodeReadinessRuleList{}
+		if err := k8sClient.List(celCtx, list); err == nil {
+			for i := range list.Items {
+				r := &list.Items[i]
+				if len(r.Name) >= 12 && r.Name[:12] == "cel-selector" {
```
yindia: nit 
```suggestion
                    if strings.HasPrefix(r.Name, "cel-selector-") {
   ```

**Comment by yindia** [comment]
@tejassinghbhati One thing I noticed: the chart CRD didn't get updated. The new `nodeSelector must not be empty` rule is in `config/crd/bases/` but not in `charts/node-readiness-controller/crds/`. Because of that, the drift check will fail and Helm installs won't actually get the new rule — which is the whole point of this PR. Could you copy the change into the chart CRD too?

File: internal/controller/nodeselector_cel_test.go (hunk 1 above)
Comment (tejassinghbhati) [reply internal/controller/nodeselector_cel_test.go]: Taken, thanks.

**Comment by tejassinghbhati** [comment]
Good catch @yindia, that was a real miss and you are right about the consequence. Fixed in 9d1cc14.

I regenerated into config/crd/bases and never copied it across, so verify-chart-drift.sh would have failed and a Helm install would not have carried the rule at all, which is the entire point of the PR. Slightly galling given #372 is me adding a drift check for the RBAC half of exactly this.

Verified after the copy: the two CRDs diff clean, and rendering the chart with --include-crds now shows the rule present. Full suite still 83 of 83, chart tests green.

**Comment by tejassinghbhati** [comment]
Rebased onto main now that #451 has landed.

Three conflicts in the webhook tests, all in the two specs that asserted the webhook rejecting an empty selector, which this PR moves to CEL. Worth noting the third one: my branch still carried the defaultStatus and conditionPolicy webhook tests that #451 deleted, so taking my side there would have resurrected tests for validation that no longer lives in the webhook. Dropped those and kept yours.

Checked that both sets of CEL rules survived rather than one clobbering the other. The CRD now carries the anyOf and defaultStatus rules from #451 alongside nodeSelector must not be empty, and the chart CRD copy matches.

validateSpec is down to just the LabelSelectorAsSelector parse check now, since CEL covers everything else and cross-object taint conflicts live in validateTaintConflicts.

85 of 85 specs, webhook suite green, lint clean.

File: internal/webhook/nodereadinessgaterule_webhook.go [hunk 2]
```diff
@@ -61,14 +61,13 @@ func (w *NodeReadinessRuleWebhook) validateSpec(
...
 	var allErrs field.ErrorList
 
-	// validate that the nodeSelector isn't empty
-	selector, err := metav1.LabelSelectorAsSelector(&spec.NodeSelector)
-	if err != nil {
+	// An empty nodeSelector is rejected by CEL on the CRD, which runs before
+	// validating webhooks, so that case cannot reach here. What CEL cannot express
+	// is whether the selector actually parses, for example an unknown
+	// matchExpressions operator, so that check stays.
+	if _, err := metav1.LabelSelectorAsSelector(&spec.NodeSelector); err != nil {
```
Karthik-K-N, line 68: I agree its better to keep the code comments descriptive but if we can maintain them crisp and short it would be great, so can we just mention the current state of logic so we have this check to validate that the node selector is valid?

Karthik-K-N (commented):
Only one nit otherwise LGTM. 
Thank you for working on this.

PR #456, opened 2026-08-25
feat(chart): add prometheus servicemonitor support

### LINKED ISSUE #443: [FEATURE] Helm chart missing ServiceMonitor template for Prometheus Operator integration

Currently, the Helm chart exposes metrics via a Service but it lacks a template for a ServiceMonitor CRD.

For users running the Prometheus, they are forced to manually write & maintain their own ServiceMonitor manifests outside of the Helm release. 

While users can technically use standard prometheus.io/scrape pod annotations, those annotations are inadequate for scraping secure HTTPS endpoints since they cannot natively pass tlsConfig, caFile, or bearerToken to Prometheus.

Adding a conditionally rendered ServiceMonitor template to the Helm chart to provide native Prometheus Operator integration.

This will be disabled by default to avoid breaking clusters without the operator installed.

It will brings NRC up to parity with standard CNCF controllers

Ayush4958:
## Description
Adds native Prometheus Operator `ServiceMonitor` support to the Helm chart. 
The template is safely guarded by `.Capabilities.APIVersions.Has` to prevent crashing non prometheus clusters & automatically handles TLS injection for secure metric endpoints.

## Related Issue
Fixes #443 

## Type of Change
/kind feature

## Testing
- Executed `helm template` testing various edge cases (secure vs insecure metrics).
- Helm unit tests added to cover all logic paths.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
Added Prometheus ServiceMonitor template to Helm chart via `metrics.serviceMonitor.enabled`.

### DIFF
--- a/charts/node-readiness-controller/templates/servicemonitor.yaml
+++ b/charts/node-readiness-controller/templates/servicemonitor.yaml
@@ -23,8 +23,15 @@ spec:
       scrapeTimeout: {{ .Values.metrics.serviceMonitor.scrapeTimeout }}
       {{- if .Values.metrics.secure }}
       scheme: https
+      {{- if .Values.metrics.serviceMonitor.bearerTokenFile }}
+      bearerTokenFile: {{ .Values.metrics.serviceMonitor.bearerTokenFile }}
+      {{- else }}
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      {{- end }}
+      {{- with .Values.metrics.serviceMonitor.tlsConfig }}
       tlsConfig:
-        insecureSkipVerify: true
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
       {{- end }}
       {{- with .Values.metrics.serviceMonitor.metricRelabelings }}
       metricRelabelings:
--- a/charts/node-readiness-controller/templates/servicemonitor.yaml
+++ b/charts/node-readiness-controller/templates/servicemonitor.yaml
@@ -1,0 +1,38 @@
+{{- if .Values.metrics.serviceMonitor.enabled }}
+{{- if not .Values.metrics.enabled }}
+{{- fail "You enabled metrics.serviceMonitor.enabled, but metrics.enabled is false. Please enable metrics to use the ServiceMonitor." }}
+{{- end }}
+{{- if .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "node-readiness-controller.fullname" . }}
+  namespace: {{ include "node-readiness-controller.namespace" . }}
+  labels:
+    {{- include "node-readiness-controller.labels" . | nindent 4 }}
+    {{- with .Values.metrics.serviceMonitor.labels }}
+    {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  selector:
+    matchLabels:
+      {{- include "node-readiness-controller.selectorLabels" . | nindent 6 }}
+  endpoints:
+    - port: {{ ternary "https" "http" .Values.metrics.secure }}
+      interval: {{ .Values.metrics.serviceMonitor.interval }}
+      scrapeTimeout: {{ .Values.metrics.serviceMonitor.scrapeTimeout }}
+      {{- if .Values.metrics.secure }}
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by Ayush4958** [comment]
Hi @ajaysundark ,
PR is ready for review 
take time to review it 

thanks ...

**Comment by AvineshTripathi** [comment]
cc @rawadhossain

File: charts/node-readiness-controller/templates/servicemonitor.yaml [hunk 1]
```diff
@@ -0,0 +1,38 @@
...
+    matchLabels:
+      {{- include "node-readiness-controller.selectorLabels" . | nindent 6 }}
+  endpoints:
+    - port: {{ ternary "https" "http" .Values.metrics.secure }}
+      interval: {{ .Values.metrics.serviceMonitor.interval }}
+      scrapeTimeout: {{ .Values.metrics.serviceMonitor.scrapeTimeout }}
+      {{- if .Values.metrics.secure }}
+      scheme: https
+      tlsConfig:
+        insecureSkipVerify: true
```
rawadhossain: when `metrics.secure` is enabled, don't we also need to configure the bearer token for the metrics endpoint? also, can we avoid `insecureSkipVerify` here and use the generated cert instead?

charts/node-readiness-controller/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,38 @@
+{{- if .Values.metrics.serviceMonitor.enabled }}
+{{- if not .Values.metrics.enabled }}
+{{- fail "You enabled metrics.serviceMonitor.enabled, but metrics.enabled is false. Please enable metrics to use the ServiceMonitor." }}
+{{- end }}
+{{- if .Capabilities.APIVersions.Has "monitoring.coreos.com/v1" }}
```
rawadhossain, line 5: if `serviceMonitor.enabled=true` but the CRD isn't installed, this just renders nothing. do we want to warn here, or is the current behavior intentional?

charts/node-readiness-controller/tests/servicemonitor_test.yaml
```diff
@@ -0,0 +1,108 @@
...
+      metrics:
+        enabled: true
+        serviceMonitor:
+          enabled: false
+    asserts:
+      - hasDocuments:
+          count: 0
+
+  # Note: testing .Capabilities.APIVersions.Has is not fully supported by helm-unittest,
+  # but assuming the environment supports it, we can test standard rendering.
```
rawadhossain: I think `helm-unittest` supports `capabilities.apiVersions` here. Since it's already been used below, can we also add a case where CRD isn't present?

rawadhossain (commented):
Took a first look at this.

one question on the secure metrics setup, how is the prometheus serviceaccount supposed to get the `metrics-reader` permission? 

also, looks like the new serviceMonitor values are missing from the README.

charts/node-readiness-controller/tests/servicemonitor_test.yaml
Ayush4958: yup, good catch I will add test case for it

Ayush4958:
> one question on the secure metrics setup, how is the prometheus serviceaccount supposed to get the `metrics-reader` permission?
> 
> also, looks like the new serviceMonitor values are missing from the README.

for rbac, we don't know the user prometheus namespace, its okay for cluster admin to manually bind the metric reader role. 
I will document this requirement in the README along with the new values.

charts/node-readiness-controller/templates/servicemonitor.yaml
Ayush4958: nice catch on token, I will add that. 
for insecureSkipVerify, since prometheus is usually in a different namespace, it can't easily mount the cert manager secret natively.
instead of hardcoding, I will expose empty tlsConfig: {} block in values.yaml so users can configure it however they need

charts/node-readiness-controller/templates/servicemonitor.yaml
Ayush4958: it was intentional but I can see its a bad UX so I will add a warning to the chart's NOTES.txt
about crd not being installed

Ayush4958:
hi @rawadhossain 
addressed every pointers u mentioned 
PR is up for 2nd review 

cc: @AvineshTripathi @ajaysundark 

thanks ...

PR #457, opened 2026-08-30
fix(reporter,examples): support local kubeconfig fallback and conflicting name fix

AvineshTripathi:
1. Currently the reporter uses `InClusterConfig` which does not respect `KUBECONFIG` env like node-readiness-controller
2. in cni-readiness example both `network-readiness-dry-run-rule.yaml` and `network-readiness-rule.yaml` share the same name, resulting in a conflict


ajaysundark (approved):
/lgtm

PR #461, opened 2026-09-03
feat(chart): expose automountServiceAccountToken and extra volumes

### LINKED ISSUE #442: [FEATURE] Make Helm chart configurable to satisfy Gatekeeper / Pod Security policies

The `charts/node-readiness-controller` Helm chart has two configurability gaps that prevent it from passing common Gatekeeper (gatekeeper-library) constraints without manually editing templates:

1. **`automountServiceAccountToken` is neither set nor exposed.** Several Gatekeeper policies require this field to be explicitly set on the ServiceAccount and/or pod spec. The chart does not render it anywhere today, so those constraints reject the workload.
2. **No `extraVolumes` / `extraVolumeMounts` values.** These are needed to inject a projected ServiceAccount token when `automountServiceAccountToken: false`. The controller uses `rest.InClusterConfig()`, which requires token + `ca.crt` + namespace at `/var/run/secrets/kubernetes.io/serviceaccount`, so there is currently no way to run token-free without patching templates.

Expose the missing knobs in the chart:

- `values.yaml`: add `serviceAccount.automountServiceAccountToken` (default `true`), plus `extraVolumes: []` and `extraVolumeMounts: []` with a documented projected-token example.
- `templates/serviceaccount.yaml`: render `automountServiceAccountToken`.
- `templates/deployment.yaml`: render pod-level `automountServiceAccountToken` and wire `extraVolumes` / `extraVolumeMounts`.

Example for running with the token disabled (all three sources required, or `InClusterConfig` fails):

```yaml
serviceAccount:
  automountServiceAccountToken: false
extraVolumes:
  - name: sa-token
    projected:
      sources:
        - serviceAccountToken:
            path: token
            expirationSeconds: 3600
        - configMap:
            name: kube-root-ca.crt
            items:
              - key: ca.crt
                path: ca.crt
        - downwardAPI:
            items:
              - path: namespace
                fieldRef:
                  fieldPath: metadata.namespace
extraVolumeMounts:
  - name: sa-token
    mountPath: /var/run/secrets/kubernetes.io/serviceaccount
    readOnly: true
```

- Keeping the chart as-is and having operators fork/patch templates to satisfy their policy — rejected as poor UX and hard to maintain.
- A token-only projected volume — does not work: missing `ca.crt` causes TLS verification failure against the API server, so the full three-source volume is required.

tejassinghbhati:
## Description

Implements #442.

The chart never rendered `automountServiceAccountToken`, so Gatekeeper policies that require it to be set explicitly reject the workload for leaving it to the cluster default. There was also no way to add volumes, which is exactly what you need in order to run without an automounted token.

This renders `serviceAccount.automountServiceAccountToken` on both the ServiceAccount and the pod spec, and adds `extraVolumes` / `extraVolumeMounts`. The default stays `true`, so nothing changes for existing installs.

Worth being explicit about one thing, because turning the flag off on its own leaves you with a controller that cannot reach the API server. It authenticates through `rest.InClusterConfig()`, which needs `token`, `ca.crt` and `namespace` under `/var/run/secrets/kubernetes.io/serviceaccount`. Disabling the automount removes all three, so they have to come back via a projected volume. Rather than leave that to be rediscovered, `values.yaml` carries the full working example inline.

## Related Issue

Fixes #442

## Type of Change

/kind feature

## Testing

New `automount_test.yaml` suite. Chart tests go from 37 to 43, all green, and `helm lint` is clean.

The suite covers the default on both objects, the explicit `false` case, extra volumes and mounts being wired through, and one case that seemed worth pinning: the chart already mounts webhook and metrics certificates, so it asserts those survive alongside the extra entries rather than being replaced.

I also rendered the exact configuration from the issue end to end:

```console
$ helm template x ./charts/node-readiness-controller -f pss.yaml --show-only templates/deployment.yaml
      automountServiceAccountToken: false
          volumeMounts:
            - mountPath: /var/run/secrets/kubernetes.io/serviceaccount
              name: sa-token
              readOnly: true
      volumes:
        - name: sa-token
          projected:
            sources:
            - serviceAccountToken:
                expirationSeconds: 3600
                path: token
            - configMap:
                items:
                - key: ca.crt
                  path: ca.crt
                name: kube-root-ca.crt
            - downwardAPI:
                items:
                - fieldRef:
                    fieldPath: metadata.namespace
                  path: namespace
```

One test assertion I got wrong first time, mentioning it since it is mildly surprising: `volumes:` is rendered unconditionally by the template, so on a default install the key exists with nothing under it rather than being absent. The test uses `isNullOrEmpty` instead of `isNull`.

## Checklist
- [x] `make test` passes
- [x] `make lint` passes

```release-note
The Helm chart now renders `automountServiceAccountToken` on the ServiceAccount and pod spec, and supports `extraVolumes` and `extraVolumeMounts`, so it can satisfy Gatekeeper and Pod Security policies that require the field to be set explicitly or forbid automounted tokens.
```


ajaysundark:
@Ayush4958 had a crt injection in #470 as well.

Can you help take a first pass review on this?

Ayush4958 (changes requested):
Rn In PR, user who needs `automountServiceAccountToken: false` forced them to manually copy paste 25 lines of complex ProjectedVolume YAML into extraVolumes just to keep the controller from crashing.

This cause problem if kubernetes ever changes the mechanism for in cluster auth, every user yaml breaks, rather than the chart fixing it centrally.

`extraVolumes` and `extraVolumeMounts` are great additions. The chart should be updated so that if `automountServiceAccountToken: false` is set, the templates automatically render the required `projected` volume & `volumeMounts` for it

Ayush4958:
> @Ayush4958 had a crt injection in #470 as well.
> 
> Can you help take a first pass review on this?

I reviewed the PR. It's a great feature but design put some burden on end users.

PR #463, opened 2026-09-06
test/scale: add api conflicts metric to scale test reporting

rawadhossain:
## Description

Extends the scale test harness to report `node_readiness_api_conflicts_total` metric and `*ConflictExhausted` failure reasons introduced in #320.

### What's added
- Scale test report now includes conflict and retry exhaustion numbers.
- Added `controller_runtime_reconcile_errors_total` as a sanity check on retry exhaustion.
- Fixed a gap where conflicts happening between phases weren't being counted.
- Made sure that fix doesn't affect CPU, API rate, or latency numbers - those still use original measurement window.

### Testing

- Full scale test ran successfully.
- Verified the generated report and ran the full test/vet/lint suite.

## Checklist

- [x] `make test` passes
- [x] `make lint` passes

### DIFF
--- a/test/scale/promqueries.go
+++ b/test/scale/promqueries.go
@@ -131,6 +131,67 @@ var metricQueries = []MetricQuery{
 		IsCounter: true,
 	},
 	{
+		Key:       "api_conflicts_add_taint",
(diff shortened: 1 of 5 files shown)


AvineshTripathi:
cc @Karthik-K-N ^^

